PR #2, opened 2016-01-08
Update and rename README.md to README.rst

mikemhenry:
You can see a preview of this [here.](https://github.com/mikemhenry/arcade/blob/master/README.rst) Since this is done with rst, pypi will look pretty again.


pvcraven:
Looks good. Just needs setup.py to be updated.

mikemhenry:
Thanks Travis!

PR #17, opened 2016-03-30
Update from original

mwreuter:


pvcraven:
This seems to have a lot of errors during the build. I don't think it would be a good idea to include it in the code base until the errors are resolved.

arcade/draw_commands.py
```diff
@@ -10,6 +10,11 @@
 import pyglet.gl as GL
 import pyglet.gl.glu as GLU
 
+from sympy import Symbol, nsolve
+import sympy
+import mpmath
+mpmath.mp.dps = 15
```
pvcraven: Any way to do this without adding two new libraries? If the math can be done without sympy, then the users won't have to download and install. All for supporting one draw command they might not even use.

PR #49, opened 2016-06-02
Reorganize avbin

malikoth:
* Reorganize the two copies of `avbin.dll` for Windows, and put them in the same directory
* Reorganize `setup.py` to identify only the version of the included avbin library that needs installed
* Make `setup.py` use a context manager to read the readme file
* Reorganize `arcade/sound.py` to load the specific avbin library that was installed

### DIFF
--- a/arcade/sound.py
+++ b/arcade/sound.py
@@ -27,8 +27,9 @@ def load_sound_library():
             path = "avbin"
         else:
             import sys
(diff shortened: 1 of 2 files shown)


malikoth:
# TODO
* [ ] Check that avbin is still loaded correctly on Windows

Issue #21, opened 2016-05-17
Testing: Environment setup via pip


pvcraven:
The library and documentation needs to tested more with Python 3.x and 2.x. Plus test platforms Linux, Mac, and Windows. Have several new people try and test out the library and give feedback on ease of use.

malikoth:
* Support for Python2 is officially deprecated.
* Installation works great on Mac now
* Installation into a virtualenv is as easy as `pip install arcade`

Todo
* [ ] Still need to update installation documentation

pauleveritt:
I turned on an old Mac, updated to El Capitan, and followed the installation instructions for Mac. I used the Homebrew Python and pyvenv, but otherwise, was the same.

Alas, I got the avbin import error. Should I switch to the downloaded Python and virtualenv just to give it a try?

pvcraven:
What do you mean by downloaded Python? It won't work with the Python that is already installed on the Mac. The homebrew python 3 it was supposed to work with.

I think the virtenv is required because otherwise the library has to be installed in a system folder, and that requires sudo. Which is bad. If you didn't use virtenv that installation of the library might have failed.

pauleveritt:
I just noticed that pip gave me 0.0.6. I forced it to give me 0.9.9b1 and it did a little better. I now have lib/libavbin.10.dylib

However, it's trying to do the import from a lib in the virtualenv:

ImportError: Library "/Users/pauleveritt/tmp/arcade/env35/lib/python3.5/site-packages/lib/site-packages/arcade/lib/libavbin.10.dylib" not found.

That lib directory is actually in arcade.

pauleveritt:
I have some other comments on the installation instructions. I suspect the comments should go on another issue.

pauleveritt:
I think this works fine. I'm going to close this and we'll report any specific bugs in different tickets.

PR #58, opened 2016-06-03
Update installation instructions for Mac

malikoth:
I can't really test the early parts of these instructions, as I'm not going to wipe my Mac back to a clean slate, but they should work fine.  I'm more familiar with markdown than RST, so please look at formatting and markup.  Also, any feedback on instruction wording would be greatly appreciated.


malikoth:
This PR makes progress towards solving #21 .

Issue #64, opened 2016-07-10
libavbin


mikemhenry:
Currently the Linux instructions don't work with Ubuntu 16.04. A few of the packages don't exist in 16.04. Once I get everything working I'll submit a PR for the installation page.

mikemhenry:
Okay, I think I worked out a solution but it requires a manual installation AVbin (not too difficult).

However, I'm getting segfaults if I use pulse audio
If I set pyglet to load the openal module first:
`pyglet.options['audio'] = ('openal', 'pulse', 'silent')`
Things work just fine. This will set pulse to be a fallback and then no audio if that fails as well. Unfortunately this means that on Linux audio will be mono. 

I'll submit a PR with the fix and a PR with the updated instructions.

pauleveritt:
@mikemhenry Other than getting the CI working, is there anything left to do on this ticket?

mikemhenry:
Other than the CI issue in #65 there is nothing to do, closing.

PR #65, opened 2016-07-10
Linux Audio Fix

mikemhenry:
Fixes half of issue #64

Will update Linux doc in separate PR


pauleveritt:
AppVeyor failed on the Linux build with: "ValueError: zlib is required unless explicitly disabled using --disable-zlib, aborting" I can take a look at it if anybody has any pointers?

mikemhenry:
It looks like that was an issue before this PR was merged in and is no longer an issue? I'm not sure how to check the latest linux build. If this is still an issue in master I can look at it this weekend.

PR #131, opened 2017-10-02
Add fullscreen support

dzervas:


pauleveritt:
Thanks for doing this. I had this on a branch last year but it was mixed up with some other stuff.

@pvcraven if we accept this, we should likely get it into some docs. I can work on that, if you're ok with this change.

pvcraven:
I'll take a look at it. Hopefully today.

PR #136, opened 2017-11-01
Added key_detector example that shows Arcade key constants for pressed keys.

r0the:
Hi

I created an Arcade example that displays key constants for pressed keys. It's intended as a quick way for students to lookup up a key constant.

### DIFF
--- a/examples/key_detector.py
+++ b/examples/key_detector.py
@@ -1,0 +1,64 @@
+"""
+Arcade Key Detector
+
+Shows arcade key constants for pressed keys.
+"""
+
+import arcade
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 500
+TITLE = "Arcade Key Detector"
+BACKGROUND_COLOR = arcade.color.WHITE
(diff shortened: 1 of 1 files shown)


examples/key_detector.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, TITLE)
+        self.key_dict = load_arcade_key_dict()
+        arcade.set_background_color(BACKGROUND_COLOR)
+
+
+    def on_draw(self):
+        arcade.start_render()
+        y = SCREEN_HEIGHT - LINE_HEIGHT
+        for pressed_key in self.pressed_keys:
```
pvcraven: Where does self.pressed_keys come from? It errors out for me on the line above, not knowing what it is. Is this in a different version of pyglet than I have?

r0the:
I thoght it a feature of pyglet, but it is only available in the X11 window implementation of pyglet. I'll fix this.

PR #168, opened 2017-12-22
Tiny modification when returning

Aareon:
Simply remove another variable when completing the task, then returning the value necessary.

Seems a better alternative to creating a variable with a default value, performing a check, modifying said variable if necessary, and returning.

### DIFF
--- a/arcade/physics_engines.py
+++ b/arcade/physics_engines.py
@@ -119,14 +119,13 @@ class PhysicsEnginePlatformer:
         hit_list = \
             check_for_collision_with_list(self.player_sprite,
                                           self.platforms)
-
-        result = False
+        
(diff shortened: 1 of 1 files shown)


Aareon:
Travis is *screaming* with errors mostly pertaining the values being undefined. Not sure if that’s due to the nature of the environment, but I’m not entirely sure if the build failure is necessarily my fault.

pvcraven:
Merged, thanks.

Aareon:
@pvcraven No problem! If you’d prefer, I can also PR [#169](https://github.com/pvcraven/arcade/issues/169)

pvcraven:
If it is an improvement, I'll happily take it!

PR #171, opened 2017-12-23
SpriteList.move() code needs to be cleaned up

Aareon:
https://github.com/pvcraven/arcade/issues/169

Apart from this, I don't really see what else I can change without breaking things


arcade/sprite.py
```diff
@@ -678,8 +676,6 @@ def move(self, change_x: float, change_y: float):
             sprite.center_y += change_y
```
pvcraven: I think the code from lines 659 to 671 just need to go away?

Issue #190, opened 2018-04-15
Distance displayed in side-scrolling examples appears to be incorrect


dlanghorne0428:
## Bug Report

examples/sprite_tiled_map.py
examples/sprite_moving_platforms.py
examples/sprite_ramps.py

### Actual behavior: 
Distance value displayed at right edge of map is over 5500 pixels

### Expected behavior: 
The map has 50 columns of tiles. Each tile is 64 pixels after scaling, so the distance should not be larger than 50 * 64 = 3200

### Steps to reproduce/example code:
`distance = self.view_left + self.player_sprite.right   `

I believe self.player_sprite.right is already in "world" coordinates, so don't need to add self.view_left. The check for game_over also adds view_left.   

I could try to submit a pull request.

pvcraven:
Hm, I think you are right. If you want to try a pull request, that would be great.

dlanghorne0428:
ok, I think I created the pull request successfully from a fork. Modified the game_over check in update() to be based on the number of tile columns read in from the map.

pvcraven:
Fixed in 1.3.2

PR #194, opened 2018-04-24
Issue #190: Update distance calculations in scrolling examples

dlanghorne0428:
Distance calculations for side-scroller examples based on actual pixels being rendered. Game Over decision based on size of tiled map.


pvcraven:
Thanks for the contribution!

PR #202, opened 2018-05-14
[WIP] Begins adding MacOS distribution documentation

craiglabenz:
Adds MacOS-specific documentation for bundling/distribution.

TODOs:

- [ ] Create Cookiecutter for project using this process
- [ ] Include reference to said Cookiecutter in documentation

### DIFF
--- a/doc/distribute_macos.rst
+++ b/doc/distribute_macos.rst
@@ -1,0 +1,37 @@
+.. _distribute_macos:
+
+Building for MacOS
+==================
+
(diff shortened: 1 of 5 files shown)


craiglabenz:
I would consider this PR _unready_ in the stage it was merged. I have a few more TODO items!

Apologies if you usually don't create PRs until they are believed to be ready -- my thinking was to create it right away since we were about to part ways, and this way, if nothing else, you'd have something concrete.

pauleveritt:
IMO it's fine to create PRs before finishing. GitHub teaches this as the right way....use a PR to manage an ongoing conversation.

Issue #206, opened 2018-05-16
Documentation Example Thumbnails Not Displaying


jgrantdev:
## Documentation request:
Images on the index page in the Example Code section of the documentation do not display with a local build of the documentation.

sphinx-build lists several warnings when building the documentation:
arcade/doc/examples/index.rst:12: WARNING: image file not readable: examples/thumbs/starting_template.png
....

### Where is it located?
doc/examples/index.rst

### What is wrong with it? How can it be improved?
All of the image URIs have the images in doc/examples/thumbs/ however they reside in doc/examples/

sprite_move_animation.png is not in the repository. It was removed in 650f1e2 when a gif of the screenshot was update. However the png was still being referenced for a thumbnail.

I am correcting the URIs and adding sprite_move_animation.png.

pvcraven:
Oh, there is a batch file that generates these images using [ImageMagick](https://www.imagemagick.org/script/index.php). Look in `/doc/examples`

pauleveritt:
As a note the Travis builds also warn/fail with these errors. Whichever direction we go, we should consider decreasing the noise in Travis builds. Perhaps we move stuff out of update_thumbnails.bat and into doc/preprocess_files.py (which is already run in .travis.yml)

pvcraven:
I'd suggest putting them in a preprocessing step. The animated gifs take forever, so you wouldn't want it in every step.

My worry is if we check in the thumbnails, different libraries might compress them differently, and we could get multiple full checkins of the thumbnails, inflating the repository.

pauleveritt:
If we get to the point where CI builds the website, then it makes sense to run them there instead of checking in the rendered results.

@jgrantdev We can in that case move the dependency to the dev_requirements.txt file that I have on a branch. Let's make sure though that we get these scripts all into obvious spots. Let me know if I can help.

jgrantdev:
The PR calls `preprocess_files.py` and generates the thumbnails if they do not exist when `sphinx-build` is ran.

It needs to be tested on Windows and macOS as the command to create the thumbnails with imagemagick might be different.

I also removed the call to `preprocess_files.py` from `travis.yaml` since `sphinx-build` should be making that happen, but that needs to be tested as well.

I was tempted to move around some of the utility scripts to clean up the project structure, but I want to do some more research on the correct way to do all of this.

jgrantdev:
I am working on a final fix for this that does not require you to manually delete the thumbs folder to regenerate thumbnails.

[...]

PR #207, opened 2018-05-16
Corrected image URIs and added missing image

jgrantdev:
Issue #206


jgrantdev:
See Issue #206 for reason this is being closed.

PR #212, opened 2018-05-17
Automate thumbnail generation

jgrantdev:
Issue #206 

When sphinx-build runs it opens and executes the file `doc/conf.py`. 

I made `doc/conf.py` call `doc/preprocess_files.py`.
I removed `preprocess_files.py` calls from `make.bat`, `make.sh` and `travis.yaml` 

I created `doc/generate_example_thumbnails.py`.
It checks for the directory `doc/examples/thumbs`. If that directory exists the script does nothing. If the directory does not exist it executes the appropriate command to generate the thumbnails.

`doc/conf.py` makes the call to `doc/generate_example_thumbnails.py`.

This PR does have some issues.
 - I commented out the line in `generate_example_thumbnails.py` that creates the gifs. The generated documentation and the live documentation are currently not using gif thumbnails.
 - I can not test this on windows or macOS. I do not know if I chose the correct command on line 17 of `generate_example_thumbnails`
 - The examples documentation page is trying to use a thumbnail that can not be generated because the source png does not exists. The thumbnail is on the live documentation though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ install:
   - pip install -r requirements.txt
 
 script:
-  - python doc/preprocess_files.py
   - sphinx-build -b html doc doc/build/html
   - coverage run --omit=arcade/examples/* --source=arcade setup.py test
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 # Run the generate quick API index script
+# Then generate thumbnails if they do not exist
 import runpy
 
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,5 +1,11 @@
 #!/usr/bin/env python
 
+# Run the generate quick API index script
+import runpy
+
+
+runpy.run_path('preprocess_files.py', run_name='__main__')
+
 BUILD = 0
 VERSION = "1.3.4"
 RELEASE = VERSION
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -5,6 +5,7 @@ import runpy
 
 
 runpy.run_path('preprocess_files.py', run_name='__main__')
+runpy.run_path('generate_example_thumbnails.py', run_name='__main__')
 
 BUILD = 0
 VERSION = "1.3.4"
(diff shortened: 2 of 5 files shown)


jgrantdev:
I never even though about importing the code and running it, that does seem like a clean solution.

I ended up with runpy when I read the documentation for the Sphinx conf.py http://www.sphinx-doc.org/en/master/usage/configuration.html. It mentioned that conf.py is executed with execfile(). That sounded like what I wanted to do with the preprocess_files.py so I searched for a python 3 alternative to execfile() and ended up with runpy. https://mail.python.org/pipermail/python-dev/2014-June/134992.html

I first started out trying to solve this by looking for a Sphinx extension that generates thumbnails. I would love to learn more about automating stuff like this with extensions.

I have no problem reimplementing this with a better solution.

Thank you.

pauleveritt:
Oops, just realized you mentioned on the related issue about "tempted to move around". I made a similar point on the PR.

I know you're reusing ImageMagick from what @pvcraven did but ugh, sure wish we could use Pillow and not need an external package installed via Homebrew. But that also can be improved later. I can work with you on an overhaul if you're interested.

Also, I don't know about the strategy of skipping regenerationg if the thumbnail exists. If the original ever changes, we'll never updated the thumbnail (locally). Of course on Travis it will never exist, so not a big deal.

It works on macOS.

pauleveritt:
If you’re up for gradually improving this, I’m up for working with you on it. I have (sigh) a lot of time invested in Sphinx extensions. Would you like to chat here on the ticket, real-time in a Hangout, over email?

FWIW there’s a thumbnail extension already but I don’t know if it’s current:

  https://pythonhosted.org/sphinxcontrib-images/ <https://pythonhosted.org/sphinxcontrib-images/>

—Paul

jgrantdev:
I am currently traveling and my schedule is not very regular.

I think email or slack would work best. My email address is my github username at gmail.com.

jgrantdev:
sphinxcontrib-images appears unmaintained. The last issue was filed in February stating it does not work with Sphinx 1.7.

Issue #215, opened 2018-05-25
Make Arcade compatible with PyInstaller / cx_Freeze


r0the:
## Enhancement request:
My students would like to create an executable file from their Arcade project. There are [various application bundlers available][1], but only two support Python 3.x: PyInstaller and cx_Freeze (py2exe seems to be abandoned).

I have experimented with both, but have had some issues:
- sound.py throws an exception because both PyInstaller and cx_Freeze have different implementations of the `site` module.
- I can't figure out how to bundle libav.dll with PyInstaller.
- With PyInstaller, resources can't be bundled in a single executable.

### What should be added/changed?
1. Make sound.py compatible with PyInstaller. PyInstaller uses virtualenv which doesn't provide `site.getsitepackages()`. I have a pull request that will fix that.
2. Make sound.py compatible with cx_Freeze, which returns an empty array from `site.getsitepackages()`.
3. Don't crash the application when libav cannot be loaded.
4. Support [loading of resources from PyInstaller bundles][2] using `sys._MEIPASS`.

### What would it help with?
Arcade projects could be bundled and distributes as executable files.

[1]: https://stackoverflow.com/a/14165470
[2]: https://stackoverflow.com/a/13790741

pauleveritt:
Very good job for working on this. I have a related idea which could benefit from this and possibly contribute to it.

I don't want to hijack your ticket so I'll be terse and let you say whether I should describe a bit more.

# Arcade Collective

- A monorepo for games and a community that support them
- Docs, tests, PyPI installable packages, easy to get and run
- A way to teach people not just how to write games but also to work on open source community projects

I could imagine a PyInstaller binary for the Arcade Collective which allowed browsing, launching, and updating by bundling the .git directory.

r0the:
I like your idea of an open source community to share projects. I think it would benefit the students to be able to share and try out the code of others.

I'd have more to say on this, but I think this goes in a different direction than what I intended with this issue. The goal is that my students can share their projects with friends with minimal effort. Just "here's the file, try it out."

Would you be so kind and open a separate ticket with you suggestion, so we can continue a discussion there?

pauleveritt:
@r0the I wrote some ideas in #218 that hopefully explain a little more. I added a part covering what I think you discuss.

In some cases, I don't think "here's the code, try it out" will work. The code might have images that need to be in the right place with the right name, for example. So I proposed a way to package into a zip archive then open a zip archive.

[...]

PR #216, opened 2018-05-25
Handle missing site.getsitepackages attribute in sound.py for virtualenv compatibility.

r0the:
With this change, sound.py will not crash when executed in virtualenv. virtualenv is used by PyInstaller.

See issue #215.


pvcraven:
Thanks!

pauleveritt:
Just ran into this with my son.

PR #217, opened 2018-05-26
Fix typo in file docstring

superking84:
It references sprite_starting_template as the filename rather than just starting_template.


pvcraven:
Thanks!

superking84:
No problem at all... I'm just diving into Arcade and love it so far, I used to fiddle with Pygame a long time ago so it's cool to see a modern game framework for Python being maintained.

PR #219, opened 2018-05-27
Fixing SpriteList.move()

henne90gen:
Also changing the last_center_x/y was preventing the draw method from detecting a change and rendering the sprite again.

### DIFF
--- a/arcade/sprite_list.py
+++ b/arcade/sprite_list.py
@@ -346,9 +346,7 @@ class SpriteList(Generic[T]):
     def move(self, change_x: float, change_y: float):
         for sprite in self.sprite_list:
             sprite.center_x += change_x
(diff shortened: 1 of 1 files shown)


pvcraven:
Wait, what's the bug?

henne90gen:
The sprite checks whether the center_x or center_y have changed since the last draw call by comparing it to last_center_x or last_center_y respectively. last_center_x/y are being initialized to center_x/y. This means that if we update them at the same time with the same amount, there is never a difference between center_x/y and last_center_x/y when we get to the draw method. So we are updating the center_x/y coordinates, but never redraw the sprite.

henne90gen:
Ups, accidentally closed the PR ^^

PR #220, opened 2018-05-27
Pytest doctest

pauleveritt:
I think this is in good enough shape to start a PR conversation about merging. I haven't converted any existing doctests, haven't increased the coverage, and perhaps there are some other ideas about structure. But all of that could happen in smaller tickets.


pvcraven:
* AppVeyor doesn't run any tests. But it is terrible for tests on an OpenGL library when AppVeyor doesn't support OpenGL.
* TravisCI is happy. 
* I can run the tests inside PyCharm which is nice.
* What's the best way to run tests off the command line? `python setup.py test` doesn't work.

pauleveritt:
- I can research whatever you'd like for AppVeyor vs. Travis
- From command line, look in .travis.yml for the example (but omit the flag that skips the not-Travis tests)

pauleveritt:
@pvcraven I'd like to get this part merged and put into place before starting a bunch of test writing. I'll offer to own and clean up any damage this does with Travis etc. along the way.

I can start a PR and we can talk there about what changes were made, how it changes dev workflow, etc.

PR #221, opened 2018-05-27
Setting window as arcades window

henne90gen:
This enables access to the window instance through arcade.get_window(), if one is using a custom window class. As a new user I expected arcade.get_window() to give me current window instance. I was quite disappointed when this was not the case. I was able to achieve it by using this code:
`
window = MyWindow()
window.setup()
arcade.set_window(window)
arcade.run()
`
It works, but it would be nice if it was done automatically on window creation.


pvcraven:
Thanks!

Issue #224, opened 2018-06-02
ver. 1.3.4: Asteroid Example Crashes When Playing Sound


pybokeh:
## Bug Report
Crash
Traceback (most recent call last):
  File "/opt/python3.6.4/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/python3.6.4/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/arcade/examples/asteroid_smasher.py", line 385, in <module>
    main()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/arcade/examples/asteroid_smasher.py", line 381, in main
    arcade.run()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/arcade/window_commands.py", line 191, in run
    pyglet.app.run()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/app/__init__.py", line 138, in run
    event_loop.run()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/app/base.py", line 142, in run
    self._run()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/app/base.py", line 155, in _run
    platform_event_loop.step(timeout)
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/app/xlib.py", line 125, in step
    device.select()
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/canvas/xlib.py", line 168, in select
    dispatch(e)
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/window/xlib/__init__.py", line 922, in dispatch_platform_event_view
    event_handler(e)
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/window/xlib/__init__.py", line 1089, in _event_key_view
    self.dispatch_event('on_key_press', symbol, modifiers)
  File "/home/pybokeh/envs/gamedev/lib/python3.6/site-packages/pyglet/window/__init__.py", line 1232, in dispatch_event
    if EventDispatcher.dispatch_event(self, *args) != False:
[...]

PR #228, opened 2018-06-05
Fix sound playback of WAV files on Linux.

r0the:
This commit fixes sound playback on Linux (issue #224). As suggested by @dangillet, WAV files are loaded to memory as StaticSource and are played back by multiple players.

This fix assumes that WAV files are used for short sound effects only. A better solution would be to distinguish between sound and music in the API as suggested by @dangillet.

### DIFF
--- a/arcade/sound.py
+++ b/arcade/sound.py
@@ -129,14 +129,20 @@ def _playsound_osx(nssound):
         nssound.copy().play()
 
 
+players = []
+MAX_PLAYERS = 5
 def _playsound_unix(sound):
     """Play a sound using GStreamer.
     Inspired by this:
     https://gstreamer.freedesktop.org/documentation/tutorials/playback/playbin-usage.html
     """
     # pathname2url escapes non-URL-safe characters
-    if isinstance(sound, pyglet.media.sources.riff.WaveSource):
-        sound.play()
+    global players
+    if isinstance(sound, pyglet.media.sources.base.StaticSource):
+        while len(players) >= MAX_PLAYERS:
+            p = players.pop(0)
+            p.delete()
+        players.append(sound.play())
         return
 
     import os
@@ -170,7 +176,7 @@ def _playsound_unix(sound):
 
 def _load_sound_unix(filename: str) -> typing.Any:
     if filename.endswith(".wav"):
-        my_sound = pyglet.media.load(filename)
+        my_sound = pyglet.media.StaticSource(pyglet.media.load(filename))
         return my_sound
     else:
         return filename
@@ -201,6 +207,3 @@ else:
     load_sound = _load_sound_unix
 
 del system
-
-
-


pvcraven:
Hi, I won't have time to look at this until next week. Thanks, sorry I can't give you a quick reply.

arcade/sound.py
```diff
@@ -170,7 +176,7 @@ def _playsound_unix(sound):
 
 def _load_sound_unix(filename: str) -> typing.Any:
     if filename.endswith(".wav"):
-        my_sound = pyglet.media.load(filename)
+        my_sound = pyglet.media.StaticSource(pyglet.media.load(filename))
```
dangillet: Why not simply use `my_sound = pyglet.media.load(filename, streaming=False)` ?

zeziba: Depending on how you want to use the file a static source will load and decode completely into memory. This has the benefit of being able to use it as many times and wanted with no drawbacks other than taking up memory. It also allows the media to be played by as many players as you want as well.

For one and done sounds that only need to be decoded right before use is why you would use ```pyglet.media.load()```., it also is limited by how many players can use that sound at once.

Here is an example.
```
def sound():
    s = pyglet.media.load("asound")
    s.play()
```
Every time a user wants to play a sound pyglet has to load and decode the sound, it is smart enough to keep the file cached for the next go around and is why a user doesn't experience any lag. From pyglet's documentation if a user is going to be using a media over and over it is smarter to use static loading. Once loaded and decoded it can be played much faster and by multiple players from the ```pyglet.media.Player()``` class. This allows for some interesting interactions with media not limited to sounds.

Check out this link for more information.
https://pyglet.readthedocs.io/en/pyglet-1.2-maintenance/api/pyglet/media/pyglet.media.StaticSource.html#pyglet.media.StaticSource

Admittedly it is better to use static loading for the ```Player``` so keep that in mind. Also people who have issues dynamically loading many things might find it a better choice as well to statically load things.

zeziba: I forgot to mention as well the base ```pyglet.media.load().play()``` works but it relies on the python garbage collection to remove it. So when using the ```arcade.load_sound()``` then ```arcade.play_sound()``` does the same thing, it just so happens if the GC happens between two plays of the sound it will have to completely decode the file again to play it.

Now that I am thinking about it, the base use case of arcade is unintended. It is implicit and not explicit in its behavior and might cause confusion down the line so going to a well defined static/non-static load would likely be a better solution.

Something like this would likely be what is required.

```python
def load_sound(sound, load_method=None):
    # w/e per-processing code here
    if load_method == arcade.Sound.static:
        return pyglet.media.StaticSource(pyglet.media.load(sound))
    elif load_method == arcade.Sound.stream:
        return pyglet.media.load(sound, streaming=True)
    else:
        return pyglet.media.load(sound, streaming=False)
```

This is mentioned in this documentation on pyglet
https://pyglet.readthedocs.io/en/pyglet-1.2-maintenance/api/pyglet/pyglet.media.html

dangillet: Adding the keyword `streaming=False` will return a `StaticSource` instance. See https://bitbucket.org/pyglet/pyglet/src/3d464716883b1a8b07a67f4932fdbe7c4af788f9/pyglet/media/sources/loader.py#lines-46

Caching is a good point. Currently with the `pyglet.resource.media` function, the `Source` is not cached while textures, images and animations are. `pyglet.media.load` does not cache the `Source` either. So it will need to be loaded and decoded every single time. Of course you can keep a reference to your `StaticSource` and re-use it on multiple players.

Regarding garbage collection, I'm actually proposing in pyglet to remove this odd feature (bug) where the `Player` stays in memory while we don't have a reference to it, just because `pyglet.clock` keeps a strong reference to some of its methods. See https://bitbucket.org/pyglet/pyglet/pull-requests/130/eventdispatcher-and-clock-only-hold/diff You will see that I'm facing unfortunately some performance penalties and I don't know yet if Pyglet maintainers will find it acceptable or not.

zeziba: Good looking, I didn't notice that streaming true/false did that when I looked over the code, albeit I did skim over it.

I am still of mind that making it have explicit behavior would be better in the long run as unintentional bugs can be mitigated before they even happen.

pvcraven:
Thanks!

Issue #229, opened 2018-06-06
Add Player for background music.


zeziba:
## Enhancement request:
Add the ability to play music in the background with little setup.
By doing this it will allow a user to add atmosphere to an environment easily and quickly.

### What should be added/changed?
Change the sound.py file so that it incorporates the above feature so that a user can easily add background music to an application.

### What would it help with?
It would make it easier to add music to a program allowing for a more user friendly experience.

I have already implemented this into my project as it allows for quick and easy use of background noise to add atmosphere to an environment.

Here is an example of what it will look like to add background music to a program.

```python
bg = arcade.Player()
bg.load_dir('resources\\sounds\\background')
bg.add('Game-Menu')
bg.play()
bg.volume(5)
```

The player I have made supports the queue system of pyglet and as such allows for multiple files to be loaded into memory for consumption. With the queue system there are features such as next, stop, looping, play, add, volume, clear, load_file and load_dir.

As I showed above I just use the file name without extension and can access the music file. Both a directory and single file can be loaded. After the files are loaded each file is put into a dictionary and associated with its filename allowing it to be accessed by name.

```python
# The files in the resources folder are wave1, wave2, wave3
bg = Player()
bg.load_dir("resources/")
print(bg.music)
#OUTPUT: ['wave1', 'wave2', 'wave3']
```

By doing the above it is easy to see what is loaded and makes it so each can be iterated over quickly and efficiently.

Overall this is a quality of life improvement that would allow a user to quickly and effectively add background music to a file.

rb-dahlb:
I am also interested in this. A way to play different music based on game state for example, one piece of music for "start menu", one for "in game" and another for "game over screen".

Would need the API to be able to set a parameter for looping sound, so if you stay in the "start menu", the music just loops, and also an API call for stopping a sound currently playing, for example when the game is switching mode from "start menu" to "in game".

pvcraven:
This enhancement is not currently in process. Please re-open if you'd like to work on it. A full list of desired enhancements is available at:

http://arcade.academy/enhancement_list.html

pvcraven:
Reopening- See issue #372. Should be possible now.

pvcraven:
Example now posted: https://arcade.academy/examples/background_music.html#background-music

PR #230, opened 2018-06-06
Update sound.py

zeziba:
Issue: #229

Add a Player class to make management of background music easier.


pvcraven:
Thanks! I'm away from a computer this week. I'll look at your PR in detail next week.

pauleveritt:
The doctest at:

```
229         >>> player.load_dir("arcade/examples/sounds/")
```

...fails on macOS with:

```
Expected nothing
Got:
    Warning: .ogg (Ogg Vorbis) files are not compatible with the Mac on the Arcade library.
    Warning: .ogg (Ogg Vorbis) files are not compatible with the Mac on the Arcade library.
    Warning: .ogg (Ogg Vorbis) files are not compatible with the Mac on the Arcade library.
```

I can make a new ticket for this if you'd like.

zeziba:
If you know how to get the doctest to check for a specific version and change its behavior based on that I will add it to the doc string but I removed it and put the general use case that I was going for.

pauleveritt:
When you say “specific version”, can you explain what you mean?

But yes, in general the doctests are not as featured as regular pytest tests. For example, .travis.yml shows the use of markers to restrict what tests won’t work right on Travis.

—Paul

zeziba:
If it is mac os it will output the x3 warnings as of right now because there are that many .ogg files in that specific directory. The player is not designed to handle the implementation errors from the file loader as it relies on the arcade implementation to always have play_sound as it backbone so if it changes the player changes as well.

Adding os specific markers would allow for the doctests though I have little knowledge of travis.

pauleveritt:
As a suggestion, have the docstring point at a directory which only contains .wav sounds, meaning, sounds which can play on all platforms. The docstring's purpose is to show loading all files in a directory, not loading different kinds of sounds.

zeziba:
Well, from a testing standpoint I would like it to fail when the file type is unsupported. This will ensure compliance with how the function is supposed to work in the future. In that vein is there a way to run multiple doctests on the same file? This would allow for it to run the test for all three systems by changing the flags pertaining to the system.

pauleveritt:
As you noted, the handling of file types per platform is further downstream than your code. Your code should just presume it works, at least from a unit test perspective. We can write some non-doctests for the sound.py in Arcade if we want per-platform exceptions, IMO. Just my two cents.

I don't know of the right way to parameterize doctests. (I'd argue that it's abusing the purpose of doctests.)

PR #232, opened 2018-06-08
Thumbnail generation

jgrantdev:
#206 

Thumbnails are now generated using Pillow instead of image magick.

The thumbnails are only generated if they do not exist, or the source file has been modified more recently than the thumbnail.

The thumbnails are a slightly different size than the ones previously generated with image magick, but the documentation still renders well. This is due to Pillow not having an easy extent function. IMO it looks better this way because it removes a large gap below some thumbnails and the text.

Due to a bug with PyCharm I only tested with Sphinx 1.6.7, but it should render the same with the latest version of Sphinx.


pvcraven:
I _really_ like this. Removing the dependency on image magick is great. Thanks.

Issue #199, opened 2018-05-10
version 1.3.3 appears to be broken on windows


jrheard:
### Actual behavior:

on windows (not sure which version, sorry, making this bug report for a student i'm helping):

`pip install --user arcade`

`python`

`import arcade`

output:

```
Traceback (most recent call last):
  File "Y:\Comp Sci 3-4\Per3_evan_arcade\arcade_test_file.py", line 1, in <module>
    import arcade
  File "C:\Users\emrazik7366\AppData\Roaming\Python\Python36\site-packages\arcade\__init__.py", line 21, in <module>
    from arcade.sound import *
  File "C:\Users\emrazik7366\AppData\Roaming\Python\Python36\site-packages\arcade\sound.py", line 173, in <module>
    _load_sound_library()
  File "C:\Users\emrazik7366\AppData\Roaming\Python\Python36\site-packages\arcade\sound.py", line 61, in _load_sound_library
    pyglet.lib.load_library(path)
  File "C:\Users\emrazik7366\AppData\Roaming\Python\Python36\site-packages\pyglet\lib.py", line 158, in load_library
    raise ImportError('Library "%s" not found.' % names[0])
ImportError: Library "C:\Program Files (x86)\Python36-32/lib/site-packages/arcade/Win32/avbin" not found.
```

### Expected behavior:

`import arcade` shouldn't raise an ImportErorr, and should just silently work successfully.

### Steps to reproduce/example code:

see above

extra context: "C:\Program Files (x86)\Python36-32/lib/site-packages/" doesn't seem to have an "arcade" directory.

version 1.3.2 works fine on these windows machines, so we'll just use that for now - but 1.3.3 reliably reproduces this issue on multiple windows machines in this classroom.

thanks!

pvcraven:
Hm, looks like it is looking in a local-to-the-user folder. I've seen something similar with virtual environments.  I probably need it to fail silently if it can't find the library. 

Thanks for the report. I'm at PyCon now. Hope to fix it early next week.

jgrantdev:
This problem occurs when you install arcade to the user site, it does not occur when you you are in a virtual environment.

We are asking pyglet to load avbin from the global site directory.

We should try to load it from the user-site directory, if an ImportError is raised then try to load it from the global site directory.

kankri:
I just learned about Arcade and decided to give it a try, but stumbled into this issue with **version 1.3.4** on Windows:

```
>pip3 install --user arcade
[...]

PR #233, opened 2018-06-08
Attempts to load avbin from user site before global site

jgrantdev:
#199 

This might need some work before being merged.

This tries to load avbin from the user directory, if pyglet raises an ImportError then it tries to load avbin from the global site directory.

 - It does not handle the ImportError appropriately if avbin can not be loaded at all.
 - It does not log any information.
 - I am unsure why hasattr() is being called for getsitepackages() and if I should be doing it for getuserbase().


pvcraven:
I'll look at it more carefully before release, when I go through all the platforms. Thanks.

Issue #235, opened 2018-06-09
Appending list of Textures to Sprite


jgrantdev:
## Enhancement request: 
Add method to append a list of Textures to a Sprite

### What should be added/changed?
Add a method to the Sprite class that allows you to pass a list of Textures to be added to the Sprite.textures list.

### What would it help with?
Loading a single texture from a file and appending it to a sprite looks like this:

```
campfire = Sprite()
campfire_texture = arcade.load_texture('CampFireFinished.png', 0, 0, 64, 64)
campfire.append_texture(campfire_texture)
```

If you want to load multiple textures it looks like this:

```
campfire = Sprite()

frame_locations = [[0, 0, 64, 64],
                  [64, 0, 64, 64],
                  [128, 0, 64, 64],
                  [192, 0, 64, 64],
                  [256, 0, 64, 64]]

campfire_textures = arcade.load_textures('CampFireFinished.png', frame_locations)
for campfire_texture in campfire_textures:
    campfire.append_texture(campfire_texture)
```

A cleaner interface that lines up with `load_texture()` and `load_textures()` would look like this:

```
campfire = Sprite()

frame_locations = [[0, 0, 64, 64],
                  [64, 0, 64, 64],
                  [128, 0, 64, 64],
                  [192, 0, 64, 64],
                  [256, 0, 64, 64]]

campfire_textures = arcade.load_textures('CampFireFinished.png', frame_locations)
campfire.append_textures(campfire_textures)
```

The implementation is simple:
```
    def append_textures(self, textures: List[Texture]):
        self.textures.extend(textures)
```

jgrantdev:
This is marked as Done in Release 1.3.5, the PR is closed and not merged.

I am a little confused. Should I leave this issue open?

pvcraven:
I close it only when there is a production release.

Paul Vincent Craven

pvcraven:
Closing.

PR #236, opened 2018-06-09
Added append_textures method to Sprite class

jgrantdev:
#235 

Adds convenience method to Sprite class for extending current textures list.


pvcraven:
Thanks!

Issue #225, opened 2018-06-04
draw_xywh_rectangle_filled position parameters define bottom left instead of top left


LexVerheesen:
### Actual behavior:
draw_xywh_rectangle_filled() Parameters define bottom_left_x and bottom_left_x

### Expected behavior:
draw_xywh_rectangle_filled() Parameters define top_left_x and top_left_x (as documented) or the documentation has to change.

### Steps to reproduce/example code:
import arcade

SCREEN_WIDTH = 800
SCREEN_HEIGHT = 600

PADDLE_WIDTH = 20
PADDLE_HEIGHT = 100

arcade.open_window(SCREEN_WIDTH, SCREEN_HEIGHT, "XYWH Draw Test")
arcade.set_background_color(arcade.color.WHITE)
arcade.start_render()
# rectangle out of bounds
arcade.draw_xywh_rectangle_filled(0, SCREEN_HEIGHT, PADDLE_WIDTH, PADDLE_HEIGHT, arcade.color.GRAPE)
arcade.finish_render()
arcade.run()

jgrantdev:
I agree.

I believe this is the error:

https://github.com/pvcraven/arcade/blob/master/arcade/draw_commands.py#L1501

To find the center_y when given the top_y we should be subtracting half of the height, not adding.

jgrantdev:
Two other methods have the same error.

https://github.com/pvcraven/arcade/blob/f1186132203e555da9b9498776adf1d25fb3c5f1/arcade/draw_commands.py#L1332

https://github.com/pvcraven/arcade/blob/f1186132203e555da9b9498776adf1d25fb3c5f1/arcade/draw_commands.py#L1642

This is a good candidate for the first-timers-only label.

pvcraven:
I suppose this is a matter of interpretation. 

If you define a box at a y of "200" and a height of 100, would you expect the y to extend to 300 or 100? I interpreted it at to "300". So therefore x, y would start in the bottom left instead of top right.

If you work with pygame or any other old-style coordinates where 0, 0 is in the upper left, then of course it x, y would be in the top left.

I'm not quite sure what is right here.

jgrantdev:
I am not at my computer, but I believe the docs and code calls the arguments top_left_x and top_left_y.

dangillet:
I believe the fix should be in the documentation and variable names. 

```
def draw_xywh_rectangle_filled(top_left_x: float, top_left_y: float,
                               width: float, height: float,
[...]

PR #240, opened 2018-06-14
Fix naming of x and y coordinates in *_xywh_* functions

eruvanos:
Fixing issue #225

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -8,14 +8,16 @@ Buffered Draw Commands.
 """
 # pylint: disable=too-many-arguments, too-many-locals, too-few-public-methods
 
-import math
 import ctypes
+import math
+from typing import List
+
 import PIL.Image
 import PIL.ImageOps
 import pyglet
 import pyglet.gl as gl
 from pyglet.gl import glu as glu
-from typing import List
+
 from arcade.arcade_types import Color
 from arcade.arcade_types import PointList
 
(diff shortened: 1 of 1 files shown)


LexVerheesen:
Thank you for your pull request.
Would be great if we also changed the description a bit.

How about:
> Draw a rectangle by specifying left, right, top, and bottom edges.
Draw a rectangle extending from bottom left to top right

> :bottom_left_y: The y coordinate of the top of the rectangle.
:bottom_left_y: The y coordinate of the bottom of the rectangle.

eruvanos:
I changed the docs and in addition added some more docs for draw_xywh_rectangle_textured.
Furthermore added some parameters, that are also available in draw_texture_rectangle.

pvcraven:
Thanks! Happy to merge this in, if you could address the PEP thing with spaces in parameters:
https://www.python.org/dev/peps/pep-0008/#other-recommendations

eruvanos:
I changed it, just checked with pep8: 
> When combining an argument annotation with a default value, use spaces around the = sign (but only for those arguments that have both an annotation and a default).
>Yes:
>
>def munge(sep: AnyStr = None): ...
>def munge(input: AnyStr, sep: AnyStr = None, limit=1000): ...
>No:
>
>def munge(input: AnyStr=None): ...
>def munge(input: AnyStr, limit = 1000): ...

Give me a hint weather I should change it back :)

eruvanos:
Any requirements left?
Greetings

pvcraven:
Thanks!

PR #261, opened 2018-07-29
Fix Spatial Hashing for Sprite update

jrrickerson:
Spatial hashing had a few issues causing the example game `sprite_collect_coins_move_circle` to fail to properly detect collisions with a moving player and coins.  Applied several fixes:
- Modified the `clear_spatial_hashes` method on Sprite to instead update the spatial hashes.
- Fixed a broken doctest for `check_for_collision_with_list` by preventing SpatialHash from returning duplicate references when the same sprite appeared in multiple hash cells.
- Removed an obsolete reference to `append_object` on SpatialHash

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -343,11 +343,13 @@ class Sprite:
     def __lt__(self, other):
         return self.texture.texture_id.value < other.texture.texture_id.value
 
-    def clear_spatial_hashes(self):
+    def update_spatial_hashes(self):
+        """Update this sprite's spatial hash for any that in which it has been
+        included."""
         for sprite_list in self.sprite_lists:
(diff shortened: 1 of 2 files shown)


pvcraven:
Check [here](https://github.com/pvcraven/arcade/blob/opengl_shaders/arcade/sprite_list.py). There are conflicts between two different sets of fixes that we need to get resolved.

pvcraven:
Also see the [sprite class](https://github.com/pvcraven/arcade/blob/opengl_shaders/arcade/sprite.py)

jrrickerson:
I probably won't be able to take a look at this until this weekend but I'll see if I can get around the conflicts without having to merge your whole shaders branch.

pvcraven:
At this point I think the fixes are now in the code in some form or fashion, even if not directly merged.

PR #286, opened 2018-12-27
[WIP] decorator alternative

### LINKED ISSUE #193: Decorator examples don't work
## Bug Report

Decorators need some help.

MrGallo:
Hello all, I wrote an alternative to the current decorator. Check it out and let me know if it's worth developing further.

Could potentially resolve #193

### DIFF
--- a/arcade/decorator_alternative.py
+++ b/arcade/decorator_alternative.py
@@ -1,0 +1,63 @@
+import arcade
+
+__all__ = ["override", "decorator_run"]
+
+_registry = {}
+
+
+class ArcadeContextError(AssertionError):
+    """An error raised when override functions are called outside the context of an arcade.Window object."""
+    pass
+
+
+def _bind(window: arcade.Window):
+    """Overrides the stored functions from the registry into the arcade.Window object.
+    Args:
+        window: The window object whose functions we want to override.
+    """
+    for method_name, func in _registry.items():
+        setattr(window, method_name, func)
+
+
+def override(func):
+    """Registers override functions for later binding with the _bind function.
+    Called as a decorator.
+
+    Args:
+        func (function): The function that will be stored into _registry
+            to eventually override the arcade.Window object method of the
+            same name.
+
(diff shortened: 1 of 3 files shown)


arcade/decorator_alternative.py
```diff
@@ -0,0 +1,63 @@
+import arcade
+
+__all__ = ["override", "decorator_run"]
+
+_registry = {}
+
+
+class ArcadeContextError(AssertionError):
```
MrGallo: Not too sure on the name of this, or whether a custom error is required. Alternatives?

arcade/examples/decorator_bouncing_balls.py
```diff
@@ -0,0 +1,111 @@
...
+    for ball in ball_list:
+        arcade.draw_circle_filled(ball.x, ball.y, ball.size, ball.color)
+
+    # Put the text on the screen.
+    output = "Balls: {}".format(len(ball_list))
+    arcade.draw_text(output, 10, 20, arcade.color.WHITE, 14)
+
+
+@arcade.override
+def update(delta_time):
```
MrGallo: All methods to be overridden need to be named and have identical parameters according to how it appears in `arcade.Window`. You cannot override a non-existent or non callable attribute.

MrGallo:
An alternative to decorators all together is [manually overriding the arcade.Window methods](https://gist.github.com/MrGallo/a41f3ad6e7f7b2159c12dbc0e0543f1a).

MrGallo:
So pyglet windows already have decorator support by using the window variable's `.event` method. Check this [arcade code](https://gist.github.com/MrGallo/c6f0a2ababa6a827e519b834cb725c76).

If this is a viable solution, I would like to suggest having `arcade.open_window()` return the window object to allow for:
```python
window = arcade.open_window(WIDTH, HEIGHT, "My Arcade Game")

@window.event
def on_draw():
    # ...
```

[Pyglet docs](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/programming_guide/events.html)

PR #290, opened 2019-01-06
Sprites with pymunk.Vec2d for position and velocity

### LINKED ISSUE #289: Feature request: Sprite position and velocity as Vector
## Enhancement request: 
Represent `arcade.Sprite.position` and `velocity` as a Vector.

I would like to suggest having the `arcade.Sprite` base-class use an arcade Vector class.

### What would it help with?
I have been working with arcade and I'm seriously considering switching to it to teach my high school classes, so, nice work on the project as a whole! Every time I make a little game, I always extend the `arcade.Sprite` to incorporate `pymunk.Vec2d` to store `Sprite.position` and `Sprite.velocity`. 

To alleviate the need to import another dependency, as noted by @pvcraven, arcade should have its own Vector class.

Working with vectors allows for greater readability and mathematical manipulation of the `position` and `velocity` fields.

If one, wishes, they can still use the `center_x` and `change_x` fields without touching vectors.
But, the change would open up the possibility for:

```python
sprite.position += sprite.velocity
sprite.velocity.y -= GRAVITY
sprite.position = 100, 100
sprite.velocity *= 0.95

difference_vector = player.position - sprite.position
if difference_vector.length < something:
    # do something
```

MrGallo:
A decent stab at incorporating `pymunk.Vec2d`  for Sprite position and velocity. The change should still accommodate all the different ways people currently manipulate the position and velocity fields. I created an example file with 5 different ways of doing so to test it out. It is in the `examples/sprite_vec2d.py` file (which can be deleted eventually since it's not an example, just an interactive test).

resolves #289

### DIFF
--- a/arcade/arcade_types.py
+++ b/arcade/arcade_types.py
@@ -5,8 +5,11 @@ from typing import Tuple
 from typing import List
 from typing import Union
 
+from pymunk.vec2d import Vec2d
+
 RGB = Union[Tuple[int, int, int], List[int]]
 RGBA = Union[Tuple[int, int, int, int], List[int]]
 Color = Union[RGB, RGBA]
-Point = Union[Tuple[float, float], List[float]]
+Point = Union[Tuple[float, float], List[float], Vec2d]
+Vector = Point
 PointList = Union[Tuple[Point, ...], List[Point]]
--- a/arcade/examples/sprite_vec2d.py
+++ b/arcade/examples/sprite_vec2d.py
@@ -1,0 +1,109 @@
+from collections import defaultdict
+import arcade
+
+
+WIDTH = 640
+HEIGHT = 480
+SPRITE_SPEED = 5
+SPRITE_SCALE = 0.5
+keys_pressed = defaultdict(bool)
+
+
+player = arcade.Sprite(filename="images/character_sprites/character0.png",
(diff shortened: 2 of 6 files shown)


pvcraven:
I've thought about something similar. But I hate to add a dependency to Pymunk for one small class.

MrGallo:
How about a `Vector` class native to Arcade? Maybe uses some [`numpy` arrays](https://stackoverflow.com/a/12051080/9754461) behind the scenes?

dangillet:
Hi,

There is a pure Python Vector class existing cocos2D: https://github.com/los-cocos/cocos/blob/master/cocos/euclid.py#L60

I don't know if that's necessarily a good idea to bring it into arcade. I just wanted to let you know that there was an already existing decent implementation.

pvcraven:
I'm not opposed to improving the implementation. We have a dependency on numpy for the 2.0 release, so that's already there and doesn't concern me like a PyMunk dependency would. 

The key is to make sure whatever implementation we have does not slow down sprite movement. We need to run the stress test for moving sprites before/after and confirm things don't slow.

MrGallo:
I'll get to work on a Vector class for arcade. Any suggestions on where to put it? I thought the best place would be in `geometry.py` (I have no idea). Feel free to close this PR.

PR #297, opened 2019-02-01
Change decorator implementation and fix decorator examples

### LINKED ISSUE #193: Decorator examples don't work

MrGallo:
Pyglet comes with a decorator in its `Window` class already. I modified arcade decorator examples to use this. I also (presumptuously) removed the old decorator code.

Let me know how this works for you. I have another example [here](https://gist.github.com/MrGallo/c6f0a2ababa6a827e519b834cb725c76).

You can use `@window.event` to decorate any pyglet events listed [here](https://github.com/adamlwgriffiths/Pyglet/blob/master/pyglet/window/__init__.py#L1563)

Resolves #193


pvcraven:
Thanks! This really needed some help. I'm looking at your PR.

PR #307, opened 2019-02-09
[WIP] Sprite vector modification

### LINKED ISSUE #289: Feature request: Sprite position and velocity as Vector

MrGallo:
This is a work-in-progress for implementing a Vector class in arcade. Inspired by [pyeuclid](https://github.com/ezag/pyeuclid/blob/master/euclid.py) and [PVector](https://processing.org/reference/PVector.html). 

Done:
- Work on `Vector` class and its tests.

Next:
- Write documentation for Vector class
- Use the `Vector` class for `arcade.Sprite.position` and `arcade.Sprite.velocity`.
- Stress-test against previous implementation.

Closes #289


pvcraven:
Vector class now starting to get merged in with different PR, and performance tests in place.

ReblochonMasque:
Why is `normalize` not returning a Vector of `magnitude == 1`

MrGallo:
How did you use `normalize`? Does your vector have an initial magnitude of 0?

ReblochonMasque:
I was reading the code, and I was looking at this:

```
    @staticmethod
    def normalize(vector: Union['Vector', Point]) -> 'Vector':
        """Returns a Vector whose magnitude is scaled to 1.0"""
        x, y = vector
        try:
            return Vector(
                x/max(abs(x), abs(y)),
                y/max(abs(x), abs(y))
            )
        except ZeroDivisionError:
            return Vector(0, 0)
```
and related tests in commit d33f3b401df01a930d2ff8c1a8f3c139dd1d21e7

But I think this was subsequently changed in commit 0f44d1193e508a56fa7e7daf4f4a7bc3104b1376 to the following, and associated tests.

```
    def normalize(self) -> 'Vector':
        """Normalizes Vector magnitude scaled to 1.0"""
        mag = self.magnitude

        try:
            self.x = self.x / mag
            self.y = self.y / mag
        except ZeroDivisionError:
            pass
        return self
```

MrGallo:
I made the choice to have `.normalize()` be an action upon a vector object. It changes the original vector. It is helpful, though, to have two options when you want to normalize. One to change the vector itself (like the current `.normalize()`) and another option to return a normalized copy of the original. Some Python `Vector` libraries out there call the normalized copy `normalized()`.

ReblochonMasque:
Yes, but this is not my point: the two methods copied above produce entirely different results (outside of whether they return a new object or mutate self).

MrGallo:
`self.magnitude` is a property. No `()` required.
The code for magnitude is inside the `Vector` class:
```python
    @property
    def magnitude_sq(self) -> float:
        return self.x**2 + self.y**2

    @property
    def magnitude(self) -> float:
        return math.sqrt(self.magnitude_sq)
```

Is there something wrong with the most recent `normalize()` method a696da6fb4f8deb7baaa097bf7c42a4fdd0db58d? Give some sample data and the incorrect result so I can have a closer look.

ReblochonMasque:
ok, I was looking at commits, which appear to have been subsequently modified. I am now looking at your branch at https://github.com/MrGallo/arcade/blob/sprite-vector-modification/arcade/vector.py, and indeed the normalize() method looks correct there.

@property decorating magnitude was not visible in the commits; I saw it in your code.

Good work, this class is going to be very useful indeed.

**still unclear:**  

Multiplication of a vector by a vector/point is not what this function returns... We probably should have alternate methods for dot & cross product, as operator overloading notation is not quite well defined. 
`__mul__` should probably limit itself to multiplication by a scalar.
```
    def __mul__(self, other: Union['Vector', Point, float]) -> 'Vector':
        try:
            return Vector(self.x * other[0], self.y * other[1])      # <<== what is this operation
        except TypeError:
            return Vector(self.x * other, self.y * other)
```
The same for division... (and `__itruediv__ `, `__rtruediv__ ` too)  
```
    def __truediv__(self, other: Union['Vector', Point, float]) -> 'Vector':
        try:
            return Vector(self.x / other[0], self.y / other[1])     <== not sure what that is supposed to do?
        except TypeError:
            return Vector(self.x / other, self.y / other)
```

**Suggestions:**  
I think that a `normalized()` method would be quite useful too, in order to obtain a normal without modifying the object.  
Maybe an `__iter__` method to be able to unpack the x and y values would be useful?   
Maybe rename `magnitude_sq` to `magnitude_square`? (my preference as it is clearer)

MrGallo:
```python
try:
    return Vector(self.x * other[0], self.y * other[1])      # <<== what is this operation
except TypeError:
    return Vector(self.x * other, self.y * other)
```
So you can both multiply a vector by a vector or a single scalar value.
```python
vector1 = Vector(2, 3)
vector2 = Vector(4, 5)
vector1 * vector2  
# code in try: multiply vector by vector.
# returns new Vector(2*4, 3*5) -> Vector(8, 15)
vector1 * 6
# multiply vector by single value.
# returns Vector(2*6, 3*6) -> Vector(12, 18)
```

In the `Vector` class, the `__slots__` allows for unpacking. https://github.com/MrGallo/arcade/blob/a696da6fb4f8deb7baaa097bf7c42a4fdd0db58d/arcade/vector.py#L8 

I don't have much more time to devote to this stuff right now. See where Mr. Craven moved this code and feel free to contribute your changes to it. I don't know where it ended up. :)

SirGnip:
FYI, as I was implementing the particle system code, I put a simple vector class in `utils.py` called `_Vec2` (https://github.com/pvcraven/arcade/blob/master/arcade/utils.py#L66).  When the `Vector` class is fully merged, I'll migrate the particle system code to use it and remove `_Vec2`.  Thanks for doing the work on this!

ReblochonMasque:
vector multiplication is either:  
**scaling:** multiplication by a scalar, returning a vector.    
**dot product:** returning a scalar.  
**cross product** also called *perp product* in the case of 2D, returning a Vector.  
(There is also a different **dot product in minkowski spaces**, but it does not have its place in games)  

definitions here: http://geomalgorithms.com/vector_products.html

The operations: `v * u` and `v / u` where the components are multiplied/divided two by two and produce a `vector(vx * ux, vy * uy)` (or the same, but divided), are not defined, and do not correspond to anything I've ever heard of.

MrGallo:
LOL I'm not much of a Physics guy. Thanks for pointing this out. Go ahead and give it a change to make it conform to actual Physics :) Again, I don't know where this code has been moved to by @pvcraven.

PR #348, opened 2019-04-07
Removed all comments from pyinstaller batch

owlnical:
Hi,

Perhaps there's something I'm not understanding here. But having rem in front of all commands in the pyinstaller bat just makes them comments, right? No commands are actually executed, making the example bat kinda useless. This patch just removes the rem from all lines.

If there's a point to this, feel free to ignore this pull request :)


pvcraven:
avbin isn't currently used. ffmpeg is. So that code needs to be fixed, not just re-enabled.

Deployment needs a fix-up. I think it was created last year during PyCon 2018 (?) and we got part-way but there was no one to finish out the code to get it working.

owlnical:
I see, thanks for the quick reply :)

PR #383, opened 2019-05-01
Add platforming multi-jump support

Abeardevil:
This adds arbitrary jump limits to the platforming physics engine. I created a different function to enable it because we don't want to assume the user wants to allow this functionality. Since it requires the user to update the jump counter themselves, if we default to a value of 1 for allowed_jumps, for example, then can_jump will break for users who don't add a call to increment_jump_counter. This would break existing code.

As written, this should allow existing code to continue working as-is, but it will require a user to call increment_jump_counter when they jump if they want to use this functionality.

### DIFF
--- a/arcade/physics_engines.py
+++ b/arcade/physics_engines.py
@@ -83,6 +83,9 @@ class PhysicsEnginePlatformer:
         self.player_sprite = player_sprite
         self.platforms = platforms
         self.gravity_constant = gravity_constant
+        self.jumps_since_ground = 0
+        self.allowed_jumps = 1
+        self.support_multi_jump = False
 
     def can_jump(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


arcade/physics_engines.py
```diff
@@ -100,10 +103,22 @@ def can_jump(self) -> bool:
...
         else:
             return False
+    
+    def allow_multi_jump(self, allowed_jumps: int):
+        self.allowed_jumps = allowed_jumps
+        self.allow_multi_jump = True
+        
+    def increment_jump_counter(self):
+        """
+        Updates the jump counter for double-jump tracking
```
therk: Do you mean multi-jump tracking?

Abeardevil: Yep I do! Thanks

Abeardevil:
@therk I agree unit tests would be good here! It seems like there isn't a ton of unit testing going on in the physics engine in general, unless I'm missing something. The practice seems to be "set up some state and then run a few frames to make sure it doesn't crash".

I'll add some testing to make sure the double jump logic works as expected.

Abeardevil:
Added unit tests and a function to disable multi-jump

PR #394, opened 2019-05-07
Add support for Sprite.forward/reverse/turn_left/right/strafe/stop

JosiahBradley:
Sprites can now move forwards based on their angle. This included adding support for radian angles.
Also included are methods to move backwards called reverse and a strafing function for lateral
movement perpendicular to the Sprite's angle. Turn left and right are added for those familiar
with turtle graphics, summing the basic functions normally included.

This should resolve Issue 323.


pvcraven:
Thanks!!!

Issue #172, opened 2017-12-25
Request for transforms


jarav:
## Enhancement request:

Support for coordinate transformations such as translations and rotations and general 2D matrix transformations as for example in [Processing](https://processing.org).

### What should be added/changed?
Ability to rotate drawing primitives and sprites about any point and not just about their centers.

### What would it help with?
We will be able to make more interesting animations in an easy way.

jarav:
At the moment, using `rotate_point`, we can manually calculate the vertices of a rectangle and rotate it about any other point but if you want nested transformations it will help if they can be pushed  into and popped off a stack.

pvcraven:
It isn't intuitive, but you can also use a shape_list:

import arcade
    
    SCREEN_WIDTH = 800
    SCREEN_HEIGHT = 600
    
    
    def make_shape():
    
        shape_list = arcade.ShapeElementList()
    
        # Shape center around which we will rotate
        center_x = 20
        center_y = 30
    
        width = 30
        height = 40
    
        shape = arcade.create_ellipse_filled(center_x, center_y, width, height, arcade.color.WHITE)
        shape_list.append(shape)
    
        return shape_list
    
    
    class MyGame(arcade.Window):
        """ Main application class. """
    
        def __init__(self):
            # Call the parent class initializer
            super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, "Shape Demo")
    
            self.shape_list = make_shape()
    
            # This specifies where on the screen the center of the shape will go
            self.shape_list.center_x = SCREEN_WIDTH / 2
            self.shape_list.center_y = SCREEN_HEIGHT / 2
    
            arcade.set_background_color(arcade.color.AMAZON)
    
        def on_draw(self):
            """
            Render the screen.
            """
    
            # This command has to happen before we start drawing
            arcade.start_render()
    
            self.shape_list.draw()
    
        def update(self, delta_time):
            """ Movement and game logic """
            self.shape_list.angle += 1
    
    
    def main():
        window = MyGame()
        window.setup()
        arcade.run()
    
    
    if __name__ == "__main__":
        main()

jarav:
Thank you for that example. With the use of the `ShapeElementList` I was able to make a very simple animation of the earth going around the sun with the moon going around the earth. But I think it would have been easier with transforms. Here is my program:

```python
"""
Earth moon sun
"""

import arcade
   
SCREEN_WIDTH = 600
SCREEN_HEIGHT = 600

[...]

PR #399, opened 2019-05-08
Add generic affine transformation and, translate and rotate as example

JosiahBradley:
For issue #172

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -120,6 +120,48 @@ def rotate_point(x: float, y: float, cx: float, cy: float,
     return x, y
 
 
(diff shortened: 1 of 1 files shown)


JosiahBradley:
This is a WIP and hopefully I can add some more docs/examples around easier to enter matrices so people can learn compound transformations.

pvcraven:
This might be better put in the geometry.py file?

pvcraven:
We could use something like this, but for reasons on speed should probably be done on an OpenGL level, or perhaps C when/if we move to a C base for sprites. That will require a different way of doing things than the pure-python way.

PR #404, opened 2019-05-16
Implement Vector class

MrGallo:
Create `vector.py` Vector class.
Modify `sprite.py`. Create `VectorSprite` class which uses `Vector` for position and velocity
Implement tests for `Vector` class.

Decided to have `VectorSprite` extend `Sprite` because there is a minor trade-off in performance if we implement `Vector`s with the original `Sprite` class.

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -940,6 +940,11 @@ def _get_points_for_points(point_list, size):
         new_point_list.append((x - hs, y - hs))
         new_point_list.append((x + hs, y - hs))
         new_point_list.append((x + hs, y + hs))
+
+        new_point_list.append((x + hs, y + hs))
+        new_point_list.append((x - hs, y - hs))
+        new_point_list.append((x - hs, y + hs))
+
     return new_point_list
 
 
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -369,9 +369,7 @@ def make_circle_texture(diameter: int, color: Color) -> Texture:
     bg_color = (0, 0, 0, 0) # fully transparent
     img = PIL.Image.new("RGBA", (diameter, diameter), bg_color)
     draw = PIL.ImageDraw.Draw(img)
-    radius = int(diameter // 2)
-    center = radius # for readability
-    draw.ellipse((radius, radius, center+radius, center+radius), fill=color)
+    draw.ellipse((0, 0, diameter-1, diameter-1), fill=color)
     name = "{}:{}:{}".format("circle_texture", diameter, color) # name must be unique for caching
     return Texture(name, img)
 
@@ -399,7 +397,7 @@ def make_soft_circle_texture(diameter: int, color: Color, center_alpha: int=255,
     for radius in range(max_radius, 0, -1):
         alpha = int(lerp(center_alpha, outer_alpha, radius/max_radius))
         clr = (color[0], color[1], color[2], alpha)
-        draw.ellipse((center-radius, center-radius, center+radius, center+radius), fill=clr)
+        draw.ellipse((center-radius, center-radius, center+radius-1, center+radius-1), fill=clr)
     name = "{}:{}:{}:{}:{}".format("soft_circle_texture", diameter, color, center_alpha, outer_alpha) # name must be unique for caching
     return Texture(name, img)
 
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -13,7 +13,9 @@ from arcade.draw_commands import load_texture
 from arcade.draw_commands import draw_texture_rectangle
 from arcade.draw_commands import Texture
 from arcade.draw_commands import rotate_point
-from arcade.arcade_types import RGB, Point
+
+from arcade.arcade_types import Point, RGB
+from arcade.vector import Vector
 
 from typing import Sequence
 from typing import Tuple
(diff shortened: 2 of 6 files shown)


MrGallo:
Performs quite poorly on `arcade\examples\perf_test\stress_test_draw_moving_arcade.py`. Almost doubles the processing time. :(

SirGnip:
Would there be any performance benefit to have your Vector class not derive from `list`, but use `__slots__` instead?  I created a simple Vector class (`_Vec2`) in https://github.com/SirGnip/arcade/blob/master/arcade/utils.py for use by the Particle system and used slots (see https://github.com/SirGnip/arcade/blob/master/arcade/utils.py#L71).  I didn't try using my `_Vec2` class in `sprite.py`, so I have no idea how it performs (though using `__slots__` was much faster than not using `__slots__`)

MrGallo:
I edited my comment above concerning `__slots__`. I never tested that version against the `arcade\examples\perf_test\stress_test_draw_moving_arcade.py` test.

SirGnip:
Ah!  Too bad.  A vector class is extremely convenient.

Is it maybe worth having a Vector class in arcade, but not use it for sprite.py (kind of what I did for the particle systems)?  This allows us to have a nice Vector class when needed, but the core of arcade isn't paying the performance tax of it...  Just brainstorming...

MrGallo:
Maybe I should work on a `VectorSprite(arcade.Sprite)` subclass that will make use of an `arcade.Vector` for `position` and `velocity`, should one desire to use it. Thoughts?

pvcraven:
I'm trying to see about implementing SpriteList/Sprite with a C-language base. I want to be faster than PyGame at everything, and I can't right now because these operations are just so much faster in C.

MrGallo:
In the meantime I would like to add the `Vector` class to arcade and extend `Sprite` with a `VectorSprite` class. I'll be modifying the pull request code soon.

sbischoff-ai:
For fast C-based linear algebra in Python the obvious choice would be to just use numpy.
I'd recommend implementing a wrapper around numpy arrays with a more pythonic API for 2D Vectors, Matrices and utils for common geometric transformations.

I could start working on that and on refactoring existing 2D math calculations accordingly.

pvcraven:
You would think numpy would be faster, but it isn't. The VBOs for OpenGL use Numpy and that's the main performance bottleneck the code has.

Numpy is fast when the numbers are in numpy. But going from Python storate to Numpy storage and back is slow. Pyglet and PyGame use native C types are are faster.

My long-term plan (starting around December) is to dive in and make the base class compile in C. This is hard because OpenGL drawing requires all data in a sprite list to be in a continuous block on memory. So sprites have to know where in the block they are. If indeed, they are in a block yet.

astronouth7303:
Hey, Astra from the PursuedPyBear project. We implemented [ppb-vector](https://github.com/ppb/ppb-vector/) already, and I thought I'd weigh in with our experience.

First of all, the golden rule of trying to optimize: Always have benchmarks, always follow the data. Examples:

* `__slots__` is faster than `__dict__`, even with some of the extra calls involved in their implementation. (At the time, I thought it was primarily a memory optimization.)
* Caching is often slower than just recomputing. We originally memoized the vector length, but we found that the memoization code was slower than the floating point operation.

As for C/Python, while yes C versions would be faster, for us we found that PyGame (and SDL1)'s render model was our biggest bottleneck, not our pure-python mechanics. We consider our current pure-python implementation Good Enough for now. Your milage may vary and always approach performance problems with data in-hand. (That said, we did experiment with pyglet and pygame and found even an initial implementation of pyglet-based rendering to be a dramatic improvement. The effort was abandoned because at the time pyglet had crashing bugs on mac, and that was unacceptable for us.)

I had mentally noodled with the idea of a next-gen version of ppb-vector that included bulk transformation features and could juggle vector buffers, but we have not needed it so far.

Even on my pretty-old T420, our performance bottlenecks are rooted in pygame/sdl1, not our code. That said, our goals are explicitly education-focused, which means that we want to focus on ease of use and readability over maximum performance or scale.

(I will also shamelessly plug and say that ppb-vector was spun out into its own project partly in hopes that other people will use it, and you are more than welcome to either vendor it or just use the package. We consider it stable with a pretty decent test suite.)

pvcraven:
Hey, thanks for the reference!

There are a lot of attempts at structure that I've found slower, and I agree that 'follow the data' is the way to go.

PyGame's drawing is super-slow. But its collision detection, in pure C, is really fast. We do spatial hashing in pure python and PyGame's C code with a straight n^2 order still performs just as well.

The time it takes to move between local storage and a block of memory set up for OpenGL rendering is a bottleneck for me. I haven't had time to try __slots__, but it is on my radar.

This is an area of the code I don't want to do changes in until it is both faster, and programmatically better.

astronouth7303:
I would not _expect_ moving to `__slots__` to make the copy to an OpenGL buffer somewhat faster, because it improves access times.

That said, I would _expect_ real improvements to come from more aggressive and systemic management of vector buffers. (eg, user-visible vector objects are allocated out of pre-sized vector buffers, and you write code to manage the alloc/dealloc lifecycle of those buffers and their contents.) Unfortunately, I'm only vaguely familiar with OpenGL mechanics and don't really know the specifics of operation performance.

pvcraven:
At this point, I've done the optimization pass that I wanted to. This was the reason for delaying this PR. Not sure if we'd be better off with our own vector, use something like ppb-vector? Not do anything? I'd like  to move away from numpy as a dependency, so I'd prefer not to use  that.

Performance has to be in-line.

MrGallo:
I don't mind if this gets dropped as long as there is the option to refer to position and speed as vectors and use vector methods on them.

pvcraven:
Closing this. We need a new PR that builds without conflict and runs as fast as current implementation.

PR #413, opened 2019-05-25
Add links to PyCon 2019 YouTube vid

SirGnip:
Also fixed a warming with the Kayzee doc


SirGnip:
Cancelling this PR because I unintentionally pulled in other commits

Issue #415, opened 2019-05-25
Tiled map is drawn shifted one column to the left


SirGnip:
### Actual behavior:
When I create a map with Tiled and draw it on the screen without any offsets/scrolling/scaling, the map is offset one column to the left, meaning the left-most column of tiles (stack of boxes in this case) is off screen:

![image](https://user-images.githubusercontent.com/5678175/58370214-8ff16900-7ec9-11e9-8318-8d0df78fccd1.png)

### Expected behavior:
When drawing a map in Arcade (without any offset/scrolling/scaling), I expect the origin of the map to be the bottom-left corner of the Arcade window.  The origin of the map should match the origin of the window.

![image](https://user-images.githubusercontent.com/5678175/58370218-9da6ee80-7ec9-11e9-98e3-c73f20db88d0.png)

### Steps to reproduce/example code:
Here is the example code I used to generate the "Actual behavior" screenshot.  Uncomment the loop to see the expected results.

```
import arcade

class MyGame(arcade.Window):
    def __init__(self):
        super().__init__(600, 600, "Tiled Map Off-by-one Repro")
        arcade.set_background_color(arcade.csscolor.CORNFLOWER_BLUE)

    def setup(self):
        map = arcade.read_tiled_map('map.tmx', 1.0)
        self.wall_list = arcade.generate_sprites(map, 'Platforms', 1.0)
        # for s in self.wall_list.sprite_list:
        #     s.center_x += 128

    def on_draw(self):
        arcade.start_render()
        self.wall_list.draw()

    def update(self, delta_time):
        self.wall_list.update()

def main():
    window = MyGame()
    window.setup()
    arcade.run()

if __name__ == "__main__":
    main()
```

pvcraven:
Question, should the lower left sprite be at 0, 0 for a center, or 0, 0 for the lower left corner of the sprite?

pvcraven:
Closing. Not breaking backwards compatibility. Please track with Issue #362 on where the origin should be for the new tmx loader.

Issue #362, opened 2019-04-21
Improve Tiled Support


Beefy-Swain:
## Enhancement request:
Support for Tiled is poor. Arcade does not support single file tsx, it does not support non-polyline hitboxes, the functionality is largely un-documented, and the code has few comments, and it is poorly tested.

### What should be added/changed?
- External tileset file (tsx) support, including multiple tsx files
- Tilesheets
- all supported hitbox types
- add a dict with custom properties to each sprite(?) object
- anything else that tiled supports that is unsupported
- more tests for Tiled related functionality

### What would it help with?
using Tiled with Arcade

---

I would be happy to make the requisite changes, as I think that I am (hopefully) qualified to do so, but I have 2 concerns:

1. The kind of changes that are likely necessary would be breaking changes. I don't know how that is handled in this project
1. I don't know how to use pytest or whatever we do for testing.

I will get started on adding the functionality I can think of, and hopefully someone else can either tell me how to test it, or can write the tests for it.

Thanks!

Beefy-Swain:
Thought I had for concern 2:

I will create another file called tiled.py, and if my changes are accepted then read_tiled_map.py can be deprecated with a warning, then eventually removed.

Beefy-Swain:
I'm really struggling here to figure out how to test this code effectively. How do I say, "if the pixel at (x,y) is value foo, fail" for instance.

Beefy-Swain:
It seems like `pyglet.image.get_buffer_manager().get_color_buffer().save('screenshot.png')` could maybe do what I want. Any idea how to get that working?

pauleveritt:
I can help. If you can give me a couple of code paths as examples, I can suggest a strategy. I admit, I know exactly zero about the Arcade code under discussion.

First, for "unit testing", you can use monkeypatch or mocks to replace the "rest of the system". Then see if those parts were called with the right values.

Or, if the tiled system keeps some state, you can reset it between tests and, during a test, assert if its "model" has the correct info.

Or, start by writing tests for the existing reported bugs for Tiled. Helps get a test base in place.

I will start by reading your comments in #360

Beefy-Swain:
@pauleveritt I appreciate the help!

[...]

PR #416, opened 2019-05-25
Fix tiled map drawing so the map's origin matches the window's origin…

SirGnip:
#415


pvcraven:
Changing the positioning of the sprites will break current code, so I don't see doing that.

There is a branch where we are creating a different function to load maps, with full tmx support. I'm ok with that having different positioning.

pvcraven:
Closing. Not breaking backwards compatibility. Please track with Issue #362 on where the origin should be for the new tmx loader.

PR #420, opened 2019-05-30
By default, have bottom-left corner of bottom-left tile be at the window's origin 

SirGnip:


pvcraven:
Thanks! This is correct, but would break backwards compatibility of code. Plus new code is now in Tilemap.

SirGnip:
Understood.  

Is there a plan for how to take these little rough edges and fix them?  If we strive to maintain 100% backwards compatibility over all time, arcade will collect lots of these little warts and become difficult to understand (see "software entropy").  We could maintain "compatability" within major version numbers, but say that major version number bumps (ex: v2.0 to v3.0) will contain changes that break backwards compatibility (see [Semantic Versioning](https://semver.org/)).  Granted, it requires more effort to maintain multiple branches, plan releases, communicate breaking changes to user community, etc.

Could issues like this at least be tagged with something like "compatability-breaking-change" so that they can be found and fixed later if/when arcade decides to do a "backwards incompatible" release?

pvcraven:
Well, there's the second version of the tilemap reader, and PR #486. That hasn't been out long to prevent a fix.

PR #428, opened 2019-07-09
Sprite default texture [Food for thought]

MrGallo:
I was messing around with a new concept for the `Sprite` class. The idea came when I noticed that the  `Sprite` class wasn't useful without access to an image file. In the event a Sprite is created without an image, I wanted the `Sprite` class, by default, to have a simple rectangle texture created, either with a supplied color, or with a random color. This beats having arcade constantly make calls to `arcade.draw_rectangle`, which causes the program to slow down significantly.  With this "default texture" idea, came significant changes to the `Sprite` class we might want to discuss.

- Abstract out cropping images when creating sprites.
  Removed `image_x`, `image_y`, `image_width`, and `image_height`
  arguments, as they are for cropping.
  Cropping should be done with `load_image()` and the resulting
  `Texture` object should be given to a `Sprite`.

- Sprites now accept Textures (as well as the traditional file name) upon creation.
  Changed Sprite kwarg `filename` to `image`, to be more general
  when accepting both file names and `Texture` objects. We could 
  still allow a `filename` key-word argument so code that
  explicitly uses it when creating a sprite doesn't break.

- `width` and `height` properties allow more precision of the size
  of the Sprite. Can be used in conjunction with or independently
  from the scale property.

- Added `Sprite.debug` property which switches out the current texture for a black box with green border and croshair in `center_x`, `center_y` for better visualization. Especially if the original image is not cropped properly and is causing far-off collisions. The aesthetic of this could be better implemented with a translucent black overlay on the original texture with the bounding box and cross-hair. What I have now is just a prototype.

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -173,7 +173,7 @@ class Texture:
         from arcade.sprite_list import SpriteList
 
         if self._sprite is None:
-            self._sprite = Sprite()
+            self._sprite = Sprite(width=width, height=height)
             self._sprite._texture = self
             self._sprite.textures = [self]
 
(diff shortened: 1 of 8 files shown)


pvcraven:
These are good ideas. I've thought of creating a v2 sprite class, leaving the v1 for backwards compatibility. Main hang-up for me is that I'd like to integrate a native C codebase for some operations. And that is turning out to be really hard.

MrGallo:
I can live with the way things are now by creating a sprite without a texture, then using `make_soft_square_texture`, then assigning that square as the sprite's texture.

```python
player = arcade.Sprite(center_x=200, center_y=100)
player.texture = arcade.make_soft_square_texture(50, arcade.color.BLUE, 255, 255)
```

I teach high school kids so anything we can do to simplify the process would be better. It would be nice to do the following and it automatically assign a texture with a random color:

```python
player = arcade.Sprite(center_x=200, center_y=100, width=30, height=40)  # random color

# Or explicitly state what color you want.

player = arcade.Sprite(center_x=200,
                       center_y=100,
                       width=30,
                       height=40,
                       color=arcade.color.BLUE)
```

pvcraven:
Is this done in a way that is backwards-compatible? I haven't had a chance to dive in and see. I like the idea of being able to create a solid-color sprite.

MrGallo:
The current parameters for `Sprite` that needed to be removed are below. These parameters are for cropping a `Sprite`'s image when initializing it. I don't know if 1) many people actually use these parameters and 2) if it is a good idea at all to give cropping arguments when initializing a `Sprite` (In my opinion). https://github.com/pvcraven/arcade/blob/07f8afc1d780c2d2cb7fdf158df31a2d084990c3/arcade/sprite.py#L93-L96

pvcraven:
The idea isn't so much cropping, but pulling individual sprites out of a sprite sheet. I'm not sure how much they are used.

MrGallo:
The **`filename`** keyword argument was changed to `image`. That means five keyword arguments were changed and will potentially break backwards-compatibility. We can leave them in for now and trigger deprecation warnings when they are used. 

Keyword args to trigger deprecation warning:
- image_x
- image_y
- image_width
- image_height
- filename

That should cause some bloat, but, preserve backward-compatibility with the goal of removing them completely in the future (if desired).

pvcraven:
Closing this PR. With the addition of build-in resources and the SolidSprite class, I think most of the functionality here should be covered.

PR #435, opened 2019-07-23
[WIP] New concept simplifying screen switching

### LINKED ISSUE #369: Add a better "how to" on putting an instruction screen in a game
We have an example for how people can add an "instruction" screen to their game. Unfortunately the current instructions are too complicated, particularly if only one page is needed. Even if multiple pages of instructions are wanted, it is still too hard to follow for many students.

### LINKED ISSUE #374: Add a better "how to" on putting a "game over" in a game

MrGallo:
Switching between screens gets a little complex with all the if statements in each window method (`on_draw`, `update`, `on_mouse_press`, etc...).

Implemented a `View` class. The `Window` class now has a `.show_view()` method and a `.current_view` field (de34215).

`window.show_view(myview)` will remove previous view's event handlers and push the event handlers of the new `View` object being shown. Have a look at the `different_screens_example.py` and `different_screens_pause.py`  to see how one would go about using `window.show_view()`.

Inspired by @SirGnip 's e0768ab and Pyglet's example program [*Astrea*](https://github.com/adamlwgriffiths/Pyglet/blob/master/examples/astraea/astraea.py#L653).

TODO:
- Complete some tests for the functionality
- Documentation

Resolves #374 
Resolves #369


wamiqurrehman093:
Woah nice its gonna make my life easier <3

MrGallo:
**Important**: We might need to change the use of the word *screen* because pyglet keeps track of all hardware screens connected to the computer.

See: [Pyglet Screens](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/modules/canvas.html#pyglet.canvas.Screen)

SirGnip:
Excellent idea!  I've used a similar approach in games of mine before.  It has really helped out.

You asked about "variable context". I think this is an issue with your current implementation because you are only rerouting the Window's *methods* to another's class's methods. In object oriented programming, methods and data are grouped together in a class.  So, we need a way to route both the methods *and* the data.  This is handled by the "state" design pattern (https://en.wikipedia.org/wiki/State_pattern). In the "state" design pattern an instance of a
class (not just its methods) are used as a "state".  A snippet is worth a thousand words:


```
# UNTESTED PSEUDOCODE...
class Window(pyglet.window.Window):
    def __init__(self, ...)
        ...
        self.current_state = None
        self.user_choice = None 

    def update(self, delta_time):
        ...
        if self.current_state is not None:
            self.current_state(delta_time)
    
    def on_draw(self):
        ...
        if self.current_state is not None:
            self.current_state.on_draw()
    
    # call the corresponding current_state's method for all overrideable methods (on_key_press, etc.)    
            
            
class State1:
    def __init__(self, parent):
        self.parent = parent
        self.state1_sprites = SpriteList()
        
    def update(self, delta_time):
        self.state1_sprites.update()

    def on_draw(self):
        self.state1_sprites.draw()
                
    def on_key_press(self, key, modifiers):
        if key == arcade.key.Y:
            self.parent.user_choice = 'Yes'
            self.parent.current_state = State2(self.parent)
        elif key == arcade.key.N:
            self.parent.user_choice = 'No'
            self.parent.current_state = State2(self.parent)

        
class State2:
    def __init__(self, parent):
        self.parent = parent
        self.state2_sprites = SpriteList()

    def update(self, delta_time):
        if self.parent.user_choice == 'Yes':
            # do stuff
        else:
            # do other stuff 

```    
    
Things to notice:

* Each state has its own methods AND its own data.  The "State" classes encapsulate the data and behavior of each game state.  This helps keep your logic and assets for your title screen separate from your scoreboard screen.
* Any "shared" data can be stored on Window and accessed by the state's self.parent reference (see `self.user_choice` in the example above).
    * This can be used to "pass" data between states.
* If you wanted to get fancy, you could also test to see if a certain method existed on a state object before calling it to allow "optional" method overrides, etc.

I'll see if I can put an example together and push up a PR for reference...

SirGnip:
In your TODO, you also mentioned that doing a "pause screen" with your original implementation wouldn't work because the code creates a new `WhateverScreen` object every time `arcade.set_screen(WhateverScreen)` is called.  This would wipe out your game state every time the game was restored from pause.

With an approach similar to my snippet above, the developer writing the game code gets to decide when to create a screen object (or "state"), meaning that screens can hold onto their data even while another screen is active.

```
class MyGame(arcade.Window):
    def setup(self):
	self.gameplay_screen = GameplayScreen(self) # only one instance of this created
        self.current_state = self.gameplay_screen

    def pause_game(self):
        self.current_state = PausedScreen(self)
        # GameplayScreen object still exists even though it isn't being updated/drawn (as it isn't the current_state)

    def unpause_game(self):
        # Restore original GameplayScreen object to be the "current_state" so it is again updated/drawn.
        # No new object has been created, so GameplayScreen continues with its previous data.
        self.current_state = self.gameplay_screen
        
class GameplayScreen:
    ...

class PausedScreen:
    ...

```
With this approach, it is up to the developer when the screen objects get created/reused. The developer could allow some screens to hold their data while other screens are active (like in snippet above).  Or, the developer could create a new instance of a screen each time it is needed (like the snippet below):

```
class MyGame(arcade.Window):
    def next_level(self):
        # Previous instance of GameplayScreen stored in .current_state is lost
        self.current_state = GameplayScreen(self)
```

This gives flexibility to the developer for them to determine the lifetime of their screen objects (either reuse existing instances or re-create fresh instances each time the screen changes).  The arcade code simply keeps track of `.current_state` and calls all the methods on `.current_state` if it exists.  Simple and flexible...

MrGallo:
@SirGnip thanks for the awesome feedback. I can’t get to a computer today to mess around. I’ll brainstorm over all this in the meantime. 

The trick is to balance the ability for beginner devs (students) to use it without much strain and yet have the more complex things available if needed.

MrGallo:
Concerning the naming of what we are discussing and the fact that piglet “screens” are already a thing, maybe we can call what we are talking about “views”. 

MenuView, GameView, InstructionsView...?

Or does it even matter is we keep the word “screen”?

MrGallo:
I implemented a screen registry as a way to preserve the state of each screen. There is also a `reset` parameter that will restart the screen. With this I was able to implement the pause screen without losing game state. Check out the example `different_screens_pause.py`. Because of this I had to change the way each screen sets a background color.

I am wondering if **1)** we can have a function that will send data in the form of a dictionary between screens like `arcade.send_data_to_screen(GameScreen, {'foo': 1, 'bar': 2})`. Or **2)** just use global variables for shared data.

On the other hand, I think @SirGnip 's idea about state pattern is still worth looking into.

SirGnip:
Apologies that this post got tl;dr...

I agree that it is important to keep things easy for beginners.  So, I've pushed up an implementation of the "State" pattern that mimics very closely what you have currently implemented, @MrGallo.  But, behind the scenes, it uses the State pattern.  

So, with my commit above it only requires the developer to call .next_state(MyScreenClass) to switch screens.  This method also uses a screen registry and `reset=` argument, just like your implementation.  Developers create their Window class like in any other arcade app, initialize things in the `__init__` method, and they need to derive their Window class from from `arcade.WindowWithStates` instead of `arcade.Window`.  Each individual State is derived from `arcade.WindowState` and looks nearly identical to your Screen classes.  One tweak is that there is a `begin()` method that is called automatically each time a State object becomes active.  This is how I solved allowing States to set their own background color.

You can see two simple implementations in 
 * `examples/instruction_and_game_over_screens.py`:  
https://github.com/SirGnip/arcade/commit/1c9af523bc6f9f003f031a70469e5176f8c990a0#diff-343e2452fcfc39d6f26bc2c8141f1445
* `examples/diferent_screens_pause.py`: https://github.com/SirGnip/arcade/commit/1c9af523bc6f9f003f031a70469e5176f8c990a0#diff-5efacbdd3cb122ac1110ccd0ea9d8842

Some of the advantages to beginning developers:
* Data that needs to be shared between states can easily be stored on the Window object and accessed in each State via `self.parent`.  Don't have to resort to global variables or custom functions to pass data around.  (see how `time_taken` is handled in the example above)
* Devs have a `startup()` method on their State they can override to initialize data each time a State is created.  They also have a `begin()` method they can overload to add code that needs to run each time a state becomes active.  This removes the need for the special-cased `background_color` class variable just to get screens to have different background colors.

Some advantages of the implementation for arcade maintainers:
* The State implementation is more straight forward to my eyes.  Your method of copying methods from one class to another is clever (I've not seen that done before!  Very slick!)  It seems like it could be brittle and solves some issues (routing of methods) but not others (how to handle shared data).  

Disadvantages to beginning developers:
* requires devs to have a Window class and separate State classes (and derive them from `WindowWithStates` and `WindowState`)

Advantages for advanced developers:
* The State implementation can be made simple for the beginning devs (could probably even provide more beginner-focused simplifications with a bit more time).  But, the State implementation also provides a lot of flexibility for more advanced devs.  In addition to the `WindowWithStates` class having a `next_state()` method (targeted at beginners), the it also has a `set_state()` method that allows devs to set whatever state object they want.  They can create, cache, manipulate, and switch states however they wish.  I created a simple example at `examples/different_screens_pause_layered.py` https://github.com/SirGnip/arcade/commit/1c9af523bc6f9f003f031a70469e5176f8c990a0#diff-c02481f7853f091b86473d3299a79ac7 that shows how to use the State system to get effects like visually layering partially-transparent pause screens on top of gameplay screens.  It wouldn't be hard to see how the State classes could be arranged to provide for a hierarchy of States.

(Great discussion!)

MrGallo:
@SirGnip I checked out your layered pause example. Looks beautiful. The code is also well done. I'm going to mess around a bit more and incorporate some things from yours I think are a good idea and I'll see how it comes out.

MrGallo:
I believe @SirGnip and I are converging quite nicely on this idea.

I created a `View` class that inherits the structure of a window. When you create a `View` object, you have the option to pass in a parent `View` object, so you can refer to the parent within a given view. This will allow passing data back and forth between views and creating a call stack.

When you want to 'show' a `View` object, you use its `.show()` method. This will trigger the view object's `on_show` method.

```python
game_view = GameView()
game_view.show()
```

This style of view switching was inspired by [a post I saw about PyQt5](https://stackoverflow.com/a/47689610).

SirGnip:
Hoping to have time in the next day or two to go through this...  I'll report back.

SirGnip:
Looking really good, @MrGallo!  I love the cleanup you've done, the choice of the name `View`, the use and naming of the `show()` and `on_show()` methods, the `.parent` member, removal of the screen registry, etc!  I agree, things are converging nicely!

I have one remaining concern regarding the event-routing implementation.  It currently does what it needs to do, but it is tricky to understand and bends some object oriented design principles:

- With the current implementation, there is a `Window` instance that has some "bound methods" that are bound to the `Window` instance and other "bound methods" that are bound to a different `View` instance (based on current `View`). To my eyes, this is hard to understand and would make any future debugging or enhancements more difficult.
- In arcade, each instance of the `Window` class represents a physical window on the screen. `View` derives from `Window`, so a developer will expect it to behave like a `Window` (with some added functionality).  But, when a developer uses a `View`, a new window is not created but the `View` represents different behaviors and content inside a single window that already exists.  This is confusing and violates the [Liskov Substitution Principle](https://stackify.com/solid-design-liskov-substitution-principle/).
- Also, related to the point above, a `View` derives from `Window`, but isn't a fully functioning `Window`.  This is partially because `View` doesn't call `super().__init__()`. And, when I tried adding `super().__init__` to `View`, the examples started throwing errors.  

A couple of the issues above could probably be fixed, but these issues are a "code smell" of using a design that isn't quite suited to the problem at hand (routing events, in this example).  I believe that all of the features currently supported by the method-overwriting event handling you coded up could also be supported with a different event-routing design.

So, options going forward as I see it:
- Look at pyglet's `EventDispatcher` class (the class responsible for triggering the event methods on a `Window` and see if it has some feature we could use that would allow us to dynamically switch what object it is routing events to so  how it routes events (ex: to the current `View`)
- Take the event-routing design out of the "State" example that I posted above and use that inside the current View/show()/on_show() implementation of yours (which I really like!)
- Something else?

Thoughts?

To serve as a proposal, I will take your current implementation, rework the design of the event routing to use the "State" design pattern (but keep the View/show()/on_show()/parent naming and implementation) and post it back to this PR.  Might take me a couple days to get to it...

MrGallo:
@SirGnip I'd like to see what you come up with. Thanks for the tip about pyglet and event dispatching. It looks like they already have something that does what I was implementing:

```python
window.pop_handlers()
# and
window.push_handlers()
```

I'll edit my code to use those.

MrGallo:
I set up the `.show()` method to use pyglet's `window.pop_handlers()` and `window.push_handlers()` to override the window's current event handlers. The `View` class no longer needs to inherit from `Window`.

Pouring over [Pyglet's event documentation](https://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/programming_guide/events.html), it seems like they set it up such that you are swapping bound methods between the window and another object whenever you use things like:
- Decorators: `@window.event`
- `window.set_handler(on_draw=some_custom_draw_function)`
- `window.set_handlers(some_object_with_bound_event_handlers)`

The only thing that's ugly (to me) is the fact that `arcade` schedules two update functions, which:
1. need to be unscheduled
2. swapped with other bound methods
3. rescheduled

At the end of the day, I'll be happy if we can get something that works with the right identifiers for classes and methods.

SirGnip:
I like the recent modifications, @MrGallo  I like that View isn't a subclass of Window anymore and that the routing of events is cleaner now (uses built in Pyglet even methods).

I pushed up my proposal (https://github.com/SirGnip/arcade/commit/613602274fe3a76d50c46b56e504fa7f47442657) in the second commit above (this is based very closely to your `add-set-screen`  branch).  The code written by the end developers is almost the same between our proposals.  It is just the implementation in `application.py` that is different. 
 
My Proposed design (rough class layout below...):
```
class WindowWithViews(Window):
    .current_view (of type View)
    .set_view(View)
    ...overload all on_update(), on_draw()... style event methods and route them to .current_view

class View:
    .parent (of type WIndowWithViews)
    .on_show() # meant to be overloaded, called when View becomes active
    .show()  # call to make View the current view
```
With my proposal, each app will normally have one `WindowWithViews` object and multiple `View` objects.  The `WindowWithViews` can only have one View that is current at a time.  All shared data is stored on the `WindowWithViews` object and can be referenced by each `View` by accessing its `.parent` member.  Events aren't routed by changing the Pyglet event routing, but by having all events passed to `WindowWithViews` (like happens with a normal `Window`), and then it routes the events to the current `View`.

I like it that the `WindowWithViews` object is at the "top"... is the "manager" of the `Views`. This is a natural place to store shared data that all `Views` can access.  It records the current state (`WindowWithStates.current_view`) for its own purposes, but it can also be tested by game code to see which `View` is current.  And, as a bonus with this design, you can also create simple overlay effects where a `WindowWithViews` draws its items, and then the current `View` can be layered on top (see the `examples/different_screens_pause.py` demo).

Also, I worked up a simple example (`examples/different_views_multiple_windows.py`) where I opened up two individual WindowWithViews and interact with them independently
where they each maintained their on current View. I was able to get it to work with my proposal but with your proposed code, I was not able to get the windows to work independently (one of the Windows wouldn't change its View when clicked).

Open to any thoughts...

SirGnip:
The commit above (e0768ab) makes a minor tweak so that the game code sets the current view by calling `.show_view(myview)` on the `WindowWithViews` object and not.

This simplifies the work end developers will need to do.  They no longer explicitly pass the "parent" into the `View`'s constructor.  It is handled automatically now.

The design simplifies a bit and becomes:

```
class WindowWithViews(Window):
    .current_view (of type View)
    .set_view(View)
    ...overload all on_update(), on_draw()... style event methods and route them to .current_view

class View:
    .parent (of type WIndowWithViews)
    .on_show() # meant to be overloaded, called when View becomes active
```

dangillet:
Hello,

I haven't read carefully everything you wrote. But what you are implementing seems to exists already in the [Cocos2D](https://github.com/los-cocos/cocos) library which is built on top of pyglet. I'm not saying you should copy-paste straight from there, but maybe you would like to get some inspirations from the following classes: `Director`, `CocosNode`, `Scene` and `Layer`.

Good luck with your implementation and keep up the nice work! :+1:

MrGallo:
@dangillet thanks for the info. Looking over it now.

MrGallo:
*Under construction (Need to step out. I'll edit and add more later)*

@ SirGnip some comments/questions.
> sets the current view by calling .show_view(myview) on the WindowWithViews object and not.

I think `arcade.set_view(myview)` would be a better implementation of this idea if views should not call themselves (like in [PyQt5](https://stackoverflow.com/a/47689610)). However, as you have noticed, this would not be good with multiple windows (comment on that below).

> Also, I worked up a simple example (examples/different_views_multiple_windows.py) where I opened up two individual WindowWithViews and interact with them independently
where they each maintained their on current View. I was able to get it to work with my proposal but with your proposed code, I was not able to get the windows to work independently (one of the Windows wouldn't change its View when clicked).

Without looking into my implementation, I'm willing to concede on the multiple windows feature. I don't know if the idea of arcade has room for multiple windows. I see it as an easy way to help beginner programmers (like my students) make a 2D game with ease. I've never needed a student of mine to open two windows for their games.

> This simplifies the work end developers will need to do. They no longer explicitly pass the "parent" into the `View`'s constructor. It is handled automatically now.

When I looked at the example code, it was a little mysterious how the self.parent field was being populated. I had to look at the source code to understand it. Also, I don't think each view needs to have a reference to a parent. With those two points, I think passing an optional `parent` kwarg when you create a new view object might be the best way to make it both more explicit where self.parent is coming from, as well as allow a view to have no parent.

>Events aren't routed by changing the Pyglet event routing, but by having all events passed to WindowWithViews (like happens with a normal Window), and then it routes the events to the current View.

Pyglet offers a couple ways to override event handlers i.e. `.set_handlers`, `push_handlers`, and `pop_handlers`. Why complicate things and reroute the methods? The *current view* is simply the collection of event handlers that are currently active. When it is time to return control to a previous view, we use `self.parent.show()` or modified, `arcade.set_view(self.parent)`.

## Readability
### Examples from `different_screens_pause.py`. Re-entering game from pause screen.
[your example link](https://github.com/SirGnip/arcade/blob/e0768abcf8afe475166e255fc8c1472362f27365/arcade/examples/different_screens_pause.py#L118-L123)
```python
    def on_key_press(self, key, modifiers):
        if key == arcade.key.ESCAPE:   # resume game
            self.parent.show_view(None) # have just the GameWindow visible with no View on top
        elif key == arcade.key.ENTER:  # reset game
            self.parent.show_view(None) # have just the GameWindow visible with no View on top
            self.parent.on_show()
```

[my example link](https://github.com/MrGallo/arcade/blob/8dda006488fe68639293f45af5d0d821b343506c/arcade/examples/different_screens_pause.py#L123-L128)
```python
    def on_key_press(self, key, modifiers):
        if key == arcade.key.ESCAPE:   # resume game
            self.parent.show()
        elif key == arcade.key.ENTER:  # reset game
            game_view = GameView()
            game_view.show()
```
It seems my implementation is a bit more descriptive. When you press escape, you show the parent. When you hit enter, a new game view is created and shown. In yours, the way it reads is confusing. "When you press escape, the parent window shows a `None` view. When you press enter, the parent shows a `None` view and the parent's on_show method is called". If we modify to have a `arcade.set_view(myview)`, it will look as follows:
```python
    def on_key_press(self, key, modifiers):
        if key == arcade.key.ESCAPE:   # resume game
            arcade.set_view(self.parent)
        elif key == arcade.key.ENTER:  # reset game
            game_view = GameView()
            arcade.set_view(game_view)
```
Which reads as "If you press escape, arcade sets the view to self.parent. When you hit enter, you create a new game view and arcade sets the view to it."

MrGallo:
I made changes to my branch with this commit: https://github.com/MrGallo/arcade/commit/1c59c07d78f9b3d15585ff6407b0952388b0e7de and it is not updating on this PR...

MrGallo:
@SirGnip checkout the latest version. Specifically de34215. I hope you enjoy it.

SirGnip:
Apologies for the long post.  I told myself I wasn't going to do another long post.  But, I'm just really excited about this feature and it gets better with each iteration..  @MrGallo, your current implementation looks really good. I like how the event routing has turned out.  I like `show_view()` being on `Window`.  I am not worried about "global event handlers" anymore.  I feel we are very close...  We continue to converge!

After looking at your recent changes, I was helpful for me to list out the requirements we are after...

Requirements:
1. use Views to separate and encapsulate screen-specific game code
1. use Views to separate and encapsulate screen-specific data
1. View-specific code should be able to call methods on the game's Window object (ex: set_mouse_visible(), set_viewport(), etc.)
1. design should be focused on the beginner experience: simple things should be very simple and obvious by looking at an example
1. one View should be able to pass data to the next View
1. all Views should have access to a set of shared data
1. (optional) whatever implementation is chosen should not get in the way of those at the hobbyist level who want to use `Views` in more advanced ways (layering Views, etc)
1. (optional) Have event handlers that are global for the entire game (ex: want to be able to press ESC at any time to exit the game, etc) (easy to implement manually by having View explicitly call other methods on `Window`)

A couple comments:
- I have to retract on the multi-window feature.  It is not important.  It appears that if you put more than just trivial graphics in a window when you have two Windows open,
you will start to get OpenGL errors.  So, it appears that arcade doesn't current support this (though pyglet does and arcade may in the future?).
- In the two main examples you posted: `different_screens_example.py` & `different_screens_pause.py`, you are using `.parent` to either store a reference to the `Window` or a reference to another `View` depending on what you are trying to do (get access to shared data or pass data from one view to another).  This to me would seem confusing for a beginner, not knowing what type of object `.parent` holds.
- You use the a global variable `window` in the two examples.  Using globals isn't great practice and often makes things more difficult to read once you get past simple examples.  Not something I'd want to model for beginners.

Regrading requirement #3, I was wondering if it would make the beginner experience better if we got rid of `View.parent`.  We are using "parent" for two different purposes... to store a `Window` in some cases and a `View` in other cases. It is confusing.  And "parent" is an ambiguous name.  So, what if we remove `.parent` and replace it with something called `View.window` that stores the `Window` the `View` is being shown by?  This makes it self-documenting what that member is.  That satisfies requirements #3 & #6.  But, it means `.parent` no longer helps with #5.  So, for #5, why don't we simply get out of the way and let the developer use standard Python features to pass data into Views?  Something like this:

```
class Level1View:
	def __init__(self):
		self.level_score = 0
		
	def on_key_press(self, ...):
		if key == ESC:
			view2 = Level2View()
			view2.level1_score = self.level_score  # store the Level1 score on the Level2 View object
			self.parent.show_view(view2)

def Level2View:
	def on_draw(...):
		# Do work on "self.level1_score" in Level2View code...
```

Or, if you want to get more formal, you could create a method like `Level2View.__init__(self, level1_score)` and store the value that way (see how the `GameView` passes itself to the `PauseView` in `different_screens_pause.py` for an example of this).

By making these changes, we:
- get rid of `.parent` which is ambiguous and replace it with `.window` (satisfies requirements #3 & #6 and also simplifies `View.__init__` and anything that overloads it because we don't have to pass around `parent`)
- get rid of the `window` global variable and have `View` code use `self.window` instead (satisfies requirement #3)
- Have the developer set values on class objects explicitly (or use more formal __init__ methods) to pass whatever data between `View`'s (satisfies requirement #5).  This lets users pass around whatever data they need to and name it however makes sense to them.

@MrGallo, I pushed up a PR (bdbd44a) based on your add-set-screen branch with the changes I listed above. Notice in `different_screens_example.py` I pass data between `View`'s in two different ways:
- First, the code in the `GameView`'s `update()` method creates a `GameOverView` object and then stores its time_taken value on it like:  `game_over_view.time_taken = self.time_taken`(demonstrates requirement #5)
- Secondly, there is a shared variable named `total_score` that was added to the `window` object just after it was created in `main()`.  Then, all Views can access this shared variable with `self.window.total_score`. (demonstrates requirement #6).

Let me know what you think.

SirGnip:
Oh, and thanks @dangillet for the reference to Cocos2D.  I skimmed the docs for the classes you mentioned. Nice, powerful system!  But, one of the drawbacks is that to get anything on the screen, you have to deal with a `Director`, `Scene` and `Layer` objects, even if your app doesn't use multiple `Scene`'s or `Layer`'s.  I'd say many of the apps written in arcade would only ever use a single `Scene` (what we call `View`) and almost all of them would not need a `Layer` and so I don't want to impose that extra complexity (boilerplate/setup) on all apps even if they aren't using the feature.  I think the implementation @MrGallo is working on is "leaner" than Cocos2D  If the developer doesn't want multiple `View`'s, they can completely ignore it and be fine.  It mat not be as flexible, but it lets the user get 60% of the features (multiple screens) with 5% of the effort (just overload a `View` class a couple times and set it as the current `View`).

MrGallo:
> @MrGallo, I pushed up a PR (bdbd44a) based on your add-set-screen branch with the changes I listed above.

@SirGnip Thanks! Merged.

SirGnip:
I put together another `View` example (38ef8a3) that demonstrates a very minimal use of the `View` class to get beginners started as gently as possible.

I think this is looking great, @MrGallo!

pvcraven:
This looks a lot cleaner, nice work. Give me a heads-up when you think it is ready to merge.

SirGnip:
I think it is ready to roll.
I'm tinkering around with writing some unit tests for this (possibly mocking out bits of the `Window` class to make it testable without having to open up a window).  But, I'll do that under another PR.

PR #440, opened 2019-08-02
Undefined name; _() in text_loc_example_done.py

cclauss:
$ __python3 -c "print(_('This is a translatable string.'))"__  # --> NameError: name '_' is not defined

Follows the example usage at https://docs.python.org/3/library/gettext.html#gettext.gettext


cclauss:
This change was already made on master.

Issue #459, opened 2019-08-13
Refactor view_test.py to use Pytest syntax


SirGnip:
Also, added @MrGallo's extra error checking to `Window.show_view()` and some more unit tests (originally proposed here: https://github.com/SirGnip/arcade/pull/1)

pvcraven:
Closing.

PR #460, opened 2019-08-13
Refactor view_test.py to use Pytest syntax #459

SirGnip:


pvcraven:
Hm, TravisCI is unhappy.

pvcraven:
Looks like Pyglet 1.4.2 was released yesterday and messed up the mac builds.

Issue #461, opened 2019-08-13
Remove OpenGL dependency from test_view.py so it can be run via CI


SirGnip:

SirGnip:
Not worth the effort.  Leaving `test_view.py` in `unit2/`

PR #462, opened 2019-08-13
Remove OpenGL dependency from test_view.py so it can run via Travis CI #461

SirGnip:


SirGnip:
CI is failing.  Should be able to fix this.

SirGnip:
A couple attempts to mock out Window and move it to unit/ and no success...  Unless someone feels strongly, I'm going to abandon this.

pvcraven:
The CI is nice, but because of things like this I've just concentrated on running things local.

SirGnip:
Not worth the effort.  Leaving `test_view.py` in `unit2/`

PR #474, opened 2019-08-28
require dataclasses on python 3.6

### LINKED ISSUE #469: ModuleNotFoundError: No module named 'dataclasses'

In the install instructions at http://arcade.academy/installation.html the docs say Arcade requires Python 3.6 or newer however, arcade relies on `dataclasses` which were introduced in python 3.7: https://github.com/pvcraven/arcade/blob/2fb93793ea1d8a11eba75e5f959d8d2eed0abec9/arcade/sprite.py#L10

This means installing on Python 3.6 throws `ModuleNotFoundError: No module named 'dataclasses'`

### Where is it located?

http://arcade.academy/installation.html 

It should either be documented that python 3.7 is required, or https://pypi.org/project/dataclasses/ could be required as a dependency to shim `dataclasses` support for python 3.6

chris48s:
Closes #469
Refs #471

You could also expand the test matrix to test on python 3.6 on a wider variety of platforms, if you wanted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,24 @@
 language: python
-python: 3.7
+python:
+  - '3.6'
+  - '3.7'
 services: xvfb
 matrix:
   include:
     - name: "Python 3.7 on xenial"
+      python: '3.7'
       dist: xenial
     - name: "Python 3.7 on bionic"
+      python: '3.7'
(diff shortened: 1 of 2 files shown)


pvcraven:
Thanks!

PR #481, opened 2019-09-13
Camera Class and Objects

wamiqurrehman093:
I have added a Camera Class which takes these parameters entity, screen_size, viewport_margin=None and only get called in on_update method like self.camera.update() then it handles the scrolling around that entity which can be players or enemies or any object that we want camera to follow.


cclauss:
Please fix conflicts.

pvcraven:
Closing, too many conflicts. Feel free to re-open with conflicts resolved.

PR #521, opened 2019-11-24
Added fonc_color arg to gui.TextButton init

lubie0kasztanki:
Added possibility to create custom TextButtons with different font colors faster.
Behavior added:
If the user passes theme argument font_color is taken from the theme
If not and user passes font_color argument the color is set to what they passed.
If the don't pass neither the color is set to face_color.


pvcraven:
Thanks!

Issue #535, opened 2019-12-09
Tiles in object layers are not placed properly


JFincher42:
## Bug Report

When processing an object layer of a tilemap, the objects are not placed properly. This causes them to be rendered in the wrong place, possibly making them collide with other tiles on the screen.

The problem appears in `_process_object_layer()', when setting the sprite location. Instead of setting the  `.right` and `.top` properties, it should set the `.left` and `.bottom` properties. This accounts for the zero-index nature of the tilemap, and the y-axis flip between `arcade` and Tiled.

PR with this change forthcoming.

### Actual behavior:

Tiles are placed one tile width to the left and one tile height lower than they should be.

### Expected behavior:

Tiles are placed where they were found in the tile map

### Steps to reproduce/example code:

Read in a tile map with an object layer. Note where tiles are placed when rendered in `arcade` as opposed to where they are placed in Tiled.

pvcraven:
Implemented in release 2.2.0

Issue #529, opened 2019-12-05
_process_object_layer() crashes when adding non-existent properties


JFincher42:
## Bug Report

During the processing of an object layer, `arcade` crashes when executing `my_sprite.properties.update(cur_object.properties)`, with a NoneType is not iterable error

### Actual behavior:

Game crashes when loading second map

### Expected behavior:

No crash

### Steps to reproduce/example code:

The tileset used for my game has been modified between map levels. The first level opens properly - there, the tiles have custom properties added to the tiles on the object layer. For the second layer, custom properties were added to the tiles in the tileset. This may be causing the problem.

### What should be added/changed?

The fix is to add a check `if cur_object.properties is not None:` before calling `my_sprite.properties.update(cur_object.properties)`. PR coming shortly.

pvcraven:
Implemented in release 2.2.0

Issue #527, opened 2019-12-04
get_tile_by_gid() incorrectly assumes tile GID cannot exceed tileset length


JFincher42:
## Bug Report

The `get_tile_by_gid()` function assumes the the GID for a tile found in the map cannot exceed the reported number of tiles in the tileset. This is incorrect. If you edit the tileset in Tiled, the tile ID's may exceed the current count.

This was introduced in #511 as part of the TMX improvement code.

### Actual behavior:

`arcade` reports:

     Warning, couldn't find tile for item ### in layer 'layer_name' in file 'map_name.tmx'.

### Expected behavior:

`arcade` uses the provided GID to read tile info from the provided tileset.

### Steps to reproduce/example code:

Generate a tilemap with a tileset in Tiled.
Edit the tileset heavily - add tiles, then delete some, then add some more.
Create a map with the new map and tileset

## Enhancement request:

A PR with a proposed change will be forthcoming

### What should be added/changed?

Change in tilemap.py

### What would it help with?

Reading TMX and tileset files.

### What documentation needs to change? 

None

pvcraven:
Implemented in release 2.2.0

PR #536, opened 2019-12-09
Issue tiles in object layers not placed properly

JFincher42:
Fix for issue #535, also contains fixes for #527 and #529


pvcraven:
Thanks!

PR #561, opened 2020-01-24
Added extend, insert and remove methods to spritelist.

maarten-dp:
Let me know if something is not as you like it, or if you have doubts about certain parts.

Kind regards,
Maarten


pvcraven:
Thank you!!!

PR #563, opened 2020-01-25
Only use RGB (not alpha) for text cache name.

### LINKED ISSUE #562: Fading text causes slow-down
Fading out text causes slowdowns because the cache name includes the alpha value and it creates a new text sprite every loop for each alpha level.

Suggestion: only include the rgb values in the cache name, and set `text_sprite.alpha` to that new incoming alpha value.

MrGallo:
Also, include the initial alpha value in the text_sprite.
Resolves #562

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -145,7 +145,10 @@ def draw_text(text: str,
     if len(draw_text.cache) > 5000:  # type: ignore # dynamic attribute on function obj
         draw_text.cache = {}  # type: ignore # dynamic attribute on function obj
 
-    key = f"{text}{color}{font_size}{width}{align}{font_name}{bold}{italic}"
+    r, g, b, alpha = get_four_byte_color(color)
+    cache_color = f"{r}{g}{b}"
+
+    key = f"{text}{cache_color}{font_size}{width}{align}{font_name}{bold}{italic}"
     if key in draw_text.cache:  # type: ignore # dynamic attribute on function obj
         label = draw_text.cache[key]  # type: ignore # dynamic attribute on function obj
         text_sprite = label.text_sprite_list[0]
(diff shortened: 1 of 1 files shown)


MrGallo:
When run on my computer, here are the times using the code below.
## Original
### Pre-fade
23 ms/frame
### Fade
43 ms/frame
## "fix"
### Pre-fade
30 ms/frame
## fade
30 ms/frame

This is drawing 130 pieces of text. There is a 0.06 ms increase per `draw_text` call with the 'fix'.

```python
from time import perf_counter as pc

import arcade


class MyGame(arcade.Window):
    def setup(self) -> None:
        arcade.set_background_color(arcade.color.BLACK)
        # self.sprite = arcade.Sprite(":resources:/images/animated_characters/male_adventurer/maleAdventurer_idle.png")
        # self.sprite.position = 0, 100
        # self.sprite.change_x = 5

        self.elapsed_time = 0

        self.text_alpha = 255

    def on_draw(self) -> None:
        arcade.start_render()
        # self.sprite.draw()

        # Draw 130 texts
        new_color = arcade.make_transparent_color(arcade.color.WHITE, self.text_alpha)
        start = pc()
        for x, y in [(x, y) for x in range(0, 640, 50) for y in range(0, 480, 50)]:
            arcade.draw_text("hello", x, y, new_color, anchor_x="left", font_size=20)
        elapsed = pc() - start

        arcade.draw_text(f"Time (ms): {round(elapsed*1000)}", 10, 550, arcade.color.GREEN, font_size=20)

        if self.elapsed_time <= 4:
            arcade.draw_text("Pre-fade", 400, 550, arcade.color.ORANGE, font_size=30)
        else:
            arcade.draw_text("Fading", 400, 550, arcade.color.ORANGE, font_size=30)


    def update(self, deltatime: float) -> None:
        # self.sprite.update()
        self.elapsed_time += deltatime
        if self.elapsed_time > 4:
            self.text_alpha = max(0, self.text_alpha - 1)


if __name__ == "__main__":
    game = MyGame()
    game.setup()
    arcade.run()

```

pvcraven:
Thanks!

PR #565, opened 2020-01-25
Make draw_text return text_sprite

### LINKED ISSUE #564: Feature request: arcade.draw_text() to return text_sprite
I think it might be useful to have `arcade.draw_text()` return the `text_sprite` it creates to have access to  potentially important information like the `width` and `height` of the text. It will also give the ability to check for collisions with other sprites, I guess more importantly, check for collision with the mouse location.

Example usage:
```python
# draw and get the sprite
title_sprite = arcade.draw_text("Menu", width//2, height//2, arcade.color.BLACK, anchor_x="center")
# underline the text
arcade.draw_line(title_sprite.left, title_sprite.bottom, title_sprite.right, title_sprite.bottom, arcade.color.BLACK, line_width=5)
```

MrGallo:
Resolves #564

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -113,7 +113,7 @@ def draw_text(text: str,
               anchor_x: str = "left",
               anchor_y: str = "baseline",
               rotation: float = 0
-              ):
+              ) -> Sprite:
     """
 
     :param str text: Text to draw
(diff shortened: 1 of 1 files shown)


MrGallo:
Unless there is a philosophical reason for not doing so?

PR #576, opened 2020-01-31
Simplify resources

### LINKED ISSUE #574: Simplify reference to resources
## Enhancement request:
It would be cool to have an easier way to reference resources (much like arcade.key, or arcade.color).

For example, do this or something similar:
```python
sprite = arcade.Sprite(arcade.resources.male_person_idle)
```

I'm always having to check the website for the proper paths. Lazy.

MrGallo:
Here is a stab at generating the `__init__.py` file for the resources.

There are some files with the same name. The image conflicts are the exact image anyway. The sound conflicts are just different file extensions of the same sound.

Feel free to mess around with it/give suggestions.

resolves #574


MrGallo:
Also, I don't know if a prefix is required for everything. e.g. `image_male_adventurer_idle`, rather than `male_adventurer_idle`. I would imagine finding sounds in a sea of images might be difficult without it. Maybe only prefix the sounds/tmx map/and gui?

pvcraven:
This is good.

I think the script to create the `__init__` should go in the `util` folder.

The `doc/preprocess_files.py` creates the index page with all the resources. Not quite sure how to lay that page out for easy reference.

MrGallo:
For the resources page layout, I'd suggest maybe having the heading just the parent directory of the asset, remove the underscores and capitalize the words. Then the resources themselves would be the pythonic name, the last part of `arcade.resources.asset_name`. Just mention at the top that the following are preceded with `arcade.resources.`, for example, `arcade.resources.image_female_adventurer_climb0`.

## Female Adventurer
```
image_female_adventurer_climb0     image_female_adventurer_climb1    image_female_adventurer_fall
```

MrGallo:
Sprites are not drawing after I rebase from the current master branch :(

MrGallo:
How is this such a mess now?

PR #581, opened 2020-02-06
Point a sprite rotates around

humansHS:
Hey, hopefully doing this correctly.

Firstly, I also tweaked a couple comments to better describe what the functions do.

For the actual rotation_point I used a float list to represent the displacement from the center_x and center_y. Then whenever the angle changes we rotate_point and set_position.

Almost certainly I have not done the get/set stuff right... I think I need  the underscore in front of get and set (_get and _set). I do not have much experience with this and some functions have it some don't???

I did:
`rotate_x, rotate_y = self.rotation_point`
Before we check to see if we need them, I think this only saves process time if it does have a rotation point (which I feel is the less likely case) so it might be better to not do this?

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -199,16 +199,16 @@ class Sprite:
 
     def _get_position(self) -> Tuple[float, float]:
         """
-        Get the center x coordinate of the sprite.
+        Get the center x and y coordinates of the sprite.
 
         Returns:
-            (width, height)
+            (center_x, center_y)
         """
         return self._position
 
     def _set_position(self, new_value: Tuple[float, float]):
         """
-        Set the center x coordinate of the sprite.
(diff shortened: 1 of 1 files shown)


humansHS:
Also worth noting, I haven't actually tested if this runs, will do that ASAP.

humansHS:
Ok had some issues with properly implementing new_value: List[float] since this declaration was new to me, but I think it is fixed now.

Not sure what the coveralls test is complaining about in my code though.

pvcraven:
I think coveralls just wants some unit tests.

pvcraven:
Ok, doing some unit tests on this code.

If I do this:
``self.character_sprite.set_rotation_point((1, 0))``
 
...with an initial center_x, center_y of (300, 50) I get (-49, 299) as a result. I feel like the correct result should be (301, 49).

Issue #589, opened 2020-02-08
Sprite rotate around a point yeilding unexpected results.


pvcraven:
Ok, doing some unit tests on this code.

If I do this:
self.character_sprite.set_rotation_point((1, 0))

...with an initial center_x, center_y of (300, 50) I get (-49, 299) as a result. I feel like the correct result should be (301, 49).

humansHS:
What is the rotation of the sprite?

I'll do some testing, but that is likely one of two possible issues:
a) The rotation that is passed into rotate_point needs to have negative sign (most likely).
b) The sprite.center_x/y and rotation_point x/y have been passed into rotate_point incorrectly.

pvcraven:
See around line 64:
https://github.com/pvcraven/arcade/blob/master/tests/unit2/test_sprite.py#L64

humansHS:
I believe I may be a moron. In _set_angle when it applies the center translation:

```
            rotate_x, rotate_y = self.rotation_point
            if rotate_x or rotate_y:
                sprite_rotate = rotate_point(self.center_x, self.center_y,
                                             rotate_x, rotate_y, new_value)
                self.set_position(sprite_rotate[0],sprite_rotate[1])
```

It is setting the rotation point as a an absolute POINT not an offset, I think this was because I originally thought I would set a static point not an offset. But for my personal implementation I wanted an offset point but forgot to apply the change to the code! The change should be simple:

```
            rotate_x, rotate_y = self.rotation_point
            if rotate_x or rotate_y:
                rotate_x += self.center_x
                rotate_y += self.center_y
                sprite_rotate = rotate_point(self.center_x, self.center_y,
                                             rotate_x, rotate_y, new_value)
                self.set_position(sprite_rotate[0],sprite_rotate[1])
```

I will create a pull request ASAP.

A couple of other things:

1) When I put in new code should I also be updating the tests to include unit tests? Is this explained somewhere in the 'contributing' instructions and I skimmed over it?

2) I assume I have done the get and set methods correctly since it appears you have not changed them. My impression is that I only need to do _get or _set when there is extra code to run when changing a property. BUT if that is correct, I believe _set is the option to be used and it is missing some code whereby it should run _set_angle whenever the rotation_point is changed because if it has a non-zero angle of rotation and the rotation_point changes then it will have a new position.

humansHS:
Wait, my solution is still wrong...

humansHS:
[...]

PR #590, opened 2020-02-09
Sprite rotate around a point fixes

humansHS:
In response to Issue #589 

Full explanation found there.

### DIFF
--- a/arcade/examples/view_screens_minimal.py
+++ b/arcade/examples/view_screens_minimal.py
@@ -21,10 +21,6 @@ import arcade
 import os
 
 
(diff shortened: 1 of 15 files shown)


humansHS:
> if an explicit value of **None is allowed, the use of Optional is appropriate**, whether the argument is optional or not.

... and yet:

> arcade/sprite.py:686: error: Incompatible types in assignment (**expression has type "Optional[List[float]]", variable has type "None"**)

I am still learning the type hinting and test running stuff... I do not understand how to fix this :(

PR #597, opened 2020-02-11
Texture transformations

specht:
This pull request introduces a `Texture.draw_transformed` method which draws a rectangular texture like `Texture.draw_sized` but takes a 3x3 2D affine transformation matrix as an additional argument which gets applied to the texture coordinates.

Here's a short [demo video](https://youtu.be/aDKakV0RUvE) which shows what it can do.

It's meant to be used for static rectangles with animated textures which may be useful for parallax scrolling and other effects. A helper class `Matrix3x3` is provided to help with creating the matrices from translation, scale, rotate and shear operations.

A new example, `texture_transform.py` is also included.

Let me know if you need changes or adjustments, I'd love to see this feature merged into Arcade.

PS. This pull request also fixes a bug where `gl.GL_FLOAT_MAT3` was defined in `shader.py` as receiving 6 floating point values instead of the 9 it requires as a 3x3 matrix.


pvcraven:
This is great! Thanks for the contribution.

PR #600, opened 2020-02-13
Issue where setting sprite outer bound coordinates would fail if width/height is different from the texture width/height

maarten-dp:
Doing the following would fail
```python
sprite.width = 100
orig_left = sprite.left
sprite.width = 200
assert sprite.left != orig_left # would fail
```
As the points calculated by `get_adjusted_hit_box()` would always return the texture points.

I just messed around a bit, but I am unsure how this interacts with `scale` (my guess is not good :) ). So I'm looking for your input on the matter. But to start you off I added a MR that would fix my usecase.


arcade/sprite.py
```diff
@@ -39,6 +39,15 @@
 FACE_DOWN = 4
 
 
+def points_by_dimensions(width, height):
```
pvcraven: If we add a function like this:
* Definitely needs type hinting added
* If used internally, it should start with an underscore
* If intended for universal use, should go into geometery.py I think? And have documentation added. Might be able to come up with a more explanitory function name.

, it probably needs to go into

pvcraven:
I'm a bit confused here. If I change the width of the sprite, I would expect its left boundary to change, wouldn't I? The main location of the sprite is always controlled by its center.

I've looked over the code and it hasn't quite hit me on what our end-goal here is.

I do know that hitbox vs. image dimension management is kind of a head-scratcher.

maarten-dp:
> If I change the width of the sprite, I would expect its left boundary to change, wouldn't I?
Indeed, but this wasn't the case. The left bound coordinates remained the same, regardless of the width.

I'll write some example code that you can play with.

arcade/sprite.py
maarten-dp: Yes, I hadn't intended on getting this code merged, I was mostly looking for feedback on the issue, as I'm not sure this solution is the desired one.

maarten-dp:
```python
import arcade

class SquareSprite(arcade.Sprite):

    def __init__(self, *args, **kwargs):
        color = kwargs.pop('color', arcade.color.RED)
        self.size = kwargs.pop('size', 100)

        super().__init__(*args, **kwargs)
        self.texture = self.get_texture(self.size, color)
        self.textures = [
            self.texture,
        ]

    def get_texture(self, size, color):
        return arcade.make_soft_square_texture(size, color, 255, 255)


class Window(arcade.Window):
    def __init__(self):
        super().__init__(500, 500, 'move')
        self.sprites = arcade.SpriteList()
        self.sprite = SquareSprite()
        self.sprite.position = (250, 250)
        self.sprites.append(self.sprite)

    def on_update(self, delta):
        arcade.start_render()
        orig_left = self.sprite.left
        self.sprite.width -= delta * 10
        self.sprite.left = orig_left
        self.sprites.draw()


window = Window()
arcade.run()
```
Without my fix, the left and right bounds of the sprite are slowly going towards the center of the sprite

With my fix, the right bound is slowly going towards the left bound.

pvcraven:
Ok, I get the problem now. 

Looking at the PR I see a couple issues.

First, in geometry.py it looks like we need an `from typing import List`

Second, I created a quick demo program to test. Resizing the width worked, but scaling the sprite caused the hit-box to be too large.
```
"""
Move Sprite With Keyboard

Simple program to show moving a sprite with the keyboard.
The sprite_move_keyboard_better.py example is slightly better
in how it works, but also slightly more complex.

Artwork from http://kenney.nl

If Python and Arcade are installed, this example can be run from the command line with:
python -m arcade.examples.sprite_move_keyboard
"""

import arcade
import os

SPRITE_SCALING = 0.5

SCREEN_WIDTH = 800
SCREEN_HEIGHT = 600
SCREEN_TITLE = "Move Sprite with Keyboard Example"

MOVEMENT_SPEED = 5


class Player(arcade.Sprite):

    def update(self):
        self.center_x += self.change_x
        self.center_y += self.change_y



class MyGame(arcade.Window):
    """
    Main application class.
    """

        """
        Initializer
        """

        # Call the parent class initializer

        # Variables that will hold sprite lists
        self.player_list = None

        # Set up the player info
        self.player_sprite = None

        # Set the background color
        arcade.set_background_color(arcade.color.AMAZON)

    def setup(self):
        """ Set up the game and initialize the variables. """

        # Sprite lists
        self.player_list = arcade.SpriteList()

        # Set up the player
        self.player_sprite = Player(":resources:images/animated_characters/female_person/femalePerson_idle.png", SPRITE_SCALING)
        self.player_sprite.center_x = 50
        self.player_sprite.center_y = 50
        self.player_list.append(self.player_sprite)

    def on_draw(self):
        """
        Render the screen.
        """

        # This command has to happen before we start drawing
        arcade.start_render()

        # Draw all the sprites.
        self.player_list.draw()
        self.player_sprite.draw_hit_box()

        """ Movement and game logic """

        # Call update on all sprites (The sprites don't do much in this
        # example though.)
        self.player_list.update()

    def on_key_press(self, key, modifiers):
        """Called whenever a key is pressed. """

        if key == arcade.key.UP:
            self.player_sprite.change_y = MOVEMENT_SPEED
        elif key == arcade.key.DOWN:
            self.player_sprite.change_y = -MOVEMENT_SPEED
        elif key == arcade.key.LEFT:
            self.player_sprite.change_x = -MOVEMENT_SPEED
        elif key == arcade.key.RIGHT:
            self.player_sprite.change_x = MOVEMENT_SPEED

    def on_key_release(self, key, modifiers):
        """Called when the user releases a key. """

        if key == arcade.key.UP or key == arcade.key.DOWN:
            self.player_sprite.change_y = 0
        elif key == arcade.key.LEFT or key == arcade.key.RIGHT:
            self.player_sprite.change_x = 0
        elif key == arcade.key.A:
            print(f"Before {self.player_sprite.scale} {self.player_sprite.left}")
            self.player_sprite.scale = 2
            print(f"After {self.player_sprite.scale} {self.player_sprite.left}")
        elif key == arcade.key.B:
            print(f"Before {self.player_sprite.scale} {self.player_sprite.left}")
            self.player_sprite.scale = 1
            print(f"After {self.player_sprite.scale} {self.player_sprite.left}")
        elif key == arcade.key.Z:
            self.player_sprite.center_x = 0
            self.player_sprite.center_y = 0
        elif key == arcade.key.X:
            self.player_sprite.left = 0
        elif key == arcade.key.C:
            self.player_sprite.width = 100
            print(f"Before {self.player_sprite.scale} {self.player_sprite.left}")
            self.player_sprite.width = 200
            print(f"After {self.player_sprite.scale} {self.player_sprite.left}")


def main():
    """ Main method """
    window = MyGame(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
    window.setup()
    arcade.run()


if __name__ == "__main__":
    main()
```
temp
temp2

pvcraven:
Thanks for your work on this.

maarten-dp:
The problem is that `scale` is influencing `width`/`height`, but not the other way around. But as soon as I update my `width`, my `scale` attribute is out of date.

Could you tell my what policy you're applying regarding `width`/`height` and its iteraction with `scale`? I always thought they were mutually exclusive as `scale = 2` implied `width * 2` and `height * 2`.

I feel like `scale` is a shorthand for `width`/`height`, like `position` is to `center_x`/`center_y`. But it's not treated like that in the code.

From a design point of view, `scale` and `width`/`height` are conflicting attributes. Manipulating both results in unwanted behaviour, depending on what you set first.

I can fix this, by doing some checks before setting the width/height ratio in  `get_adjusted_hit_box`. But I'm curious to hear your thoughts on this.

maarten-dp:
I did the following things:
- removed the point adaption by scale, and only relying on width/height ratio, as scale is setting width and height, forcing only one point of reference.
- Added an extra test for calculating the adjusted_hit_box when setting the scale.
- Imported List from types in geometry
- Ran your code and validated the above points work as intended.

pvcraven:
Right, it's confusing. I'm not quite sure the right way to manage it. 

I'll take a closer look later. Head's a bit fuzzed right now.

maarten-dp:
Hey!

Is there anything I can do to move this forward?

I'm not sure what your intentions are, so I'm a bit in limbo on this issue. Don't hesitate to tell me you're looking into your own solution. That's perfectly fine for me, I'll just close this PR.

I must admit I haven't followed the devs as much, so apologies if it has already been resolved somewhere else.

Take care!

PR #601, opened 2020-02-23
Fix for Issue #602: Sprite.get_adjusted_hit_box()

Helios713:
For some Sprites the heuristics at the start of the function don't work and get_adjusted_hit_box() would return Nonsense due to a missing line:
Line 315:  x, y = x-self.center_x, y-self.center_y
Without this, it would scale wrong, rotate wrong, and generally return wrong values. 
This has been fixed alongside various cleanup in the surrounding for loop.

Explanation: When a point is not transported to sit on the origin before scaling, it is scaled based on how far from the origin it is, not on how far from its center. Same is true for rotation. The corresponding transportation back existed already, though, so it looks like an oversight, plain and simple. 
This was not discovered earlier because for one, most Sprites have textures, and second, the problem doesn't manifest for the default parameters.

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -310,24 +310,22 @@ class Sprite:
 
         # Adjust the hitbox
         point_list = []
-        for point_idx in range(len(self._points)):
-            # Get the point
-            point = [self._points[point_idx][0], self._points[point_idx][1]]
-
+        for x, y in self._points:
+            # Transport it to around (0,0)
+            x, y = x-self.center_x, y-self.center_y
+            
             # Scale the point
             if self.scale != 1:
-                point[0] *= self.scale
-                point[1] *= self.scale
+                x, y = x*self.scale, y*self.scale
 
             # Rotate the point
             if self.angle:
-                point = rotate_point(point[0], point[1], 0, 0, self.angle)
+                x, y = rotate_point(x, y, 0, 0, self.angle)
 
-            # Offset the point
-            point = [point[0] + self.center_x,
-                     point[1] + self.center_y]
-            point_list.append(point)
-
+            # Transport it back
+            x, y = x+self.center_x, y+self.center_y
+            
+            point_list.append([x,y])
         # Cache the results
         self._point_list_cache = point_list
 


Helios713:
Here is a test for this, it consists of only the part in question and is ready to run as is.
```
import arcade

class Test():
    def __init__(self):
        # setup
        self.scale = 1.0
        self.angle = 0
        self.center_x = 100
        self.center_y = 100
        self._points = [[102, 100], [100, 102], [98, 100], [100, 98]]
        # outputs
        for self.scale in [1.0, 2.0, 0.5]:
            print()
            print(f'Testing with Scale = {self.scale} ')
            print('Hitbox: ')
            print(self._points)
            print('Original: ')
            self.original()
            print('Patch: ')
            self.patch()
        
    def original(self):
## between these is the original code from arcade.Sprite.get_adjusted_hit_box()
###############################################################################
        point_list = []
        for point_idx in range(len(self._points)):
            # Get the point
            point = [self._points[point_idx][0], self._points[point_idx][1]]

            # Scale the point
            if self.scale != 1:
                point[0] *= self.scale
                point[1] *= self.scale

            # Rotate the point
            if self.angle:
                point = rotate_point(point[0], point[1], 0, 0, self.angle)

            # Offset the point
            point = [point[0] + self.center_x,
                     point[1] + self.center_y]
            point_list.append(point)

        # Cache the results
        self._point_list_cache = point_list

        # if self.texture:
        #     print(self.texture.name, self._point_list_cache)
###############################################################################
    # should be equal to original, since scale and angle suggest no change
        print(self._point_list_cache)
    # instead doubled the center Point

### PATCH
    def patch(self):
        point_list = []
        for x, y in self._points:
            # Transport it to around (0,0)
            x, y = x-self.center_x, y-self.center_y
            
            # Scale the point
            if self.scale != 1:
                x, y = x*self.scale, y*self.scale

            # Rotate the point
            if self.angle:
                x, y = rotate_point(x, y, 0, 0, self.angle)

            # Transport it back
            x, y = x+self.center_x, y+self.center_y
            
            point_list.append([x,y])
        self._point_list_cache = point_list

        print(self._point_list_cache)
        
    

if __name__ == '__main__':
    Test()

```

Helios713:
It only took me hours to find because of the combination of a function with heavy side effects and an incosistency in Properties:
The Property self.hitbox saves and reads from a variable only.
The deprecated self.points saves to that same variable (self._points) but reads from get_adjusted_hit_box()
So no side effect on the surface if you use hit_box until get_adjusted_hit_box() gets called anywhere in the class. Then your hitboxes are all over the place for no apparent reason.

arcade/sprite.py
```diff
@@ -310,24 +310,22 @@ def get_adjusted_hit_box(self) -> List[List[float]]:
 
         # Adjust the hitbox
         point_list = []
-        for point_idx in range(len(self._points)):
-            # Get the point
-            point = [self._points[point_idx][0], self._points[point_idx][1]]
-
+        for x, y in self._points:
+            # Transport it to around (0,0)
+            x, y = x-self.center_x, y-self.center_y
```
Helios713: This is the important line. Without this, the function is mathematically wrong.

pvcraven:
Ok, perhaps a documentation update is needed, but `_points` is always relative to the center. So with the patch, this fails when it should pass:

```
import arcade


class Test():
    def __init__(self):
        # setup
        my_sprite = arcade.Sprite()
        my_sprite.scale = 1.0
        my_sprite.angle = 0
        my_sprite.center_x = 100
        my_sprite.center_y = 100
        my_sprite._points = [[102, 100], [100, 102], [98, 100], [100, 98]]

        hitbox = my_sprite.get_adjusted_hit_box()
        print('Hitbox: ', hitbox)
        assert hitbox == [[202, 200], [200, 202], [198, 200], [200, 198]]

        self.scale = 0.5

        print('Hitbox: ')
        print(my_sprite.get_adjusted_hit_box())

        self.scale = 2.0
        print('Hitbox: ')
        print(my_sprite.get_adjusted_hit_box())


def test_sprite():
    Test()
```

pvcraven:
Unless I goofed (quite possible) I think the end result needs to pass this test:
```
import arcade


class Test():
    def __init__(self):
        # setup
        my_sprite = arcade.Sprite()
        hit_box = [-10, -10], [-10, 10], [10, 10], [10, -10]
        my_sprite.set_hit_box(hit_box)
        my_sprite.scale = 1.0
        my_sprite.angle = 0
        my_sprite.center_x = 100
        my_sprite.center_y = 100

        hitbox = my_sprite.get_adjusted_hit_box()
        print('Hitbox: ', hitbox)
        assert hitbox == [[90, 90], [90, 110], [110, 110], [110, 90]]

        self.scale = 0.5
        hitbox = my_sprite.get_adjusted_hit_box()
        print('Hitbox: ', hitbox)
        assert hitbox == [[95, 95], [95, 105], [105, 105], [105, 95]]

        self.scale = 2.0
        hitbox = my_sprite.get_adjusted_hit_box()
        print('Hitbox: ', hitbox)
        assert hitbox == [[80, 80], [80, 120], [120, 120], [120, 80]]


def test_sprite():
    Test()```

Helios713:
@pvcraven Oh no, I think I misread the documentation then! If it's relative, then your code was obviously correct and I misunderstood how hitboxes work. I thought I could check against get_hit_box() and be fine, but that returns the relative, not the absolute values.
Which makes the Property 'points' even more confusing, since it returns get_adjusted_hit_box(), which is not relative to the center before or after this patch.
All in all, I think I made a mistake.

pvcraven:
Well, original naming wasn't great. And get_adj_hitbox is not taking into account scaling if you specify points? Haven't make sure yet, but it looks off.

Helios713:
No, it does scale. It takes all the specified points, whether deduced from an image or manually provided, then sees if the scale is != 1, scales, looks if the angle is != 0, rotates, then adds the center coords and caches that. That cache (I just looked) is dutifully deleted with every change to any of these values.
set_hit_box needs the relative points, thats all.

PR #609, opened 2020-03-11
Windows path for Soloud library

chpurdy:
I think this may resolve the issue referenced here: https://www.reddit.com/r/pythonarcade/comments/fgkxsb/soloud_not_found/ where Windows cannot find the library files. 

I noticed the forward slash in the copy and pasted path in the initial issue: 
```
SoLoud dynamic link library C:\Python38\lib\site-packages\arcade\soloud/soloud_x86 not found.
```


chpurdy:
This is apparently not the issue, closing.

Issue #372, opened 2019-04-28
Move sound support to SoLoud


pvcraven:
Look at [SoundFile](https://pypi.org/project/SoundFile/) and [SoundDevice](https://pypi.org/project/sounddevice/) and see if they have good ideas we can use.

pvcraven:
The SoundFile/SoundDevice combo appear to only support playback of one sound at a time. That doesn't work well for sound effects.

pvcraven:
Might need to create our own interface on top of: https://app.assembla.com/spaces/portaudio/git/source

pvcraven:
* [avbin](https://avbin.github.io/AVbin/Home/Home.html) - Pygame uses, old (2012) and unsupported
* SoundDevice - only one sound at a time. No good.
* ffmpeg - Currently can't figure out universal support on all platforms
* PyAudio - still worth looking at?

pvcraven:
Soloud works under MacOS and Windows now. Need to get Linux support and it will clear a lot of the rest of these issues up.

humansHS:
[Pyglet Sound and Audio](https://pyglet.readthedocs.io/en/latest/programming_guide/media.html)

>Decoding of compressed audio and video is provided by FFmpeg v4.X, an optional component available for Linux, Windows and Mac OS X.

> **FFmpeg needs to be installed separately.**

Pyglet's FFmpeg works fine with a bunch of features that weren't properly implemented in Arcade (I have been experimenting with it myself and was going to try to fully implement and bug test them before offering to Arcade), you just have to download and install it.

With that said, I am reading [SoLoud QuickStart](https://sol.gfxile.net/soloud/quickstart.html) and it does have quite a lot of support (and I believe it DOES work on Linux?), so seems the switch would be fine. Just thought I would present an alternative in case anything is troubling with SoLoud.

pvcraven:
I've messed with that for years. Getting new programmers to do a separate install of ffmpeg does not lead to success.

I created a separate package with ffmpeg libraries for the different platforms, but it was always a real pain because it would work for a while, then break. And debugging it through all Pyglet's layers was a time-sink.

Right now, when I build SoLoud on linux, then hook it up to Python, I get missing link errors. I'm hopeful I can get that working.

I like soloud because of its ease of use, and support for volume and panning.

humansHS:
[...]

PR #610, opened 2020-03-11
Fix volume control

chpurdy:
Added voice_handle instance variable that is needed for volume/pan control on currently playing sound.  Updated methods to use the voice_handle.  Issue #372


arcade/sound.py
```diff
@@ -52,11 +54,11 @@ def play(self, volume=1.0, pan=0.0):
         if not _audiolib:
             return
 
-        _audiolib.play(self.wav_file,
-                       aVolume=volume,
-                       aPan=pan,
-                       aPaused=0,
-                       aBus=0)
+        self.voice_handle = _audiolib.play(self.wav_file,
```
chpurdy: The volume/pan controls (while the file is playing) requires a voice handle.  This is different from the setVolume/getVolume that uses the object handle (which seems to set the default volume of the NEXT sound to be created).

pvcraven:
Thanks!

PR #631, opened 2020-04-08
Fix missing return in get system mouse cursor

LukasZahradnik:
Currently method `get_system_mouse_cursor` is returning `None` instead of the cursor instance


pvcraven:
Thanks!

PR #632, opened 2020-04-09
Cleaned up docstrings and annotations in 'texture.py'.

robertwayne:
- I went through the docstrings in _`texture.py`_ and fixed inconsistent punctuation and modified phrasing where it improved sentence flow. I also added links to object types _(ala :class:`ObjType`)_ to make the HTML documentation easier to navigate.

- I changed a stray method that was using Google docstring style to what the the rest of the project uses. I don't think this was intended?

- I fixed an error with a _`PIL.Image`_ type annotation _(should have been PIL.Image.Image)_ a string to your local project directory being displayed on the documentation page. It now links to the PIL docs appropriately.

- I added the docstring for the _`load_spritesheet`_ method.

If there is anything you disagree with please let me know and I can modify as needed.

### DIFF
--- a/arcade/texture.py
+++ b/arcade/texture.py
@@ -297,6 +297,8 @@ def load_texture(file_name: str,
     resize with it. Setting can_cache to false will prevent this issue at the expence of additional resources.
 
     :returns: New :class:`Texture` object.
+
+    :raises: ValueError
     """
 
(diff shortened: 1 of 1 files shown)


pvcraven:
Thanks!

PR #634, opened 2020-04-10
Unify side collision handling

LukasZahradnik:
Collision of two objects was handled differently for colliding from left and colliding from right. This could be visible for example in the platformer example, when player walks into a platform from the left side vs the right side.


pvcraven:
Thanks for the PR. This reminded me I don't have a lot of unit tests. So I put some in. And that showed a lot of other problems besides what you found. I worked though these in the shader branch and put them in. As a lot more needed to be fixed, I'm dropping this PR as there will be too many merge conflicts. But fixes are coming.

PR #639, opened 2020-04-15
Implement opacity for TMX object and tile layers

kfields:
Another nice to have easy to implement feature :)


pvcraven:
These things need unit tests.

PR #641, opened 2020-04-16
Allow getting tmx layer by path

kfields:
This allows getting a layer by name as well as traversing layer groups to find a layer.  Ex: 'mylayer' or 'mygroup/mylayer'.  Leading and trailing slashes are stripped out to avoid errors.

### DIFF
--- a/arcade/examples/tmx_object_demo.py
+++ b/arcade/examples/tmx_object_demo.py
@@ -1,0 +1,110 @@
+import arcade
+
+"""
+Starting Template
+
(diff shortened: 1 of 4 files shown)


pvcraven:
Any chance you can put an example tmx file in `arcade/resources/tmx_maps` and then update `tests/unit2/test_tilemap.py` with some tests confirming these new features? (Or create a new test file.)

kfields:
For sure.  That had crossed my mind.  Thanks!

Issue #647, opened 2020-04-18
Sprite top, left, bottom, right is not accurate


markjoshua12:
## Bug Report
I was messing around making my game for the Python Discord Game Jam, and I noticed something odd when I zoom in the player sprite.

![test_3](https://user-images.githubusercontent.com/48671765/79636898-bd17cb80-81ad-11ea-8b5d-cf849e8791f7.gif)

Aside from the jittering (more on that next time), the player sprite is actually 1px lower than intended.
This could be further checked by printing the values:
```
print(f"Top: {self.player.top} | Left: {self.player.right}")
print(f"Bottom: {self.player.top} | Right: {self.player.right}")
```
Output:
```
Top: 79.0 | Left: 79.0
Bottom: 79.0 | Right: 79.0
```
(Note: all the image sizes are 16x16)

Here is another test with jumping

![test_5](https://user-images.githubusercontent.com/48671765/79636811-0f0c2180-81ad-11ea-8e8f-fc50c3d9bad8.gif)

Notice the bottom and the top of the sprite is not aligned with the tiles
![image](https://user-images.githubusercontent.com/48671765/79637391-d884d580-81b1-11ea-8c7f-47eb965c0dc2.png)
![image](https://user-images.githubusercontent.com/48671765/79637411-f6523a80-81b1-11ea-9e18-9d7f9d4295e2.png)

And here is another one using the example code (04_add_gravity.py):
![image](https://user-images.githubusercontent.com/48671765/79637511-bb043b80-81b2-11ea-9f92-a0ec083a2fd9.png)
![image](https://user-images.githubusercontent.com/48671765/79637514-c2c3e000-81b2-11ea-8ffe-830051699b2a.png)
Even here the player sprite is 0.5 pixels below (which is the `TILE_SCALING` amount in the code)

### Actual behavior:
`Sprite` right and bottom is actually (right - scale) and (bottom - scale)

### Expected behavior:
`Sprite` right and bottom should be equal to its `left + image_width` and `top - image_height`

### Steps to reproduce/example code:
Print the `top`, `right`, `bottom`, or `left` of the sprite and compare it with the intended values
Example:
```
# Added in the example code 02_draw_sprites.py
# Inside the for loop to create the ground

print(f"Width: {wall._get_width()} | Height: {wall._get_height()} ")
print(f"Top: {wall.top} | Right: {wall.right} | Bottom: {wall.bottom} | Left: {wall.left}")
```
Output:
```
[...]

PR #654, opened 2020-04-19
Fixed hit_box being 1px less than intended #647

markjoshua12:
I added 1px to the `bottom_border` and `right_border` after all `image.getpixel()` calls to counteract the subtraction that was done before.

Let me know if this works. Thanks

#647


pvcraven:
Thanks for the PR. This change (or one really close) is currently in shader branch and should be out in the alpha.

https://github.com/pvcraven/arcade/blob/shader_experimental/arcade/drawing_support.py#L209

I think you tried running the alpha, but I wonder if you possibly still ran the old version?

markjoshua12:
I am pretty sure I have the experimental version I saw saw my tile grid had gaps between them when rendering. I couldn't get the `venv` to work so I uninstalled the current version and installed the experimental one.

PR #659, opened 2020-04-22
Disabled caching in draw_text

markjoshua12:
I disabled caching (by default) in `draw_text`, as it seems to take a lot of memory. I also reduced the max `global draw_text_cache` from `5000` to `1000`.

Also added an argument for `caching` if the user still wants to use it.

Thanks


pvcraven:
This would really slow text drawing. I don't see doing this by default.

Text drawing should be improved as part of a re-do once arcade moves to Pyglet 2.0.

markjoshua12:
I understand. I hope the text drawing gets improved soon. I had to resort to using bitmap images for the text drawing 😅

einarf:
@markjoshua12 Share your bitmapped test rendered?

PR #674, opened 2020-04-24
fix_texture_read

gcordonn:
I am experimenting on the edge artifact issue, and I bumped into a small bug with Texture.read, which resulted in a silent crash because of an allocation error.


einarf:
Thanks. I don't think I ever tested this one 😄 (Unit tests incoming soon)
There are more things to sort out regarding alignment and different texture formats, but that's also on the list.

Issue #660, opened 2020-04-23
Possible fix for edge artifact in SpriteList batch draw


gcordonn:
First, thanks for this nice package! I just discovered it - I used to work with heavier 3D engines, and I needed to do some 2D experiments, that's super useful. 

I quite quickly bumped into the edge artifact problem, and after looking around it seems like it already raised quite a lot of discussion, so please apologize if I am proposing something that was already rejected. 

It is a pity that was cannot use arbitrary transform or viewport resizing, especially with a package that is based on OpenGL, which is supposed to handle all of these smoothly. 

I quickly looked at the code, and it seems like the sprite list is using a texture atlas to store all textures used during batch rendering. The induced gain in batch-processing comes with issues with this technics, as discussed in http://download.nvidia.com/developer/NVTextureSuite/Atlas_Tools/Texture_Atlas_Whitepaper.pdf.

In particular, as already explained in Arcade documentation, the issue comes from the bilinear filtering: when a texel is sampled at a noninteger position, it gets interpolated with the neighboring one - which is somewhere in the texture atlas but sometimes out of the texture of the rendered sprite. 

One (easy) fix is to add a .5 offset to the texture coordinate (before normalization), in the _calculate_sub_tex_coords function of SpriteList.py, so that the texel samples are located at the middle of the pixel. 

```python
            offset=.5

            for index, image in enumerate(self.array_of_images):
                column = index % grid_width
                row = index // grid_width

                # Texture coordinates are reversed in y axis
                row = grid_height - row - 1

                x = column * (grid_item_width + MARGIN) +offset
                y = row * (grid_item_height + MARGIN) +offset

                # Because, coordinates are reversed
                y += (grid_item_height - (image.height - MARGIN))

                normalized_x = x / sprite_sheet_width
                normalized_y = y / sprite_sheet_height

                start_x = normalized_x
                start_y = normalized_y

                normalized_width = (image.width-2*offset) / sprite_sheet_width
                normalized_height = (image.height-2*offset) / sprite_sheet_height
```

But this solution will give results that are not consistent with the current behavior of arcade (although I think this is the more 'correct' for perfect pixel alignment.)

Then there is a second solution, that would also be preferred should mipmapping be supported, which is to pad the textures in the sprite map with a one-pixel border:

```python
                if name_of_texture_to_check not in new_array_of_texture_names:
                    new_array_of_texture_names.append(name_of_texture_to_check)
                    if sprite.texture is None:
[...]

PR #675, opened 2020-04-25
Fix border artifact

gcordonn:
Both versions that I proposed in #660 seems to work and fix the edge artifact issue.

I compared both versions and it seems that the half-pixel solution slightly reduces the quality of the sprites. However, duplicating the border of the sprite texture and using a full pixel offset gives a result close to the previous ones.   

The edge artifact will still happen int tilings if there is some transparent overlap in the texture (like for emulating orthographic perspective). This can only be solved by the author of the texture, by copying the RGB color of the visible edges to the neighboring transparent pixels (which alpha value is kept as 0).


einarf:
Thanks a lot for this. Looks like this is the solution!

There are of course other concerns such as if padding should be re-added and optimizing building the atlas, but that is outside of the scope here.

Issue #661, opened 2020-04-24
SpriteLists created by loading .TMX should have spatial hashing turned on by default.


pvcraven:

pvcraven:
Doesn't work to do this by default as pymunk needs it off. Instead, we are allowing it to be turned on during load, and if the user doesn't specify, we do our best guess.

PR #678, opened 2020-04-28
Added a way to turn on/change the spatial hashing of a SpriteList 

markjoshua12:
Will turn on `use_spatial_hashing` to `True`. This was made to address Issue #661 

The function name can be discussed and is open for changes.

Thanks


markjoshua12:
Removing excluded commits

PR #679, opened 2020-04-28
Added a way to turn on/change the spatial hashing of a SpriteList

markjoshua12:
Will turn on use_spatial_hashing to True. This was made to address Issue #661

The function name can be discussed and is open for changes.

Thanks


pvcraven:
Closing this PR, as something similar was done in a different check-in. Always enabling it doesn't work if we use pymunk, which would cause that engine to slow down. Thanks for pushing to make that part better.

PR #683, opened 2020-05-02
Added the ability to set margins on camera

markjoshua12:
Cleaned up some code and added margin variables for each side (`left` `right` `bottom` `top`), and a function to set them.


einarf:
Might as well just merge this into the current camera. It can be useful.

PR #705, opened 2020-06-17
Fix wrong exception type on bad type passed to texture write

### LINKED ISSUE #704: Wrong type arg to write should raise TypeError, not ValueError
### Actual behavior:
The `write` method of the GL texture abstraction raises a ValueError on unexpected types when a TypeError is more appropriate.

From the [ValueError documentation](https://docs.python.org/3/library/exceptions.html#ValueError):
> Raised when an operation or function receives an argument that has the right type but an inappropriate value

From the [TypeError documentation](https://docs.python.org/3/library/exceptions.html#TypeError):

> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

### Expected behavior:
Passing an unexpected type should produce a TypeError

Pass an invalid type to write.
```python
import arcade
from arcade.gl import geometry

# Do the math to figure out our screen dimensions
SCREEN_WIDTH = 64 * 10
SCREEN_HEIGHT = 32 * 10


class MyGame(arcade.Window):

        """CHIP-8 Screen"""

        self.quad = geometry.screen_rectangle(0,0, SCREEN_WIDTH, SCREEN_HEIGHT)
        self.texture = self.ctx.texture((8, 32), components=1, dtype='i1')
        self.texture.write("This should be a bytes object")

    def on_draw(self):
        self.clear()
        self.texture.use(0)
        self.quad.render(self.program)

if __name__ == "__main__":
    MyGame(SCREEN_WIDTH, SCREEN_HEIGHT, "example")
    arcade.run()
```

pushfoo:
Closes #704

### DIFF
--- a/arcade/gl/texture.py
+++ b/arcade/gl/texture.py
@@ -272,7 +272,7 @@ class Texture:
             gl.glTexSubImage2D(self._target, level, x, y, w, h, self._format, self._type, 0)
             gl.glBindBuffer(gl.GL_PIXEL_UNPACK_BUFFER, 0)
         else:
-            raise ValueError("data must be bytes or a Buffer")
+            raise TypeError(f"data must be bytes or a Buffer, not {type(data)}")
 
     def build_mipmaps(self, base=0, max_amount=1000) -> None:
         """Generate mipmaps for this texture.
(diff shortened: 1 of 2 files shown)


pushfoo:
Relevant tests seem to pass

PR #714, opened 2020-07-05
#701 Fix on_show called twice

### LINKED ISSUE #701: on_show called twice on start up
## Bug Report
`View.on_show` called twice on start up, because the view is pushed as a handler to the window and receives the `on_show` event which has the same name like the `Window.on_show` callback.

### Actual behavior:
`on_show` is called twice

1.  window.show_view(view)
2. arcade.run() -> view receives on_show event of the window

### Expected behavior:
on_show called once

```
import arcade

class MyView(arcade.View):
    def on_show(self):
        print('on_show')

if __name__ == '__main__':
    window = arcade.Window(width=300, height=300)
    view = MyView()
    window.show_view(view)
    arcade.run()
```

### Suggestion 
Rename `View.on_show` to prevent that it is picked up as callback.

eruvanos:
Fixes #701 

An alternative to this dynamic setup could be to introduce a `on_show_view()` which is guaranteed to be called the right amount of times and deprecate the other one.

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -450,7 +450,13 @@ class Window(pyglet.window.Window):
 
         # push new view's handlers
         self._current_view = new_view
-        self.push_handlers(self._current_view)
+        self.push_handlers(
+            **{
+                et: getattr(new_view, et, None)
+                for et in self.event_types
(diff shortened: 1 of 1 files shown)


eruvanos:
Would prefer #716

PR #715, opened 2020-07-05
Support parsing img references to load arcade.Texture

eruvanos:
This is will allow UIElements to load Arcade Textures directly from yaml


pvcraven:
Rather than add a `img:` in front, can we just use `:resources:`? Having `img::resources:` seems unnecessarily redundant.

eruvanos:
Not sure:

`img:<file path>` indicates that the value should be rendered as texture.
`:resources:` is used within arcade to reference to arcade resource files.

So for the user it would look like:
```
image_button_1:
    normal_texture: img:path_to_my_img.png
```

I agree that the combination of both look weird, but using a different separator was not better.

An alternative would be to just check if the string is a valid file path, but in case of an error this would be harder to recognize for the user.

eruvanos:
Will revisit this PR

PR #717, opened 2020-07-05
New callbacks for setup and teardown of a view

eruvanos:
From my point of view this looks better then #714


pvcraven:
This renames `Window.on_show` to `Window.on_show_view`? That will add a back-wards breaking change? While I like the name change, I don't like the idea of breaking everyone's code. We'd need to approach that with a lot of caution.

eruvanos:
@pvcraven Hi, the callback is in addition.

It was a new one, to prevent any breaking change.

Issue #727, opened 2020-07-25
Add a method to determine if a sound is done playing


xlatb:
### What should be added/changed?

It would be useful if one could determine if an `arcade.Sound` object has completed playing.

### What would it help with?

Possible use cases:
1. Looping a sound (if it has completed, play it again)
2. Maintaining background music (if the music is done, play the next one in the list)
3. Synchronizing game events with sound (show a message or graphic while a fanfare plays, and automatically continue when the sound ends)

This information is somewhat available via the current API's `get_stream_position()` method, but this is lacking because a return value of 0 could mean either the sound has just started playing, or it has ended. Granularity of the stream position is [documented](https://sol.gfxile.net/soloud/coremisc.html) as being up to 45ms, so even if a sound is playing, the return value could be 0 across multiple frame durations. And if a sound is very short, it may never return a stream position other than 0.

Another possibility is to use `get_length()` and then wait for the amount of time that we expect the sound to take. But this cannot detect sounds that have been stopped early for one reason or another. In particular, the Soloud library itself can stop sounds if you attempt to play many of them at the same time, as documented [here](https://sol.gfxile.net/soloud/concepts.html):
> If all channels are already playing and the application requests another sound to play, SoLoud finds the oldest voice and kills it.

pvcraven:
Should be in 2.4.2 release.

PR #728, opened 2020-07-25
Allow checking if a sound has completed playing.

xlatb:
Adds `is_complete()` method to `arcade.Sound`. For issue #727.


pvcraven:
Thanks!

PR #746, opened 2020-09-04
Changing to Pyglet from Soloud for Sound

Cleptomania:
This change re-works the sound module a bit, primarily abstracting the back-end behind a common interface, and switching from the SoLoud library to Pyglet. It has been tested against Windows, Mac, and Linux with .wav, .mp3, and .ogg formats.

There are no API breaking changes and everything should be able to use it in the same way as previously. However there are some feature additions.

Added Features:

- Sound looping
- Checking if a sound is playing or not
- Retrieve the length of a sound(appeared to be broken with the SoLoud implementation)
- Get/Set the volume of a sound after starting playback(appeared to broken with the SoLoud implementation)

This should also greatly improve cross-platform support.

The ffmpeg binaries for Mac have been added in a similar fashion to how SoLoud's binaries were included. They can be seen in the "arcade/lib" folder. If ffmpeg is desired for Windows/Linux all one would need to do is drop the corresponding binaries into there and they would automatically be loaded. The Windows/Linux binaries have not been included due Pyglet having out of the box support for Windows/Linux


einarf:
We should probably try to get this one merged soon. There are some PyWeekers bumping into audio problems.

Cleptomania:
Worth noting if https://github.com/pythonarcade/arcade/pull/780 gets merged then the `MANIFEST.in` file will need updated to remove packaging the soloud binaries and add packaging ffmpeg binaries for mac.

Issue #748, opened 2020-09-06
Setting a shader uniform array only sets the first element of the array


64-bit:
### Actual behavior:
Setting a shader uniform array only sets the first element of the array, leaving all remaining elements at their default values. Running the example code below produces the following image. The example code uses the uv x coordinate to lookup a value from a uniform array and output that as a color.

![image](https://user-images.githubusercontent.com/2972292/92334659-bd7a0b80-f055-11ea-85ed-0fea300aa493.png)

### Expected behavior:
Setting a shader uniform array sets all array elements. Running the example code below should produce the following image.

![image](https://user-images.githubusercontent.com/2972292/92334671-dd113400-f055-11ea-8e3b-5789ac5625a8.png)

### Steps to reproduce/example code:
1. Compile the following shader code into a program
2. Set the uniform array to a Python array of float values
3. Draw to the screen using this shader program.

Reproduction Python Code (taken from a class derived from arcade.Window)
```
    def setup(self):
        """ Set up everything with the game """

        #Create VBO & VAO for a fullscreen quad
        self.fullscreen_quad = FullscreenQuad(self.ctx)

        #Load reproduction shader
        self.program = self.ctx.load_program(
            vertex_shader='fullscreen_quad.vs',
            fragment_shader='reproduce_uniform_arrays.fs'
        )

        #Assign array to uniform array in program. Only the first value (0.2) is set 
        self.program['u_uniform_array'] = [0.2, 0.4, 0.6, 0.8, 1.0]

    def on_draw(self):
        """ Draw everything """
        self.fullscreen_quad.render(self.program)
```
fullscreen_quad.vs
```
#version 330

in vec2 in_pos;
in vec2 in_uv;

out vec2 v_uv;

void main() {
    v_uv = in_uv;
    gl_Position = vec4(in_pos, 0.0, 1);
}
```

reproduce_uniform_arrays.fs
```
#version 330

in vec2 v_uv;

out vec4 fragColor;

//Declare a uniform array
uniform float u_uniform_array [5];

void main() 
{
    //Convert UV.x into a 0->4 array index
    int index = int(v_uv.x * 5.0);
    //Read from uniform array
    float uniform_value = u_uniform_array[index];
    //Return as Color
    fragColor = vec4(vec3(uniform_value),1.0);
}
```

### Cause of issue

[...]

PR #749, opened 2020-09-06
Issue #748 - Fixed bug when setting shader uniform arrays

64-bit:
Here is my proposed fix for the issue described in #748. 

The size value obtained from glGetActiveUniform is used as the count argument for glUniformXXv, allowing for the entire uniform array to be set, and causing no change in behavior for uniforms that are not arrays, as glGetActiveUniform will return size 1 for them.


einarf:
Thanks a lot!

Issue #769, opened 2020-10-03
No way to draw sprite lists with alternative blending modes.


64-bit:
## Enhancement request:
Add support for selecting a blending mode when drawing sprite lists. Currently calling SpriteList.draw will set a default alpha-blending mode and override any changes you have made to the OpenGL blend state.

### What should be added/changed?
I am proposing that a optional varible is added to SpriteList.draw(), named 'blend_function' that the user can use to specify a custom blending mode. If not specified the original behaviour should remain. This may not be the appropiate way to solve this problem and I am open to suggestions. I will create a pull request with my proposed change.

### What would it help with?
Drawing sprites with, for example, a additive blending mode for effects or emissive layers.

64-bit:
Pull request here: #770

einarf:
Merged.

PR #770, opened 2020-10-03
Added support to specify SpriteList.draw blending mode

64-bit:
This feature is in reference to issue #769 . Support has been added to specify a optional argument blend_function when calling SpriteListl.draw, that will set the OpenGL blending function to the one specififed in the argument.


einarf:
Makes sense!

Issue #771, opened 2020-10-14
arcade.geometry.are_polygons_intersecting calculates intersections for convex hull


LiorAvrahami:
## Bug Report
arcade.geometry.are_polygons_intersecting calculates intersections for the convex hull of the given polygons. this is a problem when "hit_box_algorithm" is selected to be "Detailed", which in general doesn't generate convex hitboxes.
as far as I could find, the fact that arcade.geometry.are_polygons_intersecting calculates intersections of the convex hulls **is not documented** so at the very least this documentation issue should be fixed (altho this issue massively hinders the effectiveness of the "Detailed" "hit_box_algorithm", and thus I think we should look for a different polygon intersector).
I have an idea for a solution, see - "What is wrong with it? How can it be improved?", I would be happy to implement it myself, but am new to github so don't really know how all these forks and merges work (I have always only worked alone on git)

### Actual behavior:
given two polygons, "arcade.geometry.are_polygons_intersecting" calculates intersections of their convex hulls, so if one of them is not convex, the collision calculation will be false.

### Expected behavior:
are_polygons_intersecting should calculate the intersections of the polygons without taking their convex hulls.

### Steps to reproduce/example code:
at the bottom of this page

### What would it help with?
currently if a  "Detailed" hit_box_algorithm is selected when creating a texture, than the hitbox is in general very well made. it's too bad that the polygon collision doesn't allow for general detailed hitboxes, and works correctly for allows convex ones. 

### What documentation needs to change?
if this issue is not addressed than the documentation should state clearly at both the documentation of the **hit_box_algorithm parameter in the Texture initialization function** and at the documentation of **arcade.geometry.are_polygons_intersecting** that **collisions are calculated for convex hulls only**

### What is wrong with it? How can it be improved?
[...]

PR #776, opened 2020-10-18
changed algorithm for checking for polygon collisions. to fix issue #771

LiorAvrahami:
the old algorithm assumed that the polygons are convex, which is not the case for Detailed hitboxes.

### DIFF
--- a/arcade/geometry.py
+++ b/arcade/geometry.py
@@ -3,11 +3,33 @@ Functions for calculating geometry.
 """
 
 from typing import cast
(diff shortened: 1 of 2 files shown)


pvcraven:
Thanks for the PR. Might be a little bit before I get to it, but I will do so.

pvcraven:
Thanks for the contribution. We are now using shapely instead, which is faster than what we had before and a bit more 'complete'.

PR #780, opened 2020-11-03
Update setuptools configuration to align with PEP 517/518

Cleptomania:
[PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/) outline some ideals for the format of build system configuration.

This PR serves to align Arcade's build system more closely with these PEPs. From an end-user perspective there shouldn't be any noticeable difference. However this should serve to cleanup the development install process and simplify the build system. As well as help serve as an example of good standards for project setup.

The key change here is primarily using the `setup.cfg` file rather than the `setup.py` file, and switching to the [pep517](https://pypi.org/project/pep517/) package for building the wheel for distribution. 

To install the arcade from source after these changes, the below command can be used:
```bash
pip install -e "."
```

Extra package groups can be defined as desired by using:
```bash
pip install -e ".[dev, build, docs]"
```

With this change in order to build the wheel all that's needed(assuming the build package group was defined during install) is:
```bash
mkdir dist
python -m pep517.build .
```
This will build the wheel and tar.gz file into the dist directory.

Another change this makes is pinning crucial dependency versions, specifically, pyglet, pillow, pymunk, numpy, pytiled_parser. They are pinned at their current latest versions which is what was previously being installed and tested with. Not having the versions of these packages does not feel safe as at any time breaking changes could be released on them and Arcade would always be installing the latest version from PyPi


setup.cfg
```diff
@@ -31,10 +31,10 @@ packages = find:
 include_package_data = True
 python_requires = >=3.6
 install_requires =
-    pyglet
-    pillow
-    numpy
-    pymunk
+    pyglet>=1.5.8,<2
+    pillow == 8.0.1
```
SirGnip: Is it worth doing ranges on these other dependencies (like was done for pyglet) instead of pinning them to specific versions?  

As written, if an app depends on arcade AND other modules... there is a chance that the other module also has pillow, numpy, or pymunk in its dependency tree.  This means version conflicts could arise.  For instance, if an app imports arcade and module "X", and module "X" has numpy 1.18.1 as a dependency... there is a conflict because both numpy 1.18.1 (from module "X") and numpy 1.19.4 (from arcade) can not both be installed.  So, my understanding is that libraries (such as arcade) should use ranged dependencies where possible.

Now, this would probably require testing to find out what version ranges of pillow, numpy and pymunk would work with arcade.  Semantic versioning could help here, as API breaking changes are usually indicated by bumping the major version number (I don't know if numpy, pillow, pymunk follow this convention...)

One (of many) references: https://www.promptworks.com/blog/pin-all-dependencies/ (see the "Instance #1: Your software is a third-party module" under "When should you not pin a dependency?").

Cleptomania: That's a good catch. I'll do some research into this tonight, pinning specific versions is probably the right way to go if that packages dependencies are pinned. I will go over everything and find out specific versions and if they should be ranges or not and submit another commit to handle that if necessary.

SirGnip:
This might be jamming too much in this one commit, but I just noticed type hinting is broken for code that imports arcade and created this ticket https://github.com/pythonarcade/arcade/issues/781.  I think the fix is a one-liner in setup.py (or however pyproject.toml/setup.cfg handles stuff like this)

Cleptomania:
I think we'd be fine to fix https://github.com/pythonarcade/arcade/issues/781 in this, as it is solely related to the build system not including the proper file. It's a simple one line fix that needs added to `MANIFEST.in` to include the `py.typed` file.

I'll add it in tonight when I work on the version pinning and test it out.

setup.cfg
Cleptomania: This issue is resolved in https://github.com/pythonarcade/arcade/pull/780/commits/7d7c87d687af13669f90f320f4243cb022b417fa There are no cross dependencies for anything here, so I've pinned Pyglet to ensure it doesn't install version 2 if that releases and we're unprepared for it.

pytiled_parser is pinned at the exact version in order to ensure compatibility if we release version 1.0.0 of that before the Arcade release is ready for that.

Everything else is pinned to approximate minor versions, so that it will install the latest patch version of the most recent compatible minor version.

pvcraven:
Is this ready? Or ready-ish?

Cleptomania:
I think it's ready but I figured it would be best for you to test out building with it since I'm not 100% sure what your process was for building a release before, just to make sure everything you'd expect to still works. There should be no impact to end users, but make's the install from source and build process a little more clean.

I've looked into all branching dependencies of our dependencies and ensured there are no version conflicts. As of the versions I've pinned they're all working and good to go. Pyglet is pinned to a specific range that I've determined with @einarf and everything else is pinned to an approximate matching the minor version, so it will install the latest patch version of each but not go over the minor version.

Issue #782, opened 2020-11-06
#782 Faster has_line_of_sight function with shapely


akapkotel:
## Enhancement request:
Replace current arcade has_line_of_sight implementation with shapely library for faster visibility-checking.

#### 2. shapely imports in arcade.paths.py
![shapely-imports](https://user-images.githubusercontent.com/40523203/98357979-4c8bab80-2026-11eb-8f97-5295b64fcf55.png)

#### 3. paths.has_line_of_sight function would look like this:

![shapely-function](https://user-images.githubusercontent.com/40523203/98358348-d9cf0000-2026-11eb-8c77-6c98dad9ee82.png)
 ##
### What would it help with?
Checking visibility between Sprites would be much faster. Shapely is extremely efficient in finding intersections between geometries.

### What documentation needs to change? 
Docts for has_line_of_sight method of arcade.paths

### What is wrong with it? How can it be improved?
Current has_line_of_sight function works, but it is not so efficient. With just 25 sprites and 25 obstacles if each sprite try to find line of sight to each other, it takes roughly 7 seconds for frame to update! With Shapely it is just 0.18 second: 35 times faster! With 1 player-sprite checking line of sight against 1156 enemies and 25 obstacles, it currently takes about 16 seconds per frame, whereas shapely does the job within 0.09 second

PR #783, opened 2020-11-06
Issue #782 Using shapely to find line of sight

akapkotel:
What is wrong with it? How can it be improved?
#782 
Current has_line_of_sight function works, but it is not so efficient. With just 25 sprites and 25 obstacles if each sprite try to find line of sight to each other, it takes roughly 7 seconds for frame to update! With Shapely it is just 0.18 second: 35 times faster! With 1 player-sprite checking line of sight against 1156 enemies, it currently takes about 16 seconds per frame, whereas shapely does the job within 0.09 second

### DIFF
--- a/arcade/paths.py
+++ b/arcade/paths.py
@@ -2,50 +2,47 @@
 Path-related functions.
 
 """
+from shapely import speedups
+from shapely.geometry import LineString, Polygon
+
 from arcade import Point
-from arcade import get_distance
-from arcade import lerp_vec
-from arcade import get_sprites_at_point
 from arcade import check_for_collision_with_list
 from arcade import Sprite
 from arcade import SpriteList
 
+
+speedups.enable()
+
+
 def has_line_of_sight(point_1: Point,
                       point_2: Point,
                       walls: SpriteList,
-                      max_distance: int = -1,
-                      check_resolution: int = 2):
+                      max_distance: int = -1):
     """
-    Determine if we have line of sight between two points. Try to make sure
-    that spatial hashing is enabled on the wall SpriteList or this will be
-    very slow.
+    Determine if we have line of sight between two points. Having a line of
+    sight means, that you can connect both points with straight line without
+    intersecting any obstacle.
+    Thanks to the shapely efficiency and speedups boost, this method is very
+    fast. It can easily test 10 000 lines_of_sight.
 
-    :param Point point_1: Start position
-    :param Point point_2: End position position
-    :param SpriteList walls: List of all blocking sprites
-    :param int max_distance: Max distance point 1 can see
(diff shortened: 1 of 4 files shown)


pvcraven:
Thanks for the PR! That sounds very promising.

I'm not processing these PRs as fast as I'd like, so thanks for your patience on this.

akapkotel:
@pvcraven sure, no problem!  I've just realised, that there is another, potential gain of implementing shapely: it does not require SpriteLists to use spatial-hashing for efficiency, so moving Sprites could be tested for visibility with the same efficiency as static ones!

I use shapely in my RTS-python-game project with a good results.

pvcraven:
Oh nice. Teaching is occupying a lot of my time right now, but once break hits, I want to dive into this.

pvcraven:
We do need to be more careful about the version number of our libraries, but I'm not sure this is the PR to do that in. Thinking maybe that should be separated.

einarf:
Yes, I would revert the new version constraints for now. That needs to be carefully looked at.

Cleptomania:
My PR https://github.com/pythonarcade/arcade/pull/780 deals with the package versioning issues, and is probably the more appropriate place for that to be handled since it re-does the whole setuptools configuration.

It does only deal with the versions for core packages though, like pyglet, pillow, pytiled_parser, etc. It does not do it for the dev dependencies like pytest.

pvcraven:
With a simple test I was getting 4.77 ms with the shapely code, and 8.55 ms without shapely, but with spatial hashing, and 101 ms without spatial.

This makes me also want to explore using shapely for collision detection. Does that work?

Do we want shapely as a dependency?

pvcraven:
Real quick error-prone test:

* Intersection with no spatial hash: 13 ms
* Intersection with spatial hash 8 ms
* Shapely intersection code: 6.6 ms

einarf:
Worst case there could be a shapely implementation and a pure python one? I'm not sure if that is too much to maintain.

pvcraven:
I wonder if this also fixes issue #771 as an alternative to PR #776.

pvcraven:
If you can update the PR to not change so much in the requirements, that would be great. I'll merge it in.

akapkotel:
@pvcraven yup, shapely also has collisions for polygons, both convex and concave etc. So it could help with collisions-handling too. I can explore that, test efficiency and if it works faster than current collision-detection, start work on bigger commit. Should I? It seems that adding shapely-depenedence would be a good idea.

pvcraven:
Let's do that as a separate PR. If this one has requirements.txt cleaned up, I'll put it in for 2.5, then do a different PR for working collision detection.

akapkotel:
@pvcraven I removed requirements.

PR #788, opened 2020-11-10
Fix #771 by replacing collision-tests with shapely

### LINKED ISSUE #771: arcade.geometry.are_polygons_intersecting calculates intersections for convex hull

akapkotel:
1. Replaced all geometry collision test with shapely functions.
2. Replaced math.sqrt function with math.hypot in get_distance function because hypot is faster - see math module documentation.

Additionally:
3. Removed some redundant local assignments in sprite_list.py functions with straightforward returns.

## Enhancement request:
Replacement current collision-test functions  with shapely library geometry functions.

### What would it help with?
Faster collision tests than solution proposed in #776  but still solves issue in #771 since shapely library works fine with concave polygons.
I've already tested both solutions - mine and @LiorAvrahami - with a concave polygons and time-performance of shapely is up tu 3 times better (but, unfortuunatelly still slower than original polygons-collisions test, still the original algorithm did not work for concave poly's, so it's speed was artificial).

### Where is it located?
arcade.geometry.py, arcade.sprite_list,py

### DIFF
--- a/arcade/geometry.py
+++ b/arcade/geometry.py
@@ -1,12 +1,16 @@
 """
 Functions for calculating geometry.
 """
+from shapely import speedups
+from shapely.geometry import LineString, Polygon, Point
 
 from typing import cast
 from arcade import PointList
 import math
 
 _PRECISION = 2
+
+speedups.enable()
 
 
 def are_polygons_intersecting(poly_a: PointList,
@@ -21,38 +25,10 @@ def are_polygons_intersecting(poly_a: PointList,
     :rtype bool:
     """
 
-    for polygon in (poly_a, poly_b):
+    shapely_polygon_a = Polygon(poly_a)
+    shapely_polygon_b = Polygon(poly_b)
 
-        for i1 in range(len(polygon)):
-            i2 = (i1 + 1) % len(polygon)
-            projection_1 = polygon[i1]
-            projection_2 = polygon[i2]
-
-            normal = (projection_2[1] - projection_1[1],
-                      projection_1[0] - projection_2[0])
-
-            min_a, max_a, min_b, max_b = (None,) * 4
-
-            for poly in poly_a:
-                projected = normal[0] * poly[0] + normal[1] * poly[1]
-
-                if min_a is None or projected < min_a:
-                    min_a = projected
-                if max_a is None or projected > max_a:
-                    max_a = projected
-
-            for poly in poly_b:
-                projected = normal[0] * poly[0] + normal[1] * poly[1]
-
-                if min_b is None or projected < min_b:
-                    min_b = projected
-                if max_b is None or projected > max_b:
-                    max_b = projected
-
-            if cast(float, max_a) <= cast(float, min_b) or cast(float, max_b) <= cast(float, min_a):
-                return False
-
-    return True
+    return shapely_polygon_a.intersects(shapely_polygon_b)
 
 
 def is_point_in_polygon(x, y, polygon_point_list):
@@ -67,27 +43,12 @@ def is_point_in_polygon(x, y, polygon_point_list):
     Returns: bool
 
     """
-    n = len(polygon_point_list)
-    inside = False
-    if n == 0:
-        return False
+    shapely_point = Point(x, y)
+    shapely_polygon = Polygon(polygon_point_list)
 
-    p1x, p1y = polygon_point_list[0]
-    for i in range(n+1):
-        p2x, p2y = polygon_point_list[i % n]
-        if y > min(p1y, p2y):
-            if y <= max(p1y, p2y):
-                if x <= max(p1x, p2x):
-                    if p1y != p2y:
-                        xints = (y - p1y) * (p2x - p1x) / (p2y - p1y) + p1x
-                    # noinspection PyUnboundLocalVariable
-                    if p1x == p2x or x <= xints:
-                        inside = not inside
-        p1x, p1y = p2x, p2y
-
-    return inside
+    return shapely_polygon.contains(shapely_point)
 
 
 def get_distance(x1: float, y1: float, x2: float, y2: float):
     """ Get the distance between two points. """
-    return math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
+    return math.hypot(x1 - x2, y1 - y2)
--- a/arcade/sprite_list.py
+++ b/arcade/sprite_list.py
@@ -1168,7 +1168,7 @@ def check_for_collision_with_list(sprite: Sprite,
         sprite_list_to_check = sprite_list
 
     # print(len(sprite_list_to_check.sprite_list))
-    collision_list = [sprite2
+    return [sprite2
                       for sprite2 in sprite_list_to_check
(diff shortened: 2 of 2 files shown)


SirGnip:
I don't have a lot of context on this issue and I know nothing about shapely, but wanted to as a question.  It is obviously good to have a solution that handles concave poly collisions. But, this means that convex polygon collisions are now slower?  Would it be worth having two different collision algorithms available?  One that handles convex and concave polys (using shapely library) and another that handles only convex polys (will be faster for this specific case). That way, if someone knows they only need convex collision testing, they can use the original algorithm and not pay the speed penalty for their simpler use case.  The shapely algorithm could be the "default".  Then, for those users who need the optimizations, they could do something different (call a different function, pass an argument, etc.) to use the original algorithm?  Just brainstorming...

akapkotel:
@SirGnip your idea sounds great. Not-loosing old-concave algorithm speed seems being worth of additional work. We only need a cheap method of finding if a polygon is convex to fast-detect which case are we dealing with. It could be a counter-clockwise points order check maybe?

pvcraven:
I need to redo benchmarks, but my initial tests (not and all thorough) was shapely was faster than anything in Arcade currently.

akapkotel:
@pvcraven how do you benchmark? It seems that your results are always different than mine, so I am curious about how do you profile code execution.
I use my own custom function to decorate profiled functions and enclose them with time.perf_counter calls.

pvcraven:
I have some custom code here I like to use:

https://github.com/pythonarcade/performance_tests

pvcraven:
I carelessly merged this in, now the unit tests are getting some errors around collision and line-of-sight. Hoping I'll be getting some time to hack this out and clean up unit tests.

Issue #791, opened 2020-11-26
transform mouse event location to be in window viewport coordinates, for fixing mouse events breaking when getting the game to scale along with the window


LiorAvrahami:
not sure if bug or enhancement request.

## Enhancement request:
transform mouse event location to be in window viewport coordinates, for fixing mouse events breaking when getting the game to scale along with the window

### What should be added/changed?
the location of mouse events is currently in window coordinates. which is usually the same as the window viewport coordinates. I am suggesting that when these are different, the window viewport coordinates be used. this should be implemented by a simple scaling and translation transformation on the mouse coordinates.

### What would it help with?
as shown in the example named "Full Screen Example" (link below), getting the game to scale along with the window is implemented via setting window.set_viewport( 0, SomeConstentWidth, 0, SomeConstentHeight). this method works well! the only problem I could find with it is that the mouse click events' locations are in window coordinates and not in window viewport coordinates, which breaks custom made buttons and any arcade.gui.UIClickable.
link to "Full Screen Example": [https://arcade.academy/examples/full_screen_example.html](https://arcade.academy/examples/full_screen_example.html)

LiorAvrahami:
made a pr for this #792

einarf:
We're also worked on a camera : https://github.com/pythonarcade/arcade/blob/development/arcade/experimental/camera.py

PR #792, opened 2020-11-26
allow stretchy resizable window support

LiorAvrahami:
### this pr solves issue #791 (which i had submitted) which regards the impossibility of making stretchy resizable games that use the built in arcade gui.
in a game I'm writing I needed this to be supported, and in general I think this is an important option to have for games where the relation between the spirits' sizes to the screen's size is a critical part of the game (for example pong, space invaders, etc.). currently games where you want a change in the window's size to lead to a proportionate scaling of the game, so that the game fits exactly in the window, call the function window.set_viewport with some constant values on every window resize. the current problem with this method is that the pyglet module, witch passes the mouse event object to the gui manager, pass the xy coordinates in window coordinates, instead of passing them in the game coordinates. this is the gist of issue #791. I couldn't find a fix for this besides a basic translation-scaling transformation on the mouse events coordinates, this is what I did in the first commit of this pr. the second commit is meant to make the option of making you're game stretch along with the window more accessible. and the last commit is some example to help illustrate what I'm talking about.

_p.s. when the aspect ratio of the window changes all the game art gets proportionally smushed, in my game only **(and not in this pr)** I solved this by not allowing the screen size to change aspect ratio:_
```
# in class MainGaime(Window)
def on_resize(self, width: float = 0, height: float = 0):
     self.height = int(width*self.aspect_ratio)
```

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -58,7 +58,8 @@ class Window(pyglet.window.Window):
                  resizable: bool = False,
                  update_rate: Optional[float] = 1 / 60,
                  antialiasing: bool = True,
-                 screen: pyglet.canvas.Screen = None):
+                 screen: pyglet.canvas.Screen = None,
+                 stretch_game_with_window:bool=False):
         """
         Construct a new window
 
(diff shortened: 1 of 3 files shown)


einarf:
I think we might end up building on the experimental camera for this. There's already some work done to support it in development branch. Still there might be some useful things here. We need to figure this out during the 2.6 development.

einarf:
Thanks for the work on this. However, we wanted to do this with a `Camera` class instad.

In the future, when adding things like this it's a good idea to discuss the change before making the RP.

Issue #738, opened 2020-08-22
Autoformat Arcade using Black


benjamin-kirkbride:
I really like the [Black](https://github.com/psf/black) code formatter, and think it would benefit this (and any) project to use it.

Is this something that we would be open to? If so, I could prepare a pull request that `black`s the existing codebase, and add it as a dev dependency, so we could see what that would look like?

einarf:
I've formatted most of the `arcade.gl` module using black. Definitely like it.

eruvanos:
I just started using black a few days ago, have to change my behavior to use double quotes, but already getting used to it.

Cleptomania:
I've been experimenting with this through my update of the Sound module.

From what I've noticed, the main problem with Black formatting is that it is very adamant about imports happening before anything else. Due to the nature of a number of things in the library, there are some instances where code needs to run before importing something.

For example, in order to load ffmpeg binaries properly within the new sound update. The binaries must be added to the system path before Pyglet is imported.

I'm sure this could be overcome through configuration of some sort, but something to think about.

Cleptomania:
We've since landed on a philosophy of "format as we go" with regard to modules within Arcade's codebase.

However have decided that auto-formatting should not be used on example code as the docs do specific highlighting and/or mentioning of line numbers and auto-formatting can very easily unintentionally break that.

PR #796, opened 2020-11-30
Use black autoformatter

Cleptomania:
This PR is spawned from https://github.com/pythonarcade/arcade/issues/738

The [black autoformatter](https://github.com/psf/black) has been run against the entire codebase in this PR, so there's a big change list.

The goal with this would be going forward that the formatting is handled editor side as things are developed(and could also be done with CI in the future), most editors support autoformatting on save or something similar without much hassle.

I think using Black will help make sure our code is more readable regardless of who's writing it, rather than 10 different people having 10 different ways to format the same thing.

If you haven't yet used Black, you'll likely find it's very easy to use with autoformat in your editor, I never give a second thought to how I format code as I'm writing as Black just takes care of it for you.


benjamin-kirkbride:
I am strongly in favor of having the code be Black formatted, but I think it's a mistake to not at least enforce it at the CI level, if not autoformat it at the CI level.

People will forget, or not know, or not care, and then the next time someone touches the file that uses Black like you are supposed to, it will change a ton of lines that have nothing to do with the PR they are doing.

arcade/sprite_list.py
```diff
@@ -589,16 +590,24 @@ def _calculate_sub_tex_coords():
...
+                    tmp.paste(tmp.crop((1, 1, image.width + 1, 2)), (1, 0))
+                    tmp.paste(
+                        tmp.crop((1, image.height, image.width + 1, image.height + 1)),
+                        (1, image.height + 1),
+                    )
+                    tmp.paste(tmp.crop((1, 0, 2, image.height + 2)), (0, 0))
+                    tmp.paste(
+                        tmp.crop((image.width, 0, image.width + 1, image.height + 2)),
+                        (image.width + 1, 0),
+                    )
```
einarf, line 610: This is a typical case were black makes the code a bit less readable imho.

pvcraven:
There are tutorials, and examples that refer to and highlight soecific line
numbers of code. If you do this PR, you have to validate those are all
still correct.

arcade/gl/geometry.py
```diff
@@ -81,125 +125,316 @@ def cube(size=(1.0, 1.0, 1.0), center=(0.0, 0.0, 0.0)) -> Geometry:
...
+            1,
+            0,
+            0,
+            1,
+            0,
+            0,
+            1,
+            0,
+        ],
+    )
```
einarf, line 432: This is a bit much 😄

arcade/gl/geometry.py
```diff
@@ -20,53 +20,97 @@ def quad_2d_fs() -> Geometry:
...
+            x_pos + width / 2.0,
+            y_pos + height / 2.0,
+            1.0,
+            1.0,
+            x_pos + width / 2.0,
+            y_pos - height / 2.0,
+            1.0,
+            0.0,
+        ],
+    )
```
einarf, line 57: Borderline case. The first version is a bit more readable imho.

arcade/experimental/examples/tesellation.py
```diff
@@ -88,19 +87,19 @@ def __init__(self, width, height, title):
...
+                -1.0,
+                0.0,
+                -0.5,
+                1.0,
+                0.5,
+                -1.0,
+                1.0,
+                0.0,
+            ],
+        )
```
einarf, line 99: A bit awkward considering the comment indicate each line respresents an x and y value.

arcade/experimental/examples/instancing.py
```diff
@@ -60,17 +59,23 @@ def __init__(self, width, height, title):
...
+            [
+                # x, y
+                0.0,
+                0.8,
+                -0.8,
+                -0.8,
+                0.8,
+                -0.8,
+            ],
+        )
```
einarf, line 78: I bit weird considering the comment is still there. When specifying geometry it can be nice to have have a primitive per line.

arcade/experimental/examples/geometry_index_buffers.py
```diff
@@ -36,53 +35,82 @@ def __init__(self, width, height, title):
...
+                    1.0,
+                    1.0,  # upper left
+                    0,
+                    0,
+                    0,
+                    0,
+                    0,  # dummy data for testing
+                ],
+            )
+        )
```
einarf, line 74: Might need some cleanup here? I personally think a primitive per line is more readable in this case.

arcade/experimental/examples/chip8_display.py
```diff
@@ -62,55 +61,280 @@ def __init__(self, width, height, title):
...
+                    0x02,
+                    0x55,
+                    0x55,
+                    0xCC,
+                    0x99,
+                    0xBD,
+                    0x63,
+                    0x00,
+                    0x01,
+                ],
```
einarf, line 335: Holy 😄  ...

Cleptomania:
Re-doing this a different way through CI with Github actions with proper configuration options.

PR #798, opened 2020-12-01
Autoformat code with Black

Cleptomania:
This PR is for a second attempt at formatting Arcade's codebase using [Black](https://github.com/psf/black)

This has been setup to handle formatting with CI through Github Actions. It works by first checking the codebase to see if Black would make any changes to it. If no changes would be made, then the action passes and nothing is done. However if it finds changes to be made, then it will commit those changes to the repo with the message "cleanup: Format Python code with Black".

If the changes are from a PR, it will commit those changes as a commit to the PR, so this will work for handling code from coming Pull Requests as well as direct commits. 

The checks in the CI, as well as anyone running black locally is being configured by the `pyproject.toml` file. Black respects this file by default out of the box. Currently it is setup to exclude any directories with the word `example` found in them, as well as any file named `__init__.py`. These are regex checks that black performs on every discovered python file. 

Moving forward it would be best if code was formatted using black before being committed, however this setup ensures that it will happen regardless of if people are running it locally, so that we can hopefully have a consistent formatting across the codebase.

You can see in this PR the Black formatting check was ran against the PR. Hopefully this could serve as a jumping off point to add in some more controls for linting, and possibly in the future running tests. It would help ensure PRs submitted from outside the kind of core team from forks will still maintain the same format and consistency with one another.


einarf:
We decided to slowly apply back in the development branch instead 👍

Issue #801, opened 2020-12-08
Headless mode for machine learning/CI usecases


brandonmkunkel:
## Enhancement request:

Optional headless mode which takes advantage of pyglet's ``visible`` parameter which does not draw the window upon Window() instantiation. 

### What should be added/changed?

Add ``visible`` boolean optional keyword argument to the application.Window.__init__()

Starting at line 53 in application.py

```
class Window(pyglet.window.Window):
    """
    The Window class forms the basis of most advanced games that use Arcade.
    It represents a window on the screen, and manages events.
    """

    def __init__(self,
                 width: int = 800,
                 height: int = 600,
                 title: str = 'Arcade Window',
                 fullscreen: bool = False,
                 visible: bool = True,
                 resizable: bool = False,
                 update_rate: Optional[float] = 1 / 60,
                 antialiasing: bool = True,
                 screen: pyglet.canvas.Screen = None):
```

And mody the call to super() within the __init__ function, at line 86.

```        
        try:
            super().__init__(width=width, height=height, caption=title,
                                    resizable=resizable, config=config, visible=visible, vsync=False)
```

### What would it help with?

Machine learning use cases where a headless mode is needed such that windows are not created each time an instance of the . Additionally, would be useful for CI pipelines for testing.

einarf:
We also already wrap the `set_visible` method in the pyglet window. All looking great.

pvcraven:
Will be set with release 2.5.1 later today.

PR #802, opened 2020-12-08
Add ``visible`` kwarg to Window constructor

brandonmkunkel:
Addresses issue #801


pvcraven:
'visible' needs to go at the end of the list, so that this does not break API changes for those people assuming a parameter order.

brandonmkunkel:
Good point, I'll make that change.

einarf:
Looks good. I'll sneak in some docstrings after merge.

PR #818, opened 2021-02-01
AnimationSprite class 1st try ^^

Romuald78:
It is possible to create time based sprite animations :
- flip H and V
- apply color filter for each animation
- sets a limited number of animation loops
- set back and forth animation
- resize final animation at loading
- use several animations for each facing direction

needs to be tested more but it seems to be pretty functional ;)
Enjoy !!

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -1081,8 +1081,8 @@ class AnimatedSprite(Sprite):
         :return: tuple (int, int, int) number of textures, number of frames when back and forth is enabled, number of frames when counter is enabled (and back and forth)
         """
         nb_frames = 0
-        if anim_name in self._anims[self.state]:
-            anim_dict = self._anims[self.state][anim_name]
+        if anim_name in self._anims[self.state-1]:
+            anim_dict = self._anims[self.state-1][anim_name]
             nb_frames     = len(anim_dict["texture_list"])
             nb_frames_baf = nb_frames
             nb_frames_cnt = nb_frames
@@ -1100,8 +1100,8 @@ class AnimatedSprite(Sprite):
         """
         frame_idx  = 0
         frame_perc = 0
-        if anim_name in self._anims[self.state]:
-            anim_dict = self._anims[self.state][anim_name]
+        if anim_name in self._anims[self.state-1]:
+            anim_dict = self._anims[self.state-1][anim_name]
             # Get number of frames
             nb_frames, nb_frames_baf, nb_frames_cnt = self._get_nb_frames(anim_name)
             # compute absolute frame index according to time
@@ -1141,7 +1141,7 @@ class AnimatedSprite(Sprite):
         #     + back_and_forth : bool
         #     + counter : int
         #    }
-        self._anims = [{}] * 4
+        self._anims = [{},{},{},{}]
 
         # Current animation name
         self._current_animation_name = None
@@ -1157,8 +1157,6 @@ class AnimatedSprite(Sprite):
     def add_animation(self,
                      animation_name: str,
                      filepath: str,
-                     final_width: int,
-                     final_height: int,
                      nb_frames_x: int,
                      nb_frames_y: int,
                      frame_width: int,
@@ -1188,8 +1186,6 @@ class AnimatedSprite(Sprite):
         size ratio and input image ratio.
         :param int nb_frames_x: number of frames in the input image, along the x-axis
         :param int nb_frames_y: number of frames in the input image, along the y-axis
-        :param int frame_width: width of each frame. frame_width*nb_frames_x must be equal to the input image width.
-        :param int frame_height: width of each frame. frame_height*nb_frames_y must be equal to the input image height.
         :param int frame_start_index: index of the first frame of the current animation. Indexes start at 0. \
         Indexes are taken from left to right and from top to bottom. 0 means the top-left frame in the input image.
         :param int frame_end_index: index of the last frame for the current animation. this value cannot exceed (nb_frames_x*nb_frames_y)-1.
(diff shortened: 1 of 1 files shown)


arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+        self._current_animation_name = None
+        # Current displayed texture
+        self.cur_texture_index = 0
+        # Set elapsed duration (used to know if we have to stop the animation)
+        self._elapsed_duration = 0
+        # Set play/pause flag
+        self._playing = True
+
+
+    def addAnimation(self,
```
einarf: Should probably use snake case for all method instead of camel case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+
+            # In case of backAndForth
+            if frameIdx >= nbFrames:
+                frameIdx = nbFrames2 - frameIdx
+
+            # set current texture index and texture from the parent class
+            self.cur_texture_index = frameIdx
+            self.set_texture(self.cur_texture_index)
+
+    def pauseAnimation(self):
```
einarf: use snake case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+        """
+        Pauses the current animation. It does not rewind it.
+        :return: None
+        """
+        self._playing = False
+
+    def resumeAnimation(self):
```
einarf: use snake case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+        """
+        Resumes the current animation. It does not rewind it.
+        :return: None
+        """
+        self._playing = True
+
+    def rewindAnimation(self):
```
einarf: use snake case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+        """
+        Just rewinds the current animation to the first frame. It does not change the play/stop flag.
+        :return:
+        """
+        self._elapsed_duration = 0
+
+    def playAnimation(self):
```
einarf: use snake case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+        """
+        Rewinds and Plays the current animation.
+        :return: None
+        """
+        self.rewindAnimation()
+        self.resumeAnimation()
+
+    def stopAnimation(self):
```
einarf: use snake case

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
+
+    """"
+    New class for Sprite animation. (trial)
+    """
+
+    # static integer for unique texture ids
+    _uniqueID = 1
+
+    @staticmethod
+    def _getNextID():
```
einarf: maybe something like `_next_id()`

arcade/sprite.py
```diff
@@ -1054,6 +1054,243 @@ def update_animation(self, delta_time: float = 1/60):
...
 
+
+class AnimatedSprite(Sprite):
+
+    """"
+    New class for Sprite animation. (trial)
+    """
+
+    # static integer for unique texture ids
+    _uniqueID = 1
```
einarf: snake case

einarf:
There are many places were camelCase is used here. We like to use snake case in python. 

If this new class is something arcade needs I'll leave up to @pvcraven 😄 (I mainly work on the lower level part of the library)

Romuald78:
I didnt checked AnimatedTimeBaseSprite class, but is seems the class i provided is something similar/redundant/reworked 
I tried to use the texture list from Sprite class, and the state field for the 4 facing directions, so i really hope it can be useful for arcade 😉
Anyway i ll update according to the naming conventions

Romuald78:
Reminder : add a method is_finished() in order to knowif the animation of a Sprite is over (in case game logic wants to execute some code at this moment)
Reminder2 : ads a method get_percent or similar to get the progression of the animation in case the game logic wants to deal with it
Reminder3 : add the way to select specifically the texture index, get the number of current index etc... pairing select texture and stop can lead to have a fixed sprite with multi displays

Romuald78:
@pvcraven i guess you have a busy planning and limited time to offer for reviews, but if you have time to look at my code, i would appreciate.
I haven't put any "example" code (i have on local for the moment) but i guess this is better to push examples too
Tell me :)
C U

pvcraven:
Thanks for the contribution and the revisions.

Took a first pass looking at it. The thing that bothers me with creating 'yet another' animated sprite is I've already burned through 4-5 versions of this concept. I hate to add more without it becoming a clear improvement. My last couple attempts I didn't add to the code base.

The only one we have that's actively used is for supporting animated tiles loaded from Tiled. They are simple and time based. Past that I'd recommend people try their own because I'm not confident of a good 'common' solution yet.

Romuald78:
Raah ! So sad 😉
It was the first step to an evolution toward the use of TexturePacker, a tool i used with PhaserJS to easily integrate sprite animations
Anyway thx for the review 🙂

pvcraven:
I'd keep at this, as a good solution would be great. Demoing it with some good proof-of-concept code, and maybe even a concept game would be a nice way to make sure that this is something that would survive the test of time.

Romuald78:
Ok @pvcraven, do you mean it is worth for me to take time to create a real demo code , or that wont change much, and you dont plan to integrate it anyway
Excuse me for the question but my english level may be poor with some word/expression meanings

PR #819, opened 2021-02-03
AnimatedSprite .... the return ^^

Romuald78:
Ok here is a class that creates animations from spritesheets !
There is an example code in the examples folder called animated_sprite.py
Some images have been added too in the resources/images/spritesheets folder

This class does not extend the Sprite class anymore but there is composition link with Sprite class.
Each time an animation is added, a Sprite object is created inside the AnimatedSprite object.
This allows to avoid the texture drawing issue when all frames have not the same size.
The limitation to have all the frames the same size are only limited to one animation, and this is mostly the case

The different parameters/features are :
- add as many animations you want. You must name them in order to select them later. the animation names must be unique
- frame size, and number of frames on each axis
- start index of frame and end index of frame. IF start is greater than end, the frames will be loaded in the reverse way : that allows to create different animations from the same picture file
- counter : allows to limit the number of times the animation will be played. If 0, the animation loop is infinite.
- back and forth :  a boolean to tell if the animation must draw the frames in both ways. This do not create more textures, it simply update the frame index
- flip horizontal and vertical : the whole animation frames will be flipped according to these two booleans
- filter color : each animation can have its own color filter (this is the one applied in the Sprite class)
- facing direction : by default there is only one facing direction, but when instanciating the AnimatedSprite class, you can give a parameter to tell how many directions you want. Then you can add an animation for each direction, using the same name to tell this is the same "action". The pair "facing_direction" + "animation_name" will be used to know which frame must be drawned. You can now select the animation you want + select the state you want. State values are not the ones from the sprite.py file : these are up to the user of AnimatedSprite to define the values of each "direction".


Romuald78:
- From a AnimatedSprite ref, you can get the current animation name, and you can know if this is finished or not (in case of limited counter animations). In fact you can get the percentage in the current animation, from  to 1. This information can be useful in case of game mechanics (e.g. a character is slashing with his weapon and you want the opponent to be hit only during a certain period of the animation, when the weapon is "really" hitting )

Romuald78:
This feature allows the game mechanics to wait for the end of animations to execute some code.

About that, I am thinking about a "registering system", to any animation, at any percentage of progression : the AnimatedSprite instance, when updated, would call the registered callbacks in order to notify about this event.
...just an idea ..... for the future ;)

Romuald78:
Updated the AnimatedSprite class to use @property for center_x / center_y / scale / angle / state, so these objects can be used almost like any Sprite object

PR #823, opened 2021-02-21
Pytiled parser rewrite

Cleptomania:
Completely re-works the tilemap module's use of PyTiled Parser to work with the new re-written version of PyTiled.

There is only one API change in this, which is just a renamed function. It still follows the exact same structure so all that needs done in that regard is updating usage of the `arcade.read_tmx` function to `arcade.read_map`. The reason for the name change is that with this update we no longer support loading of TMX map files. All Tiled maps will need to be loaded as JSON now. This is extremely simply to do from within Tiled, if you have an already existing TMX map, you simple open it in Tiled and click "File -> Export As" and choose the JSON type. Exact same process for TSX tileset files. 

Enforcing people change the function name will help ensure they fully understand that they need to convert any old maps.

This process can also be done using Tiled's CLI so could easily be automated if someone had a lot of maps/tilesets to convert. 

All existing examples and unit tests have been updated for this change.

It should also be noted, in case anyone is curious or questions why we've dropped support for TMX, there is no drawback to the Tiled JSON file, both TMX and JSON support the same feature sets, just in different formats. It should be no secret that JSON is much easier to deal with than XML.


pvcraven:
How is this doing? Is it about ready to be merged in to dev?

Cleptomania:
I would say it’s ready to be merged in. I’d like to build out some more functionality or something down the line, but that probably involves re-working a lot of the existing API/functionality.

As this stands there’s just a little bit of extra functionality which is covered in an addition to the platformer tutorial.

Everything is the same API wise with the exception of changing the `read_tmx` function to `read_map`. Think it will be good to force this change so people become aware of needing to convert their maps to JSON from TMX.

All docs and tests have been updated with the changes in this already

PR #825, opened 2021-02-22
fixed bug in __setitem__ that shortened the sprite_list

lwanger:
Fixed bug in __setitem__ that removed the item and made the sprite list shrink. Old version broke Solitaire example. This patch fixes that issue.

### DIFF
--- a/arcade/sprite_list.py
+++ b/arcade/sprite_list.py
@@ -16,6 +16,7 @@ import logging
 import math
 import array
 import time
+from random import shuffle
 
 from PIL import Image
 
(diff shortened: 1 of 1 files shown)


lwanger:
Also added a shuffle method to the SpriteList class. Shuffle uses the standard library shuffle function. It's a handy method to have for lots of games. Here's how it simplifies the solitaire tutorial:

code was:

        # Shuffle the cards
        for pos1 in range(len(self.card_list)):
            pos2 = random.randrange(len(self.card_list))
            self.card_list[pos1], self.card_list[pos2] = self.card_list[pos2], self.card_list[pos1]

and becomes:

        self.card_list.shuffle()

PR #827, opened 2021-02-24
Arcade 2.5.5 : AnimatedSprite class + example

Romuald78:


pvcraven:
Going to age this out, still not settled on an animated sprite method.

PR #833, opened 2021-03-23
UI development

eruvanos:
This PR contains the new UILayout classes which provide a more convenient way to place UIElements.

Furthermore it comes with examples to create game menues, hud like inventory and inventory windows.

Goal is to get feedback before the PR is merged!


pvcraven:
How are you feeling on this? It is looking good.

I do get an exception running `show_uilayouts_start_menu.py` and hitting 'exit'

PR #848, opened 2021-04-18
FIx OutOfMemory due to rapid UILabel text changes

eruvanos:
This provides a workaround for the OutOfMemory, until these problems are taken care of (within the development branch).


einarf:
We dicussed this on voice today. It should not impact existing sprite lists and is only a small hack to deal with the excessive texture allocation. It's an important bug fix.

PR #853, opened 2021-04-20
Add Kenney.nl ttfs with spaces and Kenney prefix removed.

### LINKED ISSUE #852: Ship Libre fonts with arcade in the resource folder
Ship default fonts available under MIT, public domain, or other compatible licenses inside the resources folder

#### tl;dr

It would immediately:
1. Homogenize user experience across platforms
2. Support upcoming tickets related to text rendering (#850 , #851 , #842 )
3. Enable simplified font loading logic

Further down the line, consistent fonts could increase cross-platform reliability of tests that compare render results against reference. 

##### 1. Homogenize user experience across platforms
At the moment, font loading is complicated. It falls back on a number of predefined font paths and resolution methods.
The majority of them attempt to load some variation of Arial, but two other fonts are included as well. Because Arial is proprietary, not all systems will have it. [Other open source projects have noted users perceive applications as broken or low-quality when fonts look different across platforms](https://blog.documentfoundation.org/blog/2020/09/08/libreoffice-tt-replacing-microsoft-fonts/). 
We can improve user experience for both players and developers by shipping quality built-in defaults for beginner to intermediate users. Given arcade seems to have educational goals in mind, this would also help adoption in an academic settings.

##### 2. Support upcoming tickets related to text rendering (#850 , #851 , #842 )
Having reliable fonts will not only help our games look better, but they will give us something to feed into the new font rendering systems that we are building.

##### 3. Simplified font loading logic
If we always know what folder to find a default font in, we don't have to search through multiple locations in system paths.

#### Later: More reliable render result comparison testing
The GUI tests appear to compare render output to reference screenshots. If we homogenize the test fonts across operating systems, test reliability should go up. Minor differences in platform font rendering with the same fonts will have less of a failure rate than entirely different different fonts loaded on different operating systems or Linux package configurations.  For example, some especially purist Libre-only distros make it difficult for beginners to install non-Libre packages by omitting repos hosting them from the default repository lists. Font packages including Arial would fall under this category.

pushfoo:
Closes #852 . This only places fonts in the folders. It doesn't integrate them into loading in any way yet. This could also be merged straight into development according to github. 

Either way, refactoring font loading can be split into another ticket and probably needs to be discussed in more depth beforehand.


pushfoo:
One question: is it worth keeping `Kenney` as a prefix to avoid name collisions during lookup?

PR #877, opened 2021-05-19
Update setup.py so Arcade will work on Replit

Mr-Coxall:
I am trying to use Arcade with my students on Replit.com.
Replit will let PyGame work, so Arcade should work.
When I try to load Arcade I get the following error:
The current project's Python requirement (^3.8) is not compatible with some of the required packages Python requirement:
  - arcade requires Python >=3.6,<3.10
Replit does not like the "<3.10".
I do not think there is any need to have "<3.10" in Arcade and if you were willing to remove it, then Arcade should be able to work on Replit.

Here is an example Replit that will run PyGame but not load Arcade correctly:
https://replit.com/@mrcoxall/Python-Arcade-1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import find_namespace_packages, setup
 if sys.platform == "darwin":
     required_python_version=">=3.6,<3.9"
 else:
-    required_python_version=">=3.6,<3.10"
+    required_python_version=">=3.6"
 
 exec(open("arcade/version.py").read())
 


pvcraven:
Arcade needs a higher version of open gl than I thing repl can pull off.

einarf:
It's might still be a good thing to remove the `<3.10` part from requirements

Cleptomania:
I believe I saw this previously where that syntax was not understood Poetry or something like that. I don't see any harm in doing it. I think it was originally put in place because Numpy was causing issues with 3.10 compatibility so we put it in there to block that.

Anyways we've removed numpy as a dependency in the development branch so that's not gonna present any issues.

In regards to Replit support last any of us had checked/heard of I believe Arcade was not compatible due to OpenGL dependency

einarf:
A quick check in repl.it : `glGetString(GL_VERSION)` returned `'3.1 Mesa 20.0.8'` so it's getting closer to support arcade.

It might be worth looking into if we can get it running in repl.it in the near future. That might be more on the pyglet side.

einarf:
The exact output in repl.it
```
--> python3 -m poetry add arcade
Using version ^2.5.6 for arcade

Updating dependencies
Resolving dependencies...

[SolverProblemError]
The current project's Python requirement (^3.8) is not compatible with some of the required packages Python requirement:
  - arcade requires Python >=3.6,<3.10

Because no versions of arcade match >2.5.6,<3.0.0
 and arcade (2.5.6) requires Python >=3.6,<3.10, arcade is forbidden.
So, because arcade-test depends on arcade (^2.5.6), version solving failed.
exit status 1
```

einarf:
Created this issue to possibly expore how arcade could work on repl.it : https://github.com/pythonarcade/arcade/issues/878

einarf:
Arcade actually works on repl.it. See : https://github.com/pythonarcade/arcade/issues/878#issuecomment-844534009

einarf:
I suggest we merge this and cherrypick the commit into `maintenance` branch and release a 2.5.7 some time next week.

einarf:
Let's just merge this for now 😄

Issue #888, opened 2021-06-05
`YELLOW_GREEN` in arcade/csscolor/__init__.py is incorrect.


amosnier:
`YELLOW_GREEN` in arcade/csscolor/__init__.py is incorrect. The blue component is missing.

MisoMosiSpy:
I verified that this need to be changed from 

YELLOW_GREEN = (154, 205)

To 

YELLOW_GREEN = (154, 205, 50)

Just starting with arcade, can I take up this issue?

pvcraven:
Yes, it is a bug. Yes, please go ahead and take up the issue! Please make changes in the development branch.

MisoMosiSpy:
Thank you, I have created the pull request, hope I got everything right.

Cleptomania:
Released with 2.6.0

PR #890, opened 2021-06-06
Corrected definition of CSS color YELLOW_GREEN.

MisoMosiSpy:
Fix for issue #888.
Values cross checked from https://www.w3schools.com/cssref/css_colors.asp.


pvcraven:
Thank you for the PR!

PR #892, opened 2021-06-07
Merge ui_development into development

eruvanos:
Still some tests failing, but I would need some help with them


einarf:
Think that would be fine to merge. Let's just get that down into development and work in remaining issues.

Issue #921, opened 2021-08-01
dev branch : global_options for TileMap


EnlNovius:
## Enhancement request:

Have a global_options option for the TileMap class that allows to set a default value to all layers that can be overridden by layer_options if needed.  

I have already set up an implementation, do I have to make a pull request? If not, I will propose an implementation here.

EnlNovius:
Pull request sent =)

Cleptomania:
This is merged in and a few minor adjustments to make it fit in better with the rest of the class and maintain a bit of backward compatibility.

The new order of attributes is `load_tilemap(map_file, scaling, layer_options, use_spatial_hash, hit_box_algorithm, hit_box_detail)`. Going with this order because a lot of example code already assumes that `scaling` is the second arg and `layer_options` is the third. This also makes sense as most use cases you probably don't want to enable things like spatial hash on every layer, or are happy with the defaults for hit box algorithm alg and detail.

Not saying this isn't useful though, there's definitely cases where this will save some code from the game's perspective.

Thanks for the suggestion and PR

PR #922, opened 2021-08-01
Adding global_options argument to TileMap class

EnlNovius:
Proposed code for the issue 921: [https://github.com/pythonarcade/arcade/issues/921](https://github.com/pythonarcade/arcade/issues/921)

### DIFF
--- a/arcade/tilemap/tilemap.py
+++ b/arcade/tilemap/tilemap.py
@@ -107,6 +107,7 @@ class TileMap:
         map_file: Union[str, Path],
         scaling: float = 1.0,
         layer_options: Optional[Dict[str, Dict[str, Any]]] = None,
+        global_options: Optional[Dict[str, str]] = None
     ) -> None:
         """
         Given a .json file, this will read in a Tiled map file, and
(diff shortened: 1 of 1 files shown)


Cleptomania:
I think a better approach is to set these as attributes of the class individually. We already have `TileMap.scaling` which provides a global default for scaling of a layer specific one isn’t set. I think any global options should follow this same structure. It’s much easier to document and for people to find what options are available this way.

the dict way for layer specific options was only really handled that way out of necessity

We could take these in the constructor, and then construct a dict for each layer with the values from the layer options dict or insert the global option if it’s not there.

EnlNovius:
Something like this in the constructor parameters ? 

```
use_spatial_hash: bool = False,
hit_box_algorithm: str = "Simple",
hit_box_detail: float = 4.5,
```

Or should these parameters be optional?

EnlNovius:
New commit with parameters.

By the way, `PIL` is import but probably not used in the tilemap.py

Cleptomania:
I'm merging this but I'm going to change a little bit how it applies the global values to the layers. API still the same though.

PR #943, opened 2021-09-05
Tile map custom class

### LINKED ISSUE #942: Custom Classes for layers of Tile Map [Enchantment/Feature request]
## Enchantment/Feature request
There should be a way, to define Custom classes for each Layer within a TileMap. All tiles of said layer would be created from that custom class instead of the default Sprite.

CommentatorForAll:
fixes #942 
changed tile map to accept a custom_class layer attribue (default Sprite). When reading in the tilemaps, all Tiles of that layer will be created from that class.

### DIFF
--- a/arcade/pymunk_physics_engine.py
+++ b/arcade/pymunk_physics_engine.py
@@ -35,7 +35,15 @@ class PymunkPhysicsEngine:
     KINEMATIC = pymunk.Body.KINEMATIC
     MOMENT_INF = float('inf')
 
-    def __init__(self, gravity=(0, 0), damping: float = 1.0):
+    def __init__(self, gravity=(0, 0), damping: float = 1.0, maximum_incline_on_ground: float = 0.708):
+        """ 
+            Creates a new physics engine, based on pymunk
+            
+            :param gravity The direction where gravity is pointing
+            :param damping The dampning of the sprites
+            :param maximum_incline_on_ground The maximum incline the ground can have, before is_on_ground() becomes False
+                default = 0.708 or a little bit over 45° angle
+        """
         # -- Pymunk
         self.space = pymunk.Space()
         self.space.gravity = gravity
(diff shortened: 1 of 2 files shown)


Cleptomania:
Thanks for the PR on that, I’ll get it merged in sometime tonight probably. I’m going to expand this to be able to set it on a per layer basis as well.

CommentatorForAll:
This is already on layer basis. You can define it for each layer seperately.
The only case which doesnt work is animated sprites. That one is still created from the original animatedspeite one

Cleptomania:
Okay I see now, one potential issue here is that this assumes your custom class constructor expects the same things as the standard Sprite class. I'm certain this is something that would vary a lot from case to case, so I'm not sure I see a great solution to that, any thoughts on that?

One possible way to handle it could be having a `custom_class_args` field in `layer_properties` which is just a dict and we pass it into the constructor with the `**` operator. We would still have to provide the same base args as we do for Sprite, but at least this would let you have additonal parameters.

This would mean if for example you had two layers of Platforms, which derive from the same base class, but maybe they're destructible and have different health values, you might do something like:

```python
layer_options = {
    "low_health_platforms": {
        "custom_class": MyPlatform,
        "custom_class_args": {
            "health": 50
        }
    },
    "high_health_platforms": {
        "custom_class": MyPlatform,
        "custom_class_args": {
            "health": 100
        }
    },
}
```

This still requires your `__init__` of the custom class to accept the regular stuff that `arcade.Sprite` does though, but that might be fine, even if you wanna actually create the Sprite differently or something, you could just make it accept the values and have them be thrown away. I don't really see a better way to handle that, any ideas?

CommentatorForAll:
That is a valid argument. I was tinkering with that too, because beforehand, i had custom argumetns too. But i figured that it would be best if one assumes, that the custom constructor is compatible with the default sprite constructor. but yeah, custom argumetns would be nice ^^

Cleptomania:
Gonna merge this in now, I'll do some more testing of the custom class args and see if we need any tweaks on the dev branch then

PR #963, opened 2021-09-20
Stop saying that non-method functions are methods

### LINKED ISSUE #957: Sometimes “method” is used when “function” would be more correct
The comments in some of the example code and tutorials.

### Where is it located?
In the `arcade/examples` and `doc/tutorials` folders. I haven’t found all of the files that need changing yet, but here’s one of them: `arcade/examples/conway_alpha.py`.

[In Python, a method is a specific type of function.](https://docs.python.org/3/glossary.html#term-method) If a function is declared in a class, then it is a method. If a function is declared outside of a class, then it isn’t a method. Sometimes, the comments call a function a method when it isn’t actually a method.

Jayman2000:
In Python, a method is a specific type of function[1]. Before this
change, comments would sometimes say that a function was a method even
when it wasn't.

Fixes #957.

1. https://docs.python.org/glossary.html#term-methods

### DIFF
--- a/arcade/examples/astar_pathfinding.py
+++ b/arcade/examples/astar_pathfinding.py
@@ -262,7 +262,7 @@ class MyGame(arcade.Window):
 
 
 def main():
-    """ Main method """
+    """ Main function """
     window = MyGame(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
     window.setup()
     arcade.run()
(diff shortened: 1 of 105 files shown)


einarf:
Leaving this one for @pvcraven since it changes exmaple code.

pvcraven:
This is good. Merged.

PR #974, opened 2021-09-30
Cache generated textures

### LINKED ISSUE #973: Cache generated textures for SpriteCircle and SpriteSolidColor

**tl;dr** [`arcade.SpriteSolidColor`](https://github.com/pythonarcade/arcade/blob/d5ef3c5cd187efe199b844da9977bb3062a41383/arcade/sprite.py#L1262) and [`arcade.SpriteCircle`](https://github.com/pythonarcade/arcade/blob/d5ef3c5cd187efe199b844da9977bb3062a41383/arcade/sprite.py#L1283) should cache the auto-generated textures for each combination of constructor arguments they recieve.

Currently, instantiating either class generates an entirely new copy of a texture using PIL. This can cause slowdowns when many instances of `SpriteCircle` or subclasses are instantiated because PIL uses CPU-bound drawing to generate the textures. This is especially evident at larger sizes for soft circles. 

This example demonstrates the texture growth:
```py
from collections import Set

import arcade


class TextureAllocationExample(arcade.Window):
        self.sprites: arcade.SpriteList = arcade.SpriteList()

        # track individual ID numbers
        self.tex_id_numbers: Set[int] = set()
        self.tex_pil_image_id_numbers: Set[int] = set()
        # track texture names
        self.tex_names: Set[str] = set()

        sprite = arcade.SpriteCircle(5, arcade.color.GREEN)

        self.tex_id_numbers.add(id(sprite))
        self.tex_pil_image_id_numbers.add(id(sprite.texture.image))
        self.tex_names.add(sprite.texture.name)

        self.sprites.append(sprite)
       
        # repeated copies of the same texture are allocated
        print(
            f" tex_id_numbers {len(self.tex_id_numbers)}"
            f" tex_pil_image_id_numbers {len(self.tex_pil_image_id_numbers)}"
            f" tex_names {len(self.tex_names)}"
        )

if __name__ == "__main__":
    window = TextureAllocationExample(640, 480, "Texture allocation example")
    arcade.run()
```

Instead, these textures should be cached and re-used for instantiations of the class that have the same parameters. A final fix could use the global texture cache in `arcade/texture.py`, or a local per-class cache like this workaround I have in my prototype:
```py
class CachedSpriteCircle(arcade.Sprite):

    # quick fix until fix is made for libraries
    PARAM_CACHE: Dict[Tuple, arcade.Texture] = {}

    def __init__(self, radius, color: Color):
        super().__init__()
        diameter = int(radius * 2)
        params = (diameter, color)
        if params not in self.PARAM_CACHE:
            self.PARAM_CACHE[params] = make_circle_texture(diameter, color)

        self.radius = radius
        self.base_color = color

        self.texture = self.PARAM_CACHE[params]
        self._points = self.texture.hit_box_points
```

If a quick fix to use caches to limit PIL calls is ok, I can take this ticket.

### What would it help with?

**tl;dr** improve performance and instill good game prototyping habits in students

#### Performance

Early on in the development process, people often want to prototype gameplay without concern for assets. I did this myself recently when I used SpriteCircle to create asset stand-in projectiles and enemies for a pymunk based prototype. That lead to occasional slowdowns when I fired multiple projectiles per second or spawned many enemies in a single frame. 

#### Build Focus, Good Habit, and Project Momentum

As Arcade aims to be an education-friendly framework, it would be a good idea to instill good game prototyping habits into students early. One common problem students and other aspiring developers frequently run into is that they focus on assets before gameplay, and get overwhelmed or miss deadlines as they try to tweak the perfect assets into being. Performance issues early on may also discourage new developers. Improving the performance of these classes will help students prototype effectively and keep project momentum high.

pushfoo:
Fixes #973 and should improve performance and memory usage in many cases.

I think further optimizations for `SpriteSolidColor` (stretching images to cover rectangles of any size and increasing hitbox) should be put in another PR. The current changes cover #973 and already increase performance whenever someone is creating a lot of `SpriteCircle`s and `SpriteSolidColor`s. 

I think the changes are suitable for release in a minor version with performance improvements. Although the internal sprite cache names were never promised to be part of the stable external API, I tried to generally keep the type prefix for each generated name (ex: "circle_texture_soft") unchanged in case someone might need to search for it during debugging.

Let me know if I should squash down commits.

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -7,6 +7,8 @@ https://www.gamedev.net/articles/programming/general-and-gameplay-programming/sp
 
 import math
 
+from arcade.texture import build_cache_name
+
 try:
     import dataclasses
 except ModuleNotFoundError:
@@ -1295,10 +1297,29 @@ class SpriteCircle(Sprite):
     def __init__(self, radius: int, color: Color, soft: bool = False):
         super().__init__()
 
+        diameter = radius * 2
+
+        # determine the texture's cache name
         if soft:
-            self.texture = make_soft_circle_texture(radius * 2, color)
+            cache_name = build_cache_name("circle_texture_soft", diameter, color[0], color[1], color[2])
         else:
-            self.texture = make_circle_texture(radius * 2, color)
+            cache_name = build_cache_name("circle_texture", diameter, color[0], color[1], color[2], 255, 0)
+
+        # use the named texture if it was already made
+        if cache_name in load_texture.texture_cache:
+            texture = load_texture.texture_cache[cache_name]
+
+        # generate the texture if it's not in the cache
+        else:
+            if soft:
+                texture = make_soft_circle_texture(diameter, color, name=cache_name)
+            else:
+                texture = make_circle_texture(diameter, color, name=cache_name)
+
+            load_texture.texture_cache[cache_name] = texture
+
+        # apply results to the new sprite
+        self.texture = texture
         self._points = self.texture.hit_box_points
 
 
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -1277,9 +1277,21 @@ class SpriteSolidColor(Sprite):
         """
         super().__init__()
 
-        image = PIL.Image.new("RGBA", (width, height), color)
-        self.texture = Texture(f"Solid-{color[0]}-{color[1]}-{color[2]}", image)
-        self._points = self.texture.hit_box_points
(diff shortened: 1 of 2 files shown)


pvcraven:
My initial, quick-look-through, makes me think this is a good change.

If someone color-cycles we'll have a memory leak, but that will mess up the atlas anyways. Makes me wonder if there should be docs somewhere that explain to use .color or otherwise explain the internals a bit.

Issue #976, opened 2021-09-30
TypeError: add_sprite() got an unexpected keyword argument 'moment'


alderven:
Run code from tutorial:
https://api.arcade.academy/en/latest/tutorials/pymunk_platformer/pymunk_demo_platformer_12.html#pymunk-demo-platformer-12

Get error here:
```
self.physics_engine.add_sprite(self.player_sprite,
                               friction=PLAYER_FRICTION,
                               mass=PLAYER_MASS,
                               moment=arcade.PymunkPhysicsEngine.MOMENT_INF,
                               collision_type="player",
                               max_horizontal_velocity=PLAYER_MAX_HORIZONTAL_SPEED,
                               max_vertical_velocity=PLAYER_MAX_VERTICAL_SPEED)
```
Error message:
```
Traceback (most recent call last):
  File "C:/pymunk_demo_platformer_12/doc/tutorials/pymunk_platformer/pymunk_demo_platformer_12.py", line 524, in <module>
    main()
  File "C:/pymunk_demo_platformer_12/doc/tutorials/pymunk_platformer/pymunk_demo_platformer_12.py", line 519, in main
    window.setup()
  File "C:/pymunk_demo_platformer_12/doc/tutorials/pymunk_platformer/pymunk_demo_platformer_12.py", line 317, in setup
    self.physics_engine.add_sprite(self.player_sprite,
TypeError: add_sprite() got an unexpected keyword argument 'moment'
```

pushfoo:
1. Try `moment_of_inertia` as the argument name instead of `moment`
2. What version of arcade are you on?

alderven:
> Try `moment_of_inertia` as the argument name instead of `moment`

It works. Thanks!

> What version of arcade are you on?

`2.6.3`

pushfoo:
As noted in #977 , this was caused by an earlier commit that updated type annotations, but also renamed the argument to `PymunkPhysicsEngine.add_sprite` without updating the doc. That ticket should be addressed first so we can update the doc with the correct variable name.

pushfoo:
Duplicate of #932

einarf:
2.6.4 released

PR #979, opened 2021-09-30
Fix typo in PymunkPhysicsEngine.add_sprite arg name introduced in 2.6.2

### LINKED ISSUE #977: Misspelled argument name in PymunkPhysicsEngine
## Bug Report

This was introduced in a variable type annotation improvement that also introduced the root cause of #976 by changing the argument name. This bug should be fixed as a pre-requisite to fixing #976 so the documentation can have the correct spelling. This issue likely wasn't noticed earlier because of auto-completion in IDEs smoothing it over.

### Actual behavior:
`moment_of_intertia` is used as the argument name instead of `moment_of_inertia` (an extra T is added in the word `inertia`).

### Expected behavior:
it should be spelled `moment_of_inertia`. 

Attempt to call `PymunkPhysicsEngine.add_sprite` using the correct spelling of the argument.

pushfoo:
This closes #977 by:
1. Replacing the keyword with the correct spelling, `moment_of_inertia`
2. Adding a backward-compatible alias keyword, `moment_of_intertia`, that is marked as deprecated in docstrings.
4. Adding an `or` to choose whichever of them has been defined, if any, to pass to pymunk

It has been tested locally by running `arcade/examples/pymunk_demo_topdown.py` with both keywords. Both worked.

### DIFF
--- a/arcade/pymunk_physics_engine.py
+++ b/arcade/pymunk_physics_engine.py
@@ -58,7 +58,7 @@ class PymunkPhysicsEngine:
                    mass: float = 1,
                    friction: float = 0.2,
                    elasticity: Optional[float] = None,
-                   moment_of_inertia: Optional[float] =None,
+                   moment_of_inertia: Optional[float] =None,  # correct spelling, easy to google for further explanation
                    body_type: int =DYNAMIC,
                    damping: Optional[float] =None,
                    gravity: Union[pymunk.Vec2d, Tuple[float, float], Vec2] =None,
@@ -67,7 +67,9 @@ class PymunkPhysicsEngine:
                    max_vertical_velocity: int =None,
                    radius: float = 0,
                    collision_type: Optional[str] = "default",
-                   moment_of_intertia: Optional[float] = None
+                   # the next two arguments are for backwards compatibility with prior versions
+                   moment_of_intertia: Optional[float] = None,  # typo keyword, used by 2.6.2 and 2.6.3
+                   moment: Optional[float] = None  # used prior to 2.6.2
                    ):
         """ Add a sprite to the physics engine. 
             :param sprite: The sprite to add
@@ -84,6 +86,7 @@ class PymunkPhysicsEngine:
             :param radius:
             :param collision_type:
             :param moment_of_intertia: Deprecated alias of moment_of_inertia compatible with a typo introduced in 2.6.2
+            :param moment: Deprecated alias of moment_of_inertia compatible with versions <= 2.6.1
         """
 
         if damping is not None:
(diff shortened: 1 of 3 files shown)


arcade/pymunk_physics_engine.py
```diff
@@ -82,6 +83,7 @@ def add_sprite(self,
             :param max_vertical_velocity: maximum velocity on the y axis
             :param radius:
             :param collision_type:
+            :param moment_of_intertia: Deprecated alias of moment_of_inertia compatible with a typo introduced in 2.6.2
```
pvcraven, line 88: This has the extra t in it.

pvcraven:
If you can update with that one correction, this looks good to merge in.

arcade/pymunk_physics_engine.py
pushfoo: That version of the misspelled argument was added at the end of the arguments intentionally to preserve compatibility with 2.6.2. That's why the docstring says deprecated after the misspelled version.

There's a correctly spelled version of the keyword earlier in the arguments list. If you're ok with changing the pymunk abstraction's signature again, I can remove the backwards compatibility and leave only the `moment_of_inertia` keyword without the extra T.

pushfoo:
@pvcraven Are you sure that you want to immediately drop support for old argument names such as `moment` and `moment_of_intertia`? #976 occurred because we changed the function signature in a minor release (2.6.2), had no backwards compatibility in place, and hadn't updated the relevant documentation. Other beginners are likely run into similar issues.

The changes currently in this PR provide a version of `PymunkPhysicsEngine.add_sprite` compatible with all prior 2.6.X releases. The old keywords are marked as deprecated with the intention of removing support for them in a later release such as 2.7. This way, beginners that did not pin a specific enough version will have an easier time.

If you are sure that you want to break compatibility now, I can make the changes you requested. @einarf also suggested reverting to `moment` as the keyword while offering limited compatibility with `moment_of_intertia` (extra T) for further 2.6.X releases. This matches the language used internally by pymunk/chipmunk despite being harder for beginners to understand.

einarf:
I was worried it will break pyweek entries. They have delivered their frozen code and people have a couple of weeks to review and vote. If it breaks the game because someone did not pin to an exact version.. that is a shame.

It's exactly that time of the year things like this is risky. It might also be ok. I haven't had time to check.

pvcraven:
At this point pyweek is over, right? I'm inclined to take this PR as-is.

einarf:
> At this point pyweek is over, right? I'm inclined to take this PR as-is.

Pyweek judging ends in a couple of weeks. Every participant has to try every game and score them. 

They made the game with 2.5.7 or 2.6.0/1

What people usually do is taking a week+ break after the code is submitted and then try a few games every day up towards the juding deadline. If they did not pin the requirements version those arcade games are broken forever.

einarf:
The problem here is:

The parameter used to be called `moment` in 2.5.7 up to 2.6.1. Now it has changed so any game using keyword arguments will forever be broken (code is frozen). Even our own examples broke. 

I'm fine with using either new or old name, but changing it might not be the best for our users.

PR #1006, opened 2021-10-13
Quick fix for omitted constructor docstrings on objects

### LINKED ISSUE #1005: Constructor documentation needs to move from __init__ to class docstring
### Actual behavior:

**tl;dr The current Sphinx configuration is omitting all constructor docstrings, including their param annotations.**

An important core class that beginners _**have**_ to use is affected: `arcade.SpriteList`. It is important that we include constructor argument explanations rather than just the attribute lists because constructors sometimes expose arguments that attributes do not.

`SpriteList`'s constructor docstring contains thorough explanations for each argument:
https://github.com/pythonarcade/arcade/blob/2ad27893c66efec62e71a3316b177b0fd25f0dfe/arcade/sprite_list/sprite_list.py#L48-L67

However, the current documentation build does not explain any of it:
image

Some other classes affected by this include arcade.TileMap and arcade.Texture. I'm sure how many others there are.

### Expected behavior:

Constructor documentation is output:
image

build arcade normally with `make docs`

###  ~~One way to fix it~~

~~Make sphinx use search both constructor docstrings. This seems like a good idea because:~~
~~1. Classes seem to document their options in only one place, not both~~
~~2. Sphinx has a simple config option to enable this [autoclass_content](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content)~~
~~3. Making the config change will reveal deeper information to curious beginners immediately~~

~~Adding `autoclass_content = 'both'` in `doc\conf.py` appears to generate doc properly for the classes that currently don't print docstrings. We may need to do some cleanup work around this.~~

pushfoo:
Closes #1005 in the quickest way possible, and fixing SpriteList, TiledMap, and Texture's omitted constructor annotations in the doc. There might be side effects if classes have single-line constructor docstrings or document something in both places. 

A quick glance at a local build with `python -m http.server` doesn't show too much out of the ordinary, but I need fresher eyes to take a look at this.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -27,6 +27,9 @@ runpy.run_path('../util/update_quick_index.py', run_name='__main__')
 
 autodoc_inherit_docstrings = False
 
+# quick fix for #1005 to enable constructor docstrings to be used
+autoclass_content = 'both'
+
 sys.path.insert(0, os.path.abspath('..'))
 sys.path.insert(0, os.path.abspath('../arcade'))
 


pushfoo:
Closing after clarification on project standards from @pvcraven

Issue #1030, opened 2021-11-06
All docstrings currently fail to render during doc build


pushfoo:
### What documentation needs to change? 
All doc for the API fails to generate. @mlr07 and @nfearnley first reported this problem on discord.

### Where is it located?
Anything under the API section for 2.6.5 doc on the arcade site ([sound as an example](https://api.arcade.academy/en/2.6.5/api/sound.html)). Local doc builds for tip of development also show the same issue. [2.6.4 works fine](https://api.arcade.academy/en/2.6.4/api/sound.html).

### What is wrong with it? How can it be improved?
Restore generation for doc.

### Logs during local build
Commit c77b619b seems to work fine. I suspect this is due to changes made in 25fff4e65 as a way to fix fonts because of errors during documentation build:

```
WARNING: autodoc: failed to import function 'rand_vec_magnitude' from module 'arcade'; the following exception was raised:
Traceback (most recent call last):
  File "c:\users\user\pycharmprojects\arcade\lib\site-packages\sphinx\ext\autodoc\importer.py", line 70, in import_module
    return importlib.import_module(modname)
  File "C:\Users\user\AppData\Local\Programs\Python\Python37\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "C:\Users\user\Projects\arcade\arcade\__init__.py", line 63, in <module>
    pyglet.font.base.Font.texture_width = 4096
AttributeError: module 'pyglet.font' has no attribute 'base'

WARNING: autodoc: failed to import function 'rand_vec_spread_deg' from module 'arcade'; the following exception was raised:
Traceback (most recent call last):
  File "c:\users\user\pycharmprojects\arcade\lib\site-packages\sphinx\ext\autodoc\importer.py", line 70, in import_module
    return importlib.import_module(modname)
  File "C:\Users\user\AppData\Local\Programs\Python\Python37\lib\importlib\__init__.py", line 127, in import_module
[...]

PR #1031, opened 2021-11-06
A less hacky fix for font atlas fix breaking doc

pushfoo:
Addresses #1030 in a way that doesn't rely on weird environment conditionals.

We might be able to remove these kludges when the Font class gets exposed more cleanly within pyglet.


einarf:
Let's do it the standard way to avoid additional issues: 292c68e4032eafdf586b151fafc2b214c0ce7f94

PR #1037, opened 2021-11-16
UI: UIFlatButton use new style parameters

eruvanos:
@pvcraven 
This breaks the compatibility with the previous UIFlatButton style names.
Should we have a transition phase or just keep this until we clean it up?

Not sure how worth it is to use the time for a smooth transition, but keeping the articles feedback in mind, I would like to discuss before :)

### DIFF
--- a/arcade/gui/widgets.py
+++ b/arcade/gui/widgets.py
@@ -1,4 +1,5 @@
 from abc import abstractmethod, ABC
+from collections import ChainMap
 from random import randint
 from typing import NamedTuple, Iterable, Optional, List, Union, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


pvcraven:
Argh. I could go either way. I think I'd err on pushing it to 2.7. At least, that would be the 'proper' way to do it versioning wise. I'm not sure what else we'd put in 2.7 though. So I don't have a good time-table for that.

Wish I had an idea how much it was used now. I'm guessing it is used in at least a few projects by this  time.

pvcraven:
These should be done in a new 2.7 branch einarf is making.

pvcraven:
Can this get pulled into 2.7 now?

eruvanos:
commit is merged into 2.7 branch

PR #1053, opened 2021-12-24
Correct line references in Pymunk Platformer tutorial

GuilleHoardings:


GuilleHoardings:
There may be some other incorrect lines, but they are in big chunks of text and I'm not sure if they are wrong

pvcraven:
Thanks!

PR #1061, opened 2022-01-07
made sprites cloneable by providing an origin to the sprite constructor

CommentatorForAll:
Currently, one has to do quite a lot of things to clone a sprite manually, especially if one sublcasses it, so after some discussion on the discord server, it was decided that a clone functionality would be useful. 

To clone a sprite, one passes the original one into the constructor. This allows better accessibility with subclasses than a .clone() function returning a new sprite would.

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -104,6 +104,8 @@ class Sprite:
 
     :param float hit_box_detail: Float, defaults to 4.5. Used with 'Detailed' to hit box
 
+    :param Sprite clone: Clones the sprites information over to this one.
+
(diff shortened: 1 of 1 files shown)


einarf:
I think we need to find a different way to do this. Adding even more to the sprite initializer is not a good idea. Maybe some internal method for copying properties and a clone() that needs to be implemented in the subclass.

pvcraven:
Seems like a separate clone function might be a bit cleaner.

pvcraven:
Going to close this out, don't think this is the way forward on this.

PR #1082, opened 2022-02-02
accept_keyboard_events enhanced and Fixed mypy error on Sections

alejcas:
This is just to fix a possible error detected by Mypy

### DIFF
--- a/arcade/sections.py
+++ b/arcade/sections.py
@@ -46,9 +46,9 @@ class Section:
         return self._view
 
     @property
-    def section_manager(self) -> "SectionManager":
+    def section_manager(self) -> Optional["SectionManager"]:
         """ Returns the section manager """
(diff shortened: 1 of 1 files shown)


alejcas:
Hi, @pvcraven, this is difficult to solve due to circular imports for type annotation.

Mypy is complaining that the attribute `_view` from `Section` can't have a `section_manager` because `Section._view` is not annotated as type `View`.
And I can't annotate `Section._view` as type `View` because that will cause a circular import error.

¿How should I solve this?

¿Did you consider adding [`from __future__ import annotations`](https://www.python.org/dev/peps/pep-0563/) so you can avoid circular imports between files?

einarf:
Keep the current annotation and import view only when type checking is running
```py
if TYPE_CHECKING:
    from arcade import View
```

PR #1105, opened 2022-03-13
More Easing functions in easing.py

DragonMoffon:
Added a few more easing functions that I think many people would find helpful. Plus added a reset method to the EasingData class to make it obvious how to reset it.


pvcraven:
Thanks! https://api.arcade.academy/en/development/examples/easing_example_1.html#easing-example-1

PR #1110, opened 2022-03-18
Headless Compatibility

gaorkl:
Implemented simple Headless Compatibility:
- test if display is present in the __init__ of arcade.
- if no display, set pyglet's headless option﻿

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -53,6 +53,10 @@ else:
 # noinspection PyPep8
 import pyglet
 
+# Not sure if this is generic enough
+if not os.environ.get('DISPLAY'):
+    pyglet.options["headless"] = True
+
 # Disable shadow windows until issues with intel GPUs
 # on Windows and elsewhere are better understood.
 # Originally, this only disabled them for macs where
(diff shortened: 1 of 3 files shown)


einarf:
I was going to merge this today, but we need a 2.6.12 release before that happens. Very soon.

gaorkl:
No worries! I sent a message on the issue.
I will try something this morning and let you know.

gaorkl:
So, the way the tests are structured, I didn't manage to add an option to run headless tests in pytest.
I think that this is due to the fact that arcade is loaded in the conftest, so I can't start headless mode there.

As a workaround, I added code in the ```__init__``` of the tests directory.

einarf:
I will merge this shortly and do some more tweaks.

I think using `DISPLAY` is a bit risky, but I will remove that. it's probably better to just have the user set the headless pyglet option. We can add an `ARCADE_HEADLESS` env var as an alternative option.

Other things to weak:
* Add `Window.headless` property
* Patch `push_handlers` to do nothing in headless mode
* Emulate a simple event loop in `run()`
* Possibly patch some other window methods.
* Test some examples and dump framebuffer (possibly add a capture option)
* Add quick guide in the advanced section

Hopefully not going to bump into anything too weird.

Issue #1114, opened 2022-03-20
Enhancement: Adding a clear method in UIManager


Ibrahim2750mi:
### What should be added?
In [arcade.gui.ui_manager](https://api.arcade.academy/en/latest/_modules/arcade/gui/ui_manager.html#UIManager) there should be a method to clear the container.

### What would it help with?
Consider a situation where you want to have buttons which will open to more buttons in the same position, instead of creating a new arcade.gui.ui_manager or manually removing with [walk_widgets](https://api.arcade.academy/en/latest/api/gui.html#arcade.gui.UIManager.walk_widgets) and the remove method.

eruvanos:
Will be released with next version

eruvanos:
Released

PR #1116, opened 2022-03-20
Adding a clear method in UIManager

Ibrahim2750mi:
Adds a clear method in the class which uses `walk_widgets` and `remove`

#1114


eruvanos:
Does this also work for nested widget trees?

Ibrahim2750mi:
Does a nested widget tree created when stacking two UIBoxLayout? If so, then yes

PR #1123, opened 2022-03-22
Section type fix. Section dimensions should be integers

alejcas:
Section dimensions should use integers as Window do.


einarf:
much better

Issue #1117, opened 2022-03-21
Update install overview with requirement for Python 3.7+


pushfoo:
### What documentation needs to change? 
> Arcade requires Python 3.6 or newer. It does not run on Python 2.x.

### Where is it located?
[doc/install/index.rst](https://github.com/pythonarcade/arcade/blob/development/doc/install/index.rst) or https://api.arcade.academy/en/latest/install/index.html

### What is wrong with it? How can it be improved?

Arcade requires Python 3.7 or later as of commit b9861102 / version 2.6.12. This should be clearly stated on the install page. Python version support has tripped up users in the past ([example in issue comments](https://github.com/pythonarcade/arcade/issues/934#issuecomment-910423855))

Users of older Python versions can be directed to try specific older versions of Arcade.

PR #1129, opened 2022-03-24
Replace mentions of Python 3.6 on Linux install page

pushfoo:
Quick tweaks in line with #1117. It didn't seem worthwhile to make an issue per discussion with einarf yesterday. The doc builds locally.

I'll file further issues for bigger updates to each of the install instruction pages.


pvcraven:
Thanks!

PR #1140, opened 2022-03-29
Remove obsolete [docs] option from source install instructions

pushfoo:
Running the install from source instructions with [docs] produces an error:
```
WARNING: arcade 2.6.13 does not provide the extra 'docs'
```
This option has been obsolete since at least October 2021 per [a message from Clepto in discord](https://discord.com/channels/458662222697070613/458662222697070615/898015011736277002):
```
Docs aren't split anymore FYI, it's just pip install -e .[dev]
```

### DIFF
--- a/doc/install/source.rst
+++ b/doc/install/source.rst
@@ -24,7 +24,5 @@ To install additional documentation and development requirements:
 
 .. code-block:: bash
 
-    pip install -e .[dev,docs]
-    # .. or separately
     pip install -e .[dev]
-    pip install -e .[docs]
+


pvcraven:
A doc requirements file does exist. It is `doc/requirements.txt`

Probably need to spell out this is for installing additional libraries for building documentation, not for just having it.

einarf:
Remove `doc/requirements.txt` and make sure they are in the dev requirments i `setup.py`. I see some of them there, but they are not pinned to a version (bad).

pushfoo:
This doc PR might be orthogonal to the doc build changes einarf proposed.
[`doc/requirements.txt` was created last summer in a commit that has to do with readthedocs](https://github.com/pythonarcade/arcade/commit/64ce5cc7f80271aff0b0e538f3ff138752b600f9). The getting started instructions for readthedocs don't mention it, but there is a setuptools install option buried deeper in rtd's documentation ([example](https://docs.readthedocs.io/en/stable/config-file/v2.html#python)). Our [current rtd config file doesn't use it](https://github.com/pythonarcade/arcade/blob/968f09b818f4f31c08580f31d334795ce27e9dff/.readthedocs.yaml), but maybe it should. That set of changes should probably be a separate ticket and PR.

einarf:
Ahh right. I think we just need to move back to dev and docs groups in setup.py

* `dev` could contain everything (including docs)
* `docs` can contain just what is needed for readthedocs

https://docs.readthedocs.io/en/stable/config-file/v2.html#packages

That will make this PR still valid and we get rid of the requirements file.

einarf:
This resolves the mess. It also means we no longer will have stale pyglet version in doc requirements. 2fad1b7b2b53f618ea641909fce747cd6ae0385d

PR #1148, opened 2022-03-31
Added `on_mouse_enter` and `on_mouse_leave` to `View`

alejcas:
I don't know why `View` doesn't implement `on_mouse_enter` and `on_mouse_leave`.

Here I just add that for the `View`

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -737,6 +737,29 @@ class Window(pyglet.window.Window):
         """ Dispatch events """
         super().dispatch_events()
 
+    def on_mouse_enter(self, x: int, y: int):
+        """
+        Called when the mouse was moved into the window.
+        This event will not be triggered if the mouse is currently being
+        dragged.
+
+        :param int x:
+        :param int y:
+        """
+        pass
+
+    def on_mouse_leave(self, x: int, y: int):
+        """
+        Called when the mouse was moved outside of the window.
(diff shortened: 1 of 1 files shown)


einarf:
Should we also add a wrapper for this in the `Window` so it gets exposed in arcade docs?

alejcas:
> Should we also add a wrapper for this in the `Window` so it gets exposed in arcade docs?

Don't know what you mean... sorry

einarf:
I mean just adding the same function to `Window`. This means arcade docs will inform the user about the existance of this event. It seems weird that only `View` would have it.

alejcas:
ahh ok, the Window already have these methods. It's weird to me that View didn't.
Maybe it's something intentional?

I tried and if I define those in a View subclass they work as expected.

einarf:
Just add the same methods to `Window` so these callbacks are exposed in arcade docs. Most arcade users don't know that pyglet exists and it makes sense to add them in both places.

alejcas:
Ok!

einarf:
Awsome. Then we can tell the world that Alejandro added this great new feature in patch notes 🎉

alejcas:
> Awsome. Then we can tell the world that Alejandro added this great new feature in patch notes 🎉

yeah that took me a lot amount of work! ;)

PR #1151, opened 2022-03-31
Fixed annotations and docstrings now that hit_box_algorithm can be None

alejcas:
Fixed annotations and docstrings now that hit_box_algorithm can be None. Before that needed to be a string but not anymore.


einarf:
Good catch

Issue #1147, opened 2022-03-31
Arcade Mouse Event params are annotated as floats when they are Integers


alejcas:
The mouse event params in arcade (Window and View events) are annotated as floats when they are integers.

According to pyglet docs, they are indeed integers:

From pyglet:

```python
def on_mouse_motion(self, x, y, dx, dy):
  """The mouse was moved with no buttons held down.
  :Parameters:
      `x` : int
          Distance in pixels from the left edge of the window.
      `y` : int
          Distance in pixels from the bottom edge of the window.
      `dx` : int
          Relative X position from the previous mouse position.
      `dy` : int
          Relative Y position from the previous mouse position.
  :event:
  """
```

**Before submitting a PR... is this somehow intended?**

einarf:
I think we initially made them floats because they might be floats  in the future. Mouse coordinates can actually be floats when the pixel ratio of the window is >1.0. **It doesn't seem like this is going to change any time soon, so it's probably fine to change the annotations.**

Just to add some more explanation to this:

At an OS level you can query mouse position in framebuffer coordinates instead of window coordinates. For example: On MacOS with retina a 100 x 100 window has a framebuffer of 200 x 200 pixels meaning the mouse can move at 0.5 pixel increments. Windows and Linux also have similar scaling systems like this, but you can choose to ignore it.

The mouse position in the GLFW library do report positions in `double` values for  this reason.

```C
void function_name(GLFWwindow* window, double xoffset, double yoffset)
```

I think it's also possible for a high dpi mouse to report even smaller movements below a framebuffer pixel even though the actual mouse cursor is not visually moving. In a perfect world mose coordinates  should be floats, but they are not in pyglet for now. It's more than good enough for our use.

alejcas:
So, for the moment the types are integers so better have the annotations also as integers. If that changes in the future we can then change the annotations as well.

I'll work on PR soon with this

Thanks for the extended response, very informative!

einarf:
I agree. It was just that we didn't really know if 2.0 would have floats or not, but that has been sorted now 😄

PR #1153, opened 2022-03-31
Fix mouse param annotations and docstrings to be integers

alejcas:
See #1147
Also adding some missing docstrings


einarf:
Thank you!

Issue #1101, opened 2022-03-07
GUI enhancements planned for 3.0


eruvanos:
To enhance the GUI code, a few things have to be cleaned which might break compatibility.
The goal is to collect all these changes within a bigger update like 3.0.

## Summary of breaking changes

- Removed
  - UIWrapper (it is a UILayout)
  - UIBorder (functionality covered in UIWidget)
  - UIPadding (functionality covered in UIWidget)
  - UITexturePane (functionality covered in UIWidget)
- Renamed
  - Rect (_Rect)
  - Property (_Property)

## Planned

- [x] UIWidget supports
  - [x]  padding
  - [x] border
  - [x] background color
  - [x] states: 
    - [x] visibility, 
    - [x] disabled, 
    - [x] hover, 
    - [x] pressed, 
    - [x] maybe use a dict mapping state to style class
- [x] UIWidgets use observable `Property`
- [x] Rect is used as the space representation on screen
- [x] UILayouts shrink and grow children
  - [x] UIAnchorLayout use min_size_hint of UIBoxLayout
  - [x] UIBoxLayout use size_hint to grow children 
  - [x] UIBoxLayout respects size_hint_max while growing
  - [x] UIAnchorLayout use size_hint to grow children
  - [x] UIAnchorLayout respects size_hint_max while growing
- [x] UILayoutManager use size hints
   - [x] UIManager respects size_hint and size_hint_min
- [x] UIAnchorLayout should behave like a layout (rect set from outside, place children within bounds)
- [x] Fix stats_top_left example (fix UILabel not showing text)
- [x] Refactor UIMessageBox to fit into general event system
- [x] Support NinePatch 
  - [x] Basic support and rendering
  - [x] Use as background in UIWidget
  - [x] Use as texture in UITexturedButton
- [x] Methods do only contain supported parameters
- [x] Missing documentation
  - [x] UISlider
- [x] Replace `draw_text` with a label
  - [x] UIFlatButton
  - [x] UITexturedButton
- [x] Fix curser in inputtext
- [x] Improve UIInputText, expose more pyglet parameters (now exposes `text.py` as `UIInputText.layout`)
- [x] Improve UILabel, expose more pyglet parameters (now exposes `pyglet.text.Label` as `UILabel.layout`)
- [x] New widgets
  - [x] UIGridLayout - Finally arrange widgets within a grid
    - [x] resizes child widgets
[...]

PR #1160, opened 2022-04-03
added UIGridLayout

Ibrahim2750mi:
#1101  
- [x] UIGridLayout - Finally arrange widgets within a grid


eruvanos:
I will also replace the UIAnchorWidget with a proper UIAnchorLayout, with the same mechanic (`layout.add(child, anchor_x="left"`)

PR #1165, opened 2022-04-04
fix link to PhysicsEngineSimple documentation

vpoulailleau:


vpoulailleau:
BTW, I didn't take the time to generate the documentation, I think the link is now correct but I didn't test it.

einarf:
Ideally we should use anchors here, but this gets the job done at least!

PR #1169, opened 2022-04-05
Fix broken links on pygame comparison page

pushfoo:
Two absolute links were broken due to past doc restructuring. I replaced them with automatic sphinx references.

Built & tested locally, works fine.


einarf:
👍

Issue #1111, opened 2022-03-18
Promote using `Text` objects in examples


einarf:
The future of text is `Text` objects. `arcade.draw_text(..)` will in most cases be 10-100 times slower no matter how many fancy tricks we come up with.

We should start to promote using `Text` objects over `draw_text()` in examples to slowly start a process were we can possibly remove `arcade.draw_text` in the future.

einarf:
Examples exposed in docs will need their line numbers updated if code is highliged there.

einarf:
Added performance warning 3941029ace683bff3913496fc60a88d68368780f

pushfoo:
The View examples currently have some draw_text in them on `development` branch. If anyone reads this, feel free to fix that. I'll clean it up later if nobody beats me to it.

einarf:
Done

PR #1176, opened 2022-04-06
Two new examples. One on sprite rotation, and an example for a correct tank barrel.

DragonMoffon:


arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+SCREEN_HEIGHT = 600
+
+SCREEN_TITLE = "Rotating Sprite Example"
+
+
+class RotatingSprite(arcade.Sprite):
+
+    def rotate_around_point(self, point, degrees, change_angle=True):
+        """
+        rotates the sprite around a defined point by the set amount of degrees
```
pushfoo: This could be a good place to add a more in-depth doc string. `:param varname:` annotations can be used to explain what each argument does at the top, and simplify the other comments.

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
+"""
```
pushfoo: I would move this file to `arcade/examples/sprite_rotate_around_point.py` for two reasons:
1. It's clearer and more consistent with the heading in the docstring
2. It's consistent with the naming scheme we have for sprite examples

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+Rotating Sprites Around Points
+
+A simple program to show how someone could implement sprites rotating around points.
+
+Artwork from https://kenney.nl
```
pushfoo: It would be a good idea to add this at the bottom of the file-level docstring to help beginners skimming the doc:
```
If Python and Arcade are installed, this example can be run from the command line with:
python -m arcade.examples.sprite_rotate_around_point
```
This assumes the filename was changed in accordance with the first comment on this file.

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+import arcade
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 600
+
+SCREEN_TITLE = "Rotating Sprite Example"
+
+
+class RotatingSprite(arcade.Sprite):
```
pushfoo: A class-level docstring would be a good idea here. It could explain the following:
1. This is an example class with a generic rotate around point method
2. You can omit features like the change_angle argument if your sprites will always change the angle

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+            self.angle += degrees
+
+        # use arcade.rotate_point to rotate the sprites center x and y by the point.
+        self.center_x, self.center_y = arcade.rotate_point(self.center_x, self.center_y, point[0], point[1], degrees)
+
+
+class ExampleWindow(arcade.Window):
+
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
```
pushfoo: **Optional:** You could add 2 `arcade.Text` instances as captions, one above each side of this example. The laser bar's caption could be `change_angle = True` and the platform's could be `change_angle = False`.

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # This first example is similar to the spinning fireballs from mario.
+        self.example_sprite_1 = RotatingSprite(":resources:images/space_shooter/laserBlue01.png",
+                                               center_x=SCREEN_WIDTH // 4 + 26, center_y=SCREEN_HEIGHT // 2)
+
+        # This second example shows how the method can be used to make rotating platforms.
+        self.example_sprite_2 = RotatingSprite(":resources:images/tiles/grassHalf.png", scale=0.25,
+                                               center_x=3 * SCREEN_WIDTH // 4 + 50, center_y=SCREEN_HEIGHT // 2)
+
+        # This third example shows how the method can be used to rotate the barrel of a tank correctly.
```
pushfoo: What tank barrel? I'm confused. This comment should be changed to something like this:
```py
# The two sprites above will rotate around the centers of the boxes below
```

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.base_sprite_1 = arcade.Sprite(":resources:images/tiles/boxCrate.png", scale=0.25,
+                                           center_x=SCREEN_WIDTH // 4, center_y=SCREEN_HEIGHT // 2)
+        self.base_sprite_2 = arcade.Sprite(":resources:images/tiles/boxCrate.png", scale=0.25,
+                                           center_x=3 * SCREEN_WIDTH // 4, center_y=SCREEN_HEIGHT // 2)
+
+        self.sprites = arcade.SpriteList()
+        self.sprites.extend([self.base_sprite_1, self.example_sprite_1, self.base_sprite_2, self.example_sprite_2])
+
+    def on_update(self, delta_time: float):
```
pushfoo: :100::100::100:! The comments in this method are straight forward and extremely legible! Excellent!

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+
+Artwork from https://kenney.nl
+
+CONTROLS:
+    WASD to move the tank
+
+    MOUSE to aim the barrel
+
+    P to switch between incorrect and correct barrel code.
+"""
```
pushfoo: As with the previous example, people skimming the doc could find this addition helpful at the end of the docstring:
```
If Python and Arcade are installed, this example can be run from the command line with:
python -m arcade.examples.sprite_rotation_tank
```

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+        self.center_x, self.center_y = arcade.rotate_point(self.center_x, self.center_y, point[0], point[1], degrees)
+
+
+class ExampleWindow(arcade.Window):
+
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
+        # Generate a grassy background
+        self.background = arcade.SpriteList()
+        # since 64 is not a factor of SCREEN_WIDTH or SCREEN_HEIGHT add an extra to cover the gap
```
pushfoo: a blank line above this comment might make this a bit more readable

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+TANK_SPEED = 64  # How many pixels per second the tank travels
+TANK_TURNING_SPEED = 60  # how many degrees per second the tank spins by.
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 600
+
+SCREEN_TITLE = "Rotating Tank Example"
+
+
+class RotatingSprite(arcade.Sprite):
```
pushfoo: You can copy and paste the changes from the other file's RotatingSprite class into this one.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            arcade.draw_text("INCORRECT", SCREEN_WIDTH // 2, SCREEN_HEIGHT - 25, arcade.color.BLACK,
+                             anchor_x='center')
+
+    def on_update(self, delta_time: float):
+        self.move_tank(delta_time)
+
+    def move_tank(self, delta_time):
+        """
+        Here we do all the calculations about how to move the tank since we use 'tank style' motion.
```
pushfoo: I would change this doc string to explicitly mention that all parts of the tank are moved by this method.
```
"""
Perform all calculations about how to move the tank, including both the body and the barrel
"""
```
This way, readers will know where to look for the two kinds of rotation logic demonstrated

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        # change the tanks angle we don't need to worry about the barrel since it points towards the mouse
```
pushfoo: This comment is a bit misleading. I initially read it as suggesting that the barrel isn't moved in this method. I think a better way of phrasing it could be something like this:
```
# update the angle of the tank's body alone; the barrel will be updated after the body is moved
```

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+        self.tank.angle += TANK_SPEED * self.tank_turning * delta_time
+
+        # find how much the tank's x and y should change to move forward or back.
+        x_dir = math.cos(self.tank.radians - math.pi / 2) * self.tank_direction * TANK_SPEED * delta_time
+        y_dir = math.sin(self.tank.radians - math.pi / 2) * self.tank_direction * TANK_SPEED * delta_time
+
+        # we then move the tank and the barrel since they are connected together.
```
pushfoo: :+1:

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
+"""
```
pushfoo: This file name might be ok.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+Sprite Rotation With A Tank
+
+This is an extension of the sprite rotation example.
```
pushfoo: Which rotation example? I think you can delete this, see the next two comments.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+
+A simple program to provide an example tank.
```
pushfoo: Redundant, best deleted.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+The problem often faced when making a tank game is that the barrel is its own separate sprite.
```
pushfoo: Not just its own separate sprite, but an asymmetrical sprite. That's what makes rotating around a point necessary.
I might phrase it like this:
```
Vehicles or tower defense turrets can have parts that can rotate toward targets.
```

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+If the barrel sprite is just rotated by its angle it will not be positioned correctly.
```
pushfoo: This line could be replaced with something more like this:
```
These parts are usually represented with separate sprites drawn relative to attachment points on the main body. 

Because these sprites are usually asymmetrical, we have to rotate them around their attachment points on the main body or they will look wrong!

This example allows the player to switch between two ways of rotating a tank's turret and barrel:
1. correctly, around a point where the turret's center is on the body
2. incorrectly, around the center of the sprite
```

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+        self.tank_sprite_list.extend([self.tank, self.barrel])
+
+        self.correct = True
+
+    def on_draw(self):
+        self.clear()
+        self.background.draw()
+        self.tank_sprite_list.draw()
+
+        if self.correct:
```
pushfoo: It would be a good idea to replace this if/else pair with a single call of the the draw method of an `arcade.Text` instance (ie `self.mode_caption`). There are a number of reasons for this:
1. `arcade.Text` is faster (see #1111) and we're moving towards it as a default
2. A single `self.mode_caption.draw()` call is shorter
3. We can update the caption whenever `self.correct` is set. Updating the caption should probably be a method of its own, or included in a property setter for `self.correct`

You can use the exact same arguments for `arcade.Text` as you would with `draw_text`. See the comment further down in `on_key_press` for more info on how to implement this.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+    def on_key_press(self, symbol: int, modifiers: int):
+        if symbol == arcade.key.W:
+            self.tank_direction += 1
+        elif symbol == arcade.key.S:
+            self.tank_direction -= 1
+        elif symbol == arcade.key.A:
+            self.tank_turning += 1
+        elif symbol == arcade.key.D:
+            self.tank_turning -= 1
+        elif symbol == arcade.key.P:
```
pushfoo: It would be a good idea to update an `arcade.Text` instance somewhere in this block. The label's contents can be set using an f-string that would also display controls information:
```py
self.mode_caption = f"Turret display mode :"\
      f"{'Correct' if self.correct else 'Incorrect'} - Press P to swap"
```

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+This is an extension of the sprite rotation example.
+
+A simple program to provide an example tank.
+The problem often faced when making a tank game is that the barrel is its own separate sprite.
+If the barrel sprite is just rotated by its angle it will not be positioned correctly.
+
+Artwork from https://kenney.nl
+
+CONTROLS:
+    WASD to move the tank
```
pushfoo: It would be very helpful to display controls on screen. Use an `arcade.Text` instance for this. Remember, people will be launching these examples for very brief periods of time and won't have time to learn the controls.

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+            self.tank_direction -= 1
+        elif symbol == arcade.key.S:
+            self.tank_direction += 1
+        elif symbol == arcade.key.A:
+            self.tank_turning -= 1
+        elif symbol == arcade.key.D:
+            self.tank_turning += 1
+
+    def on_mouse_motion(self, x: int, y: int, dx: int, dy: int):
+        self.mouse_pos = [x, y]
```
pushfoo: I don't think you don't need list brackets here. Multiple assignment to a single variable is automatically treated as assigning a tuple. ie, this should work:
```py
self.mouse_pos = x, y
```

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+        rotates the sprite around a defined point by the set amount of degrees
+        """
+
+        if change_angle:
+            # This is so the direction the sprite faces changes when rotating.
+            # It isn't necessary to have this. For example, you would want a rotating platform to always face upwards.
+            self.angle += degrees
+
+        # use arcade.rotate_point to rotate the sprites center x and y by the point.
+        self.center_x, self.center_y = arcade.rotate_point(self.center_x, self.center_y, point[0], point[1], degrees)
```
pushfoo: To my understanding, there's no need to set these separately unless benchmarks proved this to be faster. Setting `position` should be enough. Setting both `center_x` and `center_y` will repeat the series of spatial hash resets that `position` does only once:
https://github.com/pythonarcade/arcade/blob/c1784151b6dace5104e4282362f43478a2d7e161/arcade/sprite.py#L667-L677
https://github.com/pythonarcade/arcade/blob/c1784151b6dace5104e4282362f43478a2d7e161/arcade/sprite.py#L684-L694

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ExampleWindow(arcade.Window):
+
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
+
+        # This first example is similar to the spinning fireballs from mario.
```
pushfoo: Although it's difficult to imagine, not everyone has played Mario. I'd turn this comment into 2 lines phrased something like the following:
```
# The first example is similar to the spinning lines of fireballs from the Super Mario series
# See https://www.mariowiki.com/Fire_Bar for more information
```

arcade/examples/sprite_rotation.py
```diff
@@ -0,0 +1,76 @@
...
+        self.example_sprite_1 = RotatingSprite(":resources:images/space_shooter/laserBlue01.png",
```
pushfoo: **Optional:** You could give the moving sprites more descriptive variable names. Perhaps something like `self.left_laser_bar` and `self.right_circling_platform`?

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # we then move the tank and the barrel since they are connected together.
+        self.tank.center_x += x_dir
+        self.tank.center_y += y_dir
+
+        self.barrel.center_x += x_dir
+        self.barrel.center_y += y_dir
+
+        if self.correct:
+            # This creates the correct motion for the barrel.
```
pushfoo: Redundant, best deleted

arcade/examples/sprite_rotation_tank.py
```diff
@@ -0,0 +1,168 @@
...
+            # We want to rotate the barrel around the center of the tank. Not the center of the barrel sprite
```
pushfoo: You could rephrase it like this:
```py
# Rotate the barrel sprite around the center of the tank, not the center of the barrel sprite
```

pushfoo (changes requested):
The mechanics you chose to demonstrate this concept are excellent!

However, I think these files need some changes to improve legibility and compliance with standards.

You should also considering limiting line length. You don't need to stick to a hard 80, but it seems like you might be going past 120 in some places.

arcade/examples/sprite_rotation.py
DragonMoffon: Yeah, setting the position is faster. I just forgot.

arcade/examples/sprite_rotation.py
DragonMoffon: whoops, I forgot to remove that. I was going to add the tank into this example origianlly.

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        rotates the sprite around a defined point by the set amount of degrees
+
+        :param point: The point that the sprite will rotate about
+        :param degrees: How many degrees to rotate the sprite
+        :param change_angle: whether the sprite's angle should also be adjusted.
+        """
+
+        if change_angle:
+            # This is so the direction the sprite faces changes when rotating.
```
pushfoo, line 35: These comment lines are still hard to understand. I think you can remove them because you clarified this in the class docstring already.

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+            # It isn't necessary to have this. For example, you may only use this for rotating platforms that will
+            # never change angle.
+            self.angle += degrees
+
+        # use arcade.rotate_point to rotate the sprites center x and y by the point.
```
pushfoo, line 38: this could be shorter:
```
# rotate the sprite's position around the passed point
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -1,20 +1,20 @@
 """
-Sprite Rotation With A Tank
+Sprite Rotation With A Tank.
 
-This is an extension of the sprite rotation example.
+The problem often fis that vehicles or tower defense turrets can have parts that can rotate toward targets.
```
pushfoo: Typo in this line. "fis". You can probably delete everything before Vehicles.

pushfoo:
Review is still WIP. I'll try to finish it at some point in the next 24 hrs

pvcraven:
Hi. If you want to make this into a full example that people can see, please also take a look at doing the following:

1. Go to doc/examples and add a .rst file with the same name as the .py file in your example.
2. Same directory, add a .png for a screenshot of your example.
3. Same directory, edit index.rst so that it includes a link to your example.

einarf:
It could also possibly be a good idea to reduce the horizontal space used by the code if this is supposed to be displayed as an example in docs?

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ExampleWindow(arcade.Window):
+
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
+
+        # The first example is similar to the spinning lines of fireballs from the Super Mario series
+        # See https://www.mariowiki.com/Fire_Bar for more information
+        self.left_rotating_laser_sprite = RotatingSprite(":resources:images/space_shooter/laserBlue01.png",
```
pushfoo, line 49: To help the code fit into the website's embedded view, you could set the positions of sprites on a new line after instantiating them:
```py
self.left_rotating_laser_sprite = RotatingSprite(":resources:images/space_shooter/laserBlue01.png")
self.left_rotating_laser_sprite.position = QUARTER_WIDTH + 26, HALF_HEIGHT
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+
+class ExampleWindow(arcade.Window):
+
+    def __init__(self):
+        super().__init__(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
+
+        # Generate a grassy background
+        self.background = arcade.SpriteList()
+
+        # since 64 is not a factor of SCREEN_WIDTH or SCREEN_HEIGHT add an extra to cover the gap
```
pushfoo: although it looks nice, I think we can drop the background texture from the example because it doesn't help demonstrate the principle.  Using [arcade.set_background_color](https://api.arcade.academy/en/latest/api/window.html?highlight=background#id14) to set the background to a similar shade of green would be enough.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        # The tank and barrel sprite. These Can be all added to their own separate class in a more advanced program.
+        self.tank = arcade.Sprite(":resources:images/topdown_tanks/tankBody_dark_outline.png",
+                                  center_x=SCREEN_WIDTH // 2, center_y=SCREEN_HEIGHT // 2)
+        self.barrel = RotatingSprite(":resources:images/topdown_tanks/tankDark_barrel3_outline.png",
+                                     center_x=SCREEN_WIDTH // 2, center_y=SCREEN_HEIGHT // 2 - 13)
+
+        self.tank_direction = 0.0  # If the tank is moving forward or backwards.
+        self.tank_turning = 0.0  # If the tank is turning left or right.
+
+        self.mouse_pos = [0, 0]
```
pushfoo, line 81: The list format isn't needed. This should be enough since the mouse position is only overwritten instead of modified:
```py
self.mouse_pos = 0, 0
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        self.barrel.center_y += y_dir
+
+        if self.correct:
+            # Rotate the barrel sprite around the center of the tank, not the center of the barrel sprite
+
+            # we need to add 90 to the angle due to orientation of the barrel texture.
+            # we need to remove the barrels angle as we only want the change in angle.
+            angle_change = arcade.get_angle_degrees(self.tank.center_y, self.tank.center_x,
+                                                    self.mouse_pos[1], self.mouse_pos[0]) - self.barrel.angle + 90
+            self.barrel.rotate_around_point((self.tank.center_x, self.tank.center_y), angle_change)
```
pushfoo: the point can be gotten directly:
```py
self.barrel.rotate_around_point(self.tank.position, angle_change)
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+    def move_tank(self, delta_time):
+        """
+        Perform all calculations about how to move the tank, including both the body and the barrel
+        """
+
+        # update the angle of the tank's body alone. The barrel will be updated after the body is moved
+        self.tank.angle += TANK_SPEED * self.tank_turning * delta_time
+
+        # find how much the tank's x and y should change to move forward or back.
+        x_dir = math.cos(self.tank.radians - math.pi / 2) * self.tank_direction * TANK_SPEED * delta_time
```
pushfoo: you can make these lines shorter by pre-calculating the movement magnitude and storing it as a variable:
```py
movement_magnitude = self.tank_direction * TANK_SPEED * delta_time
x_dir = math.cos(self.tank.radians - math.pi / 2) * movement_magnitude
y_dir = math.sin(self.tank.radians - math.pi / 2) * movement_magnitude
```

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+                             self.platform_base_sprite, self.right_rotating_platform_sprite])
+
+        self.laser_text = arcade.Text("change_angle = True", SCREEN_WIDTH // 4, SCREEN_HEIGHT // 2 - 150,
+                                      anchor_x='center')
+        self.platform_text = arcade.Text("change_angle = False", 3 * SCREEN_WIDTH // 4, SCREEN_HEIGHT // 2 - 150,
+                                         anchor_x='center')
+
+    def on_update(self, delta_time: float):
+        # Rotate the laser sprite and change its angle
+        self.left_rotating_laser_sprite.rotate_around_point((self.laser_base_sprite.center_x,
```
pushfoo, line 73: You can make this method's body fit better into the web embed view by storing centers under convenient names
```py
left_center = self.laser_base_sprite.position
self.left_rotating_laser_sprite.rotate_around_point(left_center, 120 * delta_time)
```
Remember to also make this change for the platform on the right.

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+Artwork from https://kenney.nl
+
+If Python and Arcade are installed, this example can be run from the command line with:
+python -m arcade.examples.sprite_rotate_around_point
+"""
+import arcade
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 600
+
```
pushfoo, line 15: you could make this file's lines shorter and more readable by adding and using the following constants:
```py
HALF_WIDTH = SCREEN_WIDTH // 2
QUARTER_WIDTH = SCREEN_WIDTH // 4
HALF_HEIGHT = SCREEN_HEIGHT // 2
BARREL_WIDTH = 13 
```
Per discord discussion, double check the barrel width.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+
+If Python and Arcade are installed, this example can be run from the command line with:
+python -m arcade.examples.sprite_rotate_tank
+"""
+
+import arcade
+import math
+
+TANK_SPEED = 64  # How many pixels per second the tank travels
+TANK_TURNING_SPEED = 60  # how many degrees per second the tank spins by.
```
pushfoo, line 29: you could make this file's lines shorter and more readable by adding and using the following constants:
```py
HALF_WIDTH = SCREEN_WIDTH // 2
HALF_HEIGHT = SCREEN_HEIGHT // 2
HALF_PI = math.pi / 2
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @property
+    def correct(self):
+        return self._correct
+
+    @correct.setter
+    def correct(self, value):
+        if value:
+            self._correct = True
+            angle = math.radians(arcade.get_angle_degrees(self.tank.center_y, self.tank.center_x,
```
pushfoo, line 185: You can use `get_angle_radians` here instead as #1180 just got merged. It'll simplify the current radian to degree to radian conversion and shorten line length.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        elif symbol == arcade.key.S:
+            self.tank_direction -= 1
+        elif symbol == arcade.key.A:
+            self.tank_turning += 1
+        elif symbol == arcade.key.D:
+            self.tank_turning -= 1
+        elif symbol == arcade.key.P:
+            self.correct = bool(1 - self.correct)
+
+            self.correct_text.text = f"Turret Rotation is {'Correct' if self.correct else 'Incorrect'}," \
```
pushfoo: I think you can split this up to be even cleaner:
```py
            self.correct_text.text = \
                    f"Turret Rotation is {'Correct' if self.correct else 'Incorrect'}," \
                    f" Press P to Switch"
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+    def on_update(self, delta_time: float):
+        self.move_tank(delta_time)
+
+    def move_tank(self, delta_time):
+        """
+        Perform all calculations about how to move the tank, including both the body and the barrel
+        """
+
+        # update the angle of the tank's body alone. The barrel will be updated after the body is moved
+        self.tank.angle += TANK_SPEED * self.tank_turning * delta_time
```
pushfoo, line 114: Did you mean `TANK_TURNING_SPEED` here? If not, you should remove that constant.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        :param point: The point that the sprite will rotate about
+        :param degrees: How many degrees to rotate the sprite
+        """
+
+        # This is so the direction the sprite faces changes when rotating.
+        # It isn't necessary to have this. For example, you would want a rotating platform to always face upwards.
+        self.angle += degrees
+
+        # use arcade.rotate_point to rotate the sprites center x and y by the point.
+        self.center_x, self.center_y = arcade.rotate_point(self.center_x, self.center_y, point[0], point[1], degrees)
```
pushfoo: `self.position` is faster than both components.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        self.barrel.center_x += x_dir
+        self.barrel.center_y += y_dir
+
+        if self.correct:
+            # Rotate the barrel sprite around the center of the tank, not the center of the barrel sprite
+
+            # we need to add 90 to the angle due to orientation of the barrel texture.
+            # we need to remove the barrels angle as we only want the change in angle.
+            angle_change = arcade.get_angle_degrees(self.tank.center_y, self.tank.center_x,
+                                                    self.mouse_pos[1], self.mouse_pos[0]) - self.barrel.angle + 90
```
pushfoo: You could try something like this to reduce the horizontal space taken up by the code:
```py
angle_change = arcade.get_angle_degrees(
    self.tank.center_y, self.tank.center_x,
    self.mouse_pos[1], self.mouse_pos[0])
    
# subtract the current angle because we want only the change in angle to rotate the barrel
angle_change -= self.barrel.angle

# compensate for the vertical orientation of the sprite's image file
angle_change += 90
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+            self.barrel.rotate_around_point((self.tank.center_x, self.tank.center_y), angle_change)
+        else:
+            # In this situation we only change the angle without changing the position which is incorrect.
+
+            # we need to add 90 to the angle due to orientation of the barrel texture.
+            angle = arcade.get_angle_degrees(self.tank.center_y, self.tank.center_x,
```
pushfoo, line 144: You can apply a similar change here as suggested above for the correct rotation method.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+        self.mouse_pos = [0, 0]
+
+        self.tank_sprite_list = arcade.SpriteList()
+        self.tank_sprite_list.extend([self.tank, self.barrel])
+
+        self._correct = True
+        self.correct_text = arcade.Text("Turret Rotation is Correct, Press P to Switch",
+                                        SCREEN_WIDTH // 2, SCREEN_HEIGHT - 25, anchor_x='center')
+
+        self.control_text = arcade.Text("WASD to move tank, Mouse to aim", SCREEN_WIDTH // 2, 15, anchor_x='center')
```
pushfoo: This can be broken up across multiple lines.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
+"""
+Sprite Rotation With A Tank.
+
+The problem often fis that vehicles or tower defense turrets can have parts that can rotate toward targets.
+These parts are usually represented with separate sprites drawn relative to attachment points on the main body.
+
+Because these sprites are usually asymmetrical, we have to rotate them around their attachment points on the main body,
```
pushfoo: Try to limit line lengths in docstrings. [PEP-8 recommends 72 characters per line as the maximum for docstrings](https://peps.python.org/pep-0008/#maximum-line-length). You don't have to meet that exactly, but shorter line lengths than what you currently have are a good idea. You can split sentences across multiple lines safely as long as you aren't introducing a blank line right afterward.

arcade/examples/sprite_rotate_around_point.py
```diff
@@ -0,0 +1,98 @@
...
+SCREEN_TITLE = "Rotating Sprite Example"
+
+
+class RotatingSprite(arcade.Sprite):
+    """
+    This is an example sprite subclass which implements a generic rotate_around_point method.
+    The change_angle keyword could be removed if you know that sprites will always / never change angle.
```
pushfoo, line 23: I would add a newline before the second sentence.

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 600
+
+SCREEN_TITLE = "Rotating Tank Example"
+
+
+class RotatingSprite(arcade.Sprite):
+    """
+    This is an example sprite subclass which implements a rotate_around_point method
```
pushfoo: You can make these two lines shorter:
```
Sprite subclass which can be rotated around a point
```

arcade/examples/sprite_rotate_tank.py
```diff
@@ -0,0 +1,185 @@
...
+    which also rotates the sprites angle.
+    """
+    def rotate_around_point(self, point, degrees):
+        """
+        rotates the sprite around a defined point by the set amount of degrees
```
pushfoo: Capitalize the R at the start of this line, and remove defined

pushfoo (changes requested):
I think you should try to make the docstrings at the top shorter. You don't have to fully comply with PEP-8's 72 character limit, but it would be good to make them shorter. As a general rule for docstrings, you can split a sentence in the middle of a line safely as long as there isn't a blank newline after. Add a blank newline between sentences to mark paragraph boundaries.

einarf:
@pushfoo will address some remaining things in new PR.

PR #1180, opened 2022-04-08
Add get_angle_radians function

pushfoo:
Added to support #1176 and simplify its code, as well as help users who are using libraries that expect radians instead of degrees.


arcade/__init__.py
```diff
@@ -140,6 +140,7 @@ def configure_logging(level: int = None):
 from .geometry_generic import rotate_point
 from .geometry_generic import clamp
 from .geometry_generic import get_angle_degrees
+from .geometry_generic import get_angle_radians
```
einarf, line 143: It should also be  in `__all__` for

pushfoo: There's a comment above `__all__` that says it's generated. Is that out of date or do I need to run a script for that?

einarf: Remove that comment. It's no longer generated because it has too manu issues,

pushfoo: Done and pushed.

PR #1188, opened 2022-04-11
Fix doc for Sprite's turn_left and turn_right

pushfoo:
The doc for these two methods is currently misformatted: https://api.arcade.academy/en/latest/api/sprites.html#arcade.Sprite.turn_left

Changes in this PR:
* Fix param formatting in generated doc
* Add units to param lines
* Replace int keyword arg values with matching floats

Built and tested locally.


einarf:
👍

PR #1189, opened 2022-04-12
doc: Added code and documentation for keeping the score after a level

sabadam32:


sabadam32:
Noticed the score reset at the beginning of level 2.  I couldn't let all that hard work go to waste.

pvcraven:
Thanks!

Issue #1118, opened 2022-03-21
Enhancement: Addition of more arguments in UIInputText


Ibrahim2750mi:
### What should be added?
In [UIInputText](https://api.arcade.academy/en/latest/api/gui_widgets.html?#arcade-gui-uiinputtext) there should be an argument to
 1. only accept numeric values i.e (0, 1, 2, 3, 4, 5, 6, 7, 8, 9)  in the input box.
 2. limit the number of characters to be entered in the input box. 

### What would it help with?
1. When the game wants the user to enter its response in only numeric values.
2. When the game wants to asks for user's name etc.

pushfoo:
Maybe adding a callable validator argument to labels would be a good idea?

Ibrahim2750mi:
> Maybe adding a callable validator argument to labels would be a good idea?

Can it not be done with just regex? like `acceptable_format: re.Pattern`

pushfoo:
Remember, beginners are one of the core audiences for arcade. Requiring them to learn regex may overwhelm them if they're still struggling with the fundamentals of Python.

pushfoo:
@Ibrahim2750mi I took some more time to think about this. I agree that we should have some form of support for regex. 

However, I think callables are the best option for the validator argument. They would allow us to have a decent balance of beginner friendliness and accessibility for advanced users. Advanced users can do whatever they want, while beginners could use default helper functions to handle things like regex, integers, etc. I'm largely taking inspiration from [attr's validators](https://www.attrs.org/en/stable/examples.html#validators):
```py
from typing import Any, Callable
import arcade
# the validators package would include reasonable presets for regex, digits, etc
from arcade.validation import validators

class ValidatedWidget:
    """
    Mock-up of how we might implement validation. 
    
    It might not align with the exact widget implementation we have now
    """
    def __init__(self, value: Any = None, validator: Callable = None):
        self._validator = validator
        self._value = None

        # use the property setter to validate and set
        self.value = value
       
    @property
    def value(self) -> Any:
        return self._value

    @value.setter
    def value(self, new: Any):
        if self._validator and not self._validator(new):
            # we'd need a much better reporting mechanism in the final version
            raise ValueError(f"Invalid value for validator {self._validator}")
        else:
            self._value = new


widget = ValidatedWidget(
    value="123",
    # this validator function would do type checking once and return a callable wrapping a regex 
    validator=validators.base_10_integer
)

# this produces an exception
widget.value = "1f4b"
```

[...]

PR #1195, opened 2022-04-18
Added validation to UIInputText

eschan145:
The validation limits what the user can type. For example, if the parameter validate is set to "0123456789", the user can only type numerical digits into the input. The maximum parameter is the maximum number of characters the user can input.

### DIFF
--- a/arcade/gui/widgets.py
+++ b/arcade/gui/widgets.py
@@ -18,6 +18,7 @@ from arcade.gui.surface import Surface
 if TYPE_CHECKING:
     from arcade.gui.ui_manager import UIManager
 
+from string import printable
 
 class _Rect(NamedTuple):
     x: float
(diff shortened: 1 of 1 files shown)


Ibrahim2750mi:
#1118

eruvanos:
Will have a look on it

pushfoo:
I'm not sure that this PR's approach is the best in the long run. I mentioned it as a [comment on #1118](https://github.com/pythonarcade/arcade/issues/1118#issuecomment-1115422519), but tl;dr I think we should use callables as the validator arguments. The functionality included in this PR could be one of the default validator callables we provide to users.

pvcraven:
I think input validation will need a lot more thought on how (of if) we want to manage it in the main library.

Issue #991, opened 2021-10-06
Add support for drawing infinite repeating textures with offset


pushfoo:
## Enhancement request:
**tl;dr** add a way to draw infinite repeating textures to screen, with an offset. 

### What should be added/changed?

A camera-sensitive way to draw a repeating texture in the background, using the player's position to offset the texture drawing.  This should probably be implemented somewhere using lower level OpenGL calls rather than `Sprite` objects.

Based on [discord discussion](https://discord.com/channels/458662222697070613/458662458198982676/895129822416732231) with @einarf , this could take the form of a `TextureGrid` object. It could have `repeat_x` and `repeat_y` parameters that specify how many times a passed pattern or individual texture should repeat in a given direction. The default could be infinite for ease of use.

This ticket does not necessarily include integration with loading from Tiled or other map sources, and that may be a separate ticket.

### What would it help with?

**tl;dr** provides an easy way to have easy, instant, and infinite non-interactive background for top-down games

For beginners who want to make top down games with restricted view, one of the easiest ways to get started is to have the camera always center on the player. This frees beginners from the difficulties common to writing good camera behavior that is aware of environments.

However, a solid-color background will not clearly show that the player's character actually moving in the world. There have to be external visual cues aside from player sprite animation to show that player characters are actually moving. Adding support for a non-sprite infinite background pattern allows the following:
1. Allow beginners to get started quickly
2. Avoid overhead for creating sprite objects for background layers
3. Improve developer experiences by enabling quicker iteration
4. Provide a default "empty" grid background for games that generate terrain

einarf:
Closing this as a duplicate for now

PR #1197, opened 2022-04-22
Background class, and 5 examples.

### LINKED ISSUE #1152: Complete and add a minimal BackgroundTexture class
Having a `BackgroundTexture` class solves quite a few use cases that are difficult for most users to implement without having to use shaders. The name can still be debated. It's pretty close to a `RenderTargetTexture`  currently in experimental, so maybe we should look into that relation. The `BackgroundTexture` is just a wrapper over an `arcade.gl.Texture`. 

Some ponts:

* It should not use the texture atlas because using this type is a good way to avoid flooding the atlas with large images
* Default initializer can take a width and height
* `BackgroundTexure.from_file(path)` would load an image from file. Possibly using the texture cache
* Support scrolling texture coordiantes in pixels
* Support rotating texture coordiantes with preserved aspect ratio
* Support enabling and disabling texture repeat
* Support enabling mipmaps (useful if zooming)
* Support setting a `Mat3` directly for finer control over texture coordinate transformations
* It can also work as a render target, so it also needs to be a framebuffer. `with bg.enabled()`
* Should it have the same pixel scale as the window framebuffer?
* Create a default texture shader users can load. It should be compatible with the vertex format in `arcade.gl.geometry`.

More thinking is needed.

DragonMoffon:
The push Implements a background class, and BackgroundTexture, BackgroundGroup, and ParallaxGroup. Plus five examples: static Background, scrolling Background, BackgroundGroup, ParallaxGroup, and Blending between Backgrounds.

These close #1152 and #991


arcade/__init__.py
```diff
@@ -294,6 +294,11 @@ def configure_logging(level: int = None):
 from .sprite_list import get_sprites_at_exact_point
 from .sprite_list import get_sprites_at_point
 
+from .backgrounds import BackgroundTexture
+from .backgrounds import Background
+from .backgrounds import BackgroundGroup
+from .backgrounds import ParallaxGroup
```
einarf: I think we should stop adding things to the arcade namespace. It's getting way too cluttered. Users having to import from the `arcade.backgrounds` module is fine.

arcade/backgrounds.py
```diff
@@ -0,0 +1,408 @@
+from PIL import Image
+
+from arcade.context import ArcadeContext
+from arcade.window_commands import get_window
+from arcade.resources import resolve_resource_path
+import arcade.gl as gl
+from pyglet.math import Mat3
+
+
+class BackgroundTexture:
```
einarf: Possibly break this module into smaller parts so we can build on it easier. `arcade/backgrounsd` directly with an `__init__py` importing the classes located in sub-modules.

arcade/backgrounds.py
```diff
@@ -0,0 +1,408 @@
...
+class BackgroundGroup:
+    """
+    If you have many backgrounds which you would like to draw together and move together this can help.
+
+    The pos of the Background Group is independent of each Background pos.
+    The offset of the BackgroundGroup is the same as each background.
+    """
+
+    def __init__(self):
+        self.backgrounds: list[Background] = []
```
einarf: It's probably better to keep this as a private member and expose a backgrounds property with proper docstrings.

arcade/resources/shaders/background_fs.glsl
```diff
@@ -0,0 +1,26 @@
...
+uniform float blend = 1;
+
+uniform vec2 pos;
+uniform vec2 size;
+uniform vec2 bounds;
+
+out vec4 fragColor;
+
+void main() {
+    vec2 texSize = vec2(textureSize(backgroundTexture, 0));
```
einarf, line 19: Implicit conversion from int to float

arcade/resources/shaders/background_fs.glsl
```diff
@@ -0,0 +1,26 @@
...
+    vec2 adjustedUV = frag_uv * size;
+
+    vec2 adjusted = (pixelTransform * vec3(adjustedUV, 1.0)).xy;
+
+    adjusted = adjusted / texSize;
+    fragColor = texture(backgroundTexture, adjusted, 0);
```
einarf, line 25: Bias argument it optional. Just leave it out

einarf:
Some quick intial comments. This actually looks really solid, but I am not 100% sure about the details around what properties should be on the texture and the background itself to give the most flexbility.

Also there are missing some type annotations here, but I think fixing the more obvious things and just merging it is a better approach here. Play around with the feature and tweak things instead.

Adding a RGBA color multiplier can also add a lot of flexiblity here instead of just blend. We can really merge this as it is now to be honest. Knitpicking on smaller things that are subject to change around a little bit won't move us forward.

DragonMoffon:
For the colour should, do you think I should apply via multiply or addition? Currently, have it as multiply, but if someone wanted to fade in the color, it would be easier if it was added. 

Also, I have it supporting both normalised and not, but I decide if the colour is normalised by seeing if the sum of the colour is less than or equal to 3, which may cause some confusing bugs. Is there a better way of deciding?

einarf:
* We do color multiplication everywhere in arcade right now, so that would make the most sense
* If you want to support byte RGBA and normalized RGBA color they can have different properties. `color`, `color_normalized`

DragonMoffon:
Some of the functions are still missing docstrings. Which will need to be fixed before 2.7 is out. But otherwise, it should be mergeable.

PR #1204, opened 2022-04-27
Updated sprite_health.py example to use an indicator bar object

### LINKED ISSUE #1202: Rewrite sprite_health.py example
The current code has multiple things wrong with it:
1. Uses `arcade.draw_text` instead of cached `arcade.Text` instances
2. Uses shape drawing rather than stretched sprites
3. Uses slow m x n collision instead of something more reasonable 
4. Formatting of the code can be significantly improved
5. No on-screen instructions for how to play the example

It should be rewritten to be simpler and more readable.

### What would it help with?
First, it will show beginners a better way to do health bars.

Second, we have PyCon coming up, with a mentored sprint for the [Community RPG project](https://github.com/pythonarcade/community-rpg) as part of it. Having a reasonable health bar implementation as reference in the doc would give participants an ea

JackAshwell11:
I have updated the sprite_health.py example to use an indicator bar object. This should be much more efficient than the previous example due to it not using draw_text or shapes. This fixes #1202


doc/examples/sprite_health.rst
```diff
@@ -8,7 +8,10 @@ Hit Points and Health Bars
 .. image:: sprite_health.png
     :width: 600px
     :align: center
-    :alt: Screenshot of using sprites to shoot things
+    :alt: Screenshot of an enemy shooting at a player with an indicator bar
+
+This example demonstrates how a enemy can shoot a bullet towards a player. Then if the bullet hits, the player's health
```
pushfoo: There are multiple suggestions I would like to make here.

First, try to keep line length shorter in RSTs. 72-80 seems reasonable.

Second, I would rephrase this part. The first sentence should explain what is being demonstrated, ie:
```
This example demonstrates a reasonably efficient way of drawing a health bar above a character.
```
The rest of the text can explain how it does so, and how to play:
```
The enemy at the center of the screen shoots bullets at the player, 
while the player attempts to dodge the bullets by moving the mouse.
Each bullet that hits the player reduces the player's health, which
is shown by the bar above the player's head. When the player's health
bar is empty (zero), the game ends.
```

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-    def setup(self):
-
-        """ Set up the game and initialize the variables. """
-
-        # Sprite lists
-        self.player_list = arcade.SpriteList()
-        self.coin_list = arcade.SpriteList()
         self.bullet_list = arcade.SpriteList()
+        self.bar_list = arcade.SpriteList()
+        self.player_sprite = Player(":resources:images/animated_characters/female_person/femalePerson_idle.png", SPRITE_SCALING_PLAYER, self.bar_list)
```
pushfoo: PEP8 and our style conventions both suggest shorter line lengths than this when possible. It's especially important for the embedded code views we use to display examples in the online documentation.

I would recommend putting these strings into constants at the top, or using the pre-defined constants in the `arcade.resources` module.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
+        super().__init__(filename=filename, scale=scale)
+        self.indicator_bar: IndicatorBar = IndicatorBar(self, bar_list, (self.center_x, self.center_y))
+        self.health: int = PLAYER_HEALTH
+
+
+class IndicatorBar:
+    """
+    Represents a bar which can display information about a sprite.
+
+    Parameters
```
pushfoo: It's good that you've matched arcade's unusual convention for where constructor arguments are documented.

However, it would be useful if you matched the format used elsewhere. Aside from line length issues, the docstring for `arcade.Camera` is fairly good reference:
https://github.com/pythonarcade/arcade/blob/98fb1809363ccc537d6852be487aeae0b5fb7fb8/arcade/camera.py#L12-L22

You can omit the blank line at the end.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-    def on_mouse_press(self, x, y, button, modifiers):
-        """
-        Called whenever the mouse button is clicked.
-        """
-        # Gunshot sound
-        arcade.play_sound(self.gun_sound)
-        # Create a bullet
-        bullet = arcade.Sprite(":resources:images/space_shooter/laserBlue01.png", SPRITE_SCALING_LASER)
+        # Update the player's indicator bar position
+        self.player_sprite.indicator_bar.position = self.player_sprite.center_x, self.player_sprite.center_y + INDICATOR_BAR_OFFSET
```
pushfoo: This can be split across multiple lines to make it legible in the web view:
```suggestion
self.player_sprite.indicator_bar.position =\
    self.player_sprite.center_x,\
    self.player_sprite.center_y + INDICATOR_BAR_OFFSET
```

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-        bullet.change_y = BULLET_SPEED
+        # Check if the enemy can attack. If so, shoot a bullet from the enemy to the player
+        if self.enemy_timer >= ENEMY_ATTACK_COOLDOWN:
+            self.enemy_timer = 0
 
-        # Position the bullet
-        bullet.center_x = self.player_sprite.center_x
-        bullet.bottom = self.player_sprite.top
+            # Create the bullet
+            bullet = arcade.Sprite(":resources:images/space_shooter/laserBlue01.png", SPRITE_SCALING_BULLET)
```
pushfoo: There are at least two ways of keeping this code legible when embedded in a narrow web view:
1. Split the arguments up across multiple lines
2. Use a constant for the laser path at the top.

In this case, I favor the first because we already have a lot of constants.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-        # Create a bullet
-        bullet = arcade.Sprite(":resources:images/space_shooter/laserBlue01.png", SPRITE_SCALING_LASER)
+        # Update the player's indicator bar position
+        self.player_sprite.indicator_bar.position = self.player_sprite.center_x, self.player_sprite.center_y + INDICATOR_BAR_OFFSET
 
-        # The image points to the right, and we want it to point up. So
-        # rotate it.
-        bullet.angle = 90
+        # Call updates on bullet sprites
+        self.bullet_list.update()
```
pushfoo: `SpriteList` instances also have an `on_update` method that calls `on_update` for each sprite with the passed dt.

I know it'll be ugly, but it might be worth adding a `Bullet` class in this example with an implemented `on_update` function that takes a dt to make up for [`Sprite`'s empty `on_update` implementation](https://github.com/pythonarcade/arcade/blob/98fb1809363ccc537d6852be487aeae0b5fb7fb8/arcade/sprite.py#L1039-L1043). This way, bullets won't jump unexpectedly if the frame dt changes. Per discord discussion with einarf, [we'll be making `update` a dt-based function in 2.7](https://discord.com/channels/458662222697070613/458662458198982676/968955131607547984). Until then, an ugly but reliable example would be a good stopgap.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
         # Loop through each bullet
         for bullet in self.bullet_list:
-
-            # Check this bullet to see if it hit a coin
-            hit_list = arcade.check_for_collision_with_list(bullet, self.coin_list)
-
-            # If it did, get rid of the bullet
-            if len(hit_list) > 0:
+            # Check if the bullet has gone off-screen. If so, kill it
+            if bullet.top < 0 or bullet.bottom > SCREEN_HEIGHT or bullet.right < 0 or bullet.left > SCREEN_WIDTH:
```
pushfoo: To keep line length short, I'd consider abstracting this check into a function at the top of the file. Something with a signature like this:
```py
def sprite_off_screen(sprite, screen_height=SCREEN_HEIGHT, screen_width=SCREEN_WIDTH) -> bool:
```
The resulting if statement would then read like this:
```py
if sprite_off_screen(bullet):
    bullet.remove_from_sprite_lists()
    continue
```

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-            # If the bullet flies off-screen, remove it.
-            if bullet.bottom > SCREEN_HEIGHT:
+            # Check if the bullet has hit the player
+            if arcade.check_for_collision(bullet, self.player_sprite):
+                # Damage the player and remove the bullet
+                self.player_sprite.health -= BULLET_DAMAGE
                 bullet.remove_from_sprite_lists()
 
+                # Set the player's indicator bar fullness
+                self.player_sprite.indicator_bar.fullness = self.player_sprite.health / PLAYER_HEALTH
```
pushfoo: line length, you could put a `'\'` after the equals and put the rest on a new line.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
 import arcade
 
 SPRITE_SCALING_PLAYER = 0.5
-SPRITE_SCALING_COIN = 0.2
-SPRITE_SCALING_LASER = 0.8
-COIN_COUNT = 50
+SPRITE_SCALING_ENEMY = 0.5
+SPRITE_SCALING_BULLET = 1
+INDICATOR_BAR_OFFSET = 32
+ENEMY_ATTACK_COOLDOWN = 0.5
```
pushfoo: Playing this with the current values is a bit frantic and overwhelming, especially with a touchpad.

I think you should do one of the following to make the example easier to understand on first launch:
1. Increase `ENEMY_ATTACK_COOLDOWN` to 1.0, and decrease `PLAYER_HEALTH` to 5
2. Set `BULLET_SPEED` to 2

I tried playing both changes locally, and I think either is ok. You could also combine the approaches, but you will need to fiddle with the values to find something that plays reasonably fast without being overwhelming.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
+
+    @property
+    def position(self) -> tuple[float, float]:
+        """Returns the current position of the bar."""
+        return self._center_x, self._center_y
+
+    @position.setter
+    def position(self, new_position: tuple[float, float]) -> None:
+        """Sets the new position of the bar."""
+        # Check if the position has changed. If so, change the bar's position
```
pushfoo, line 144: I would add a blank line between the docstring and the comment to increase legibility.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
+
+    @property
+    def fullness(self) -> float:
+        """Returns the fullness of the bar."""
+        return self._fullness
+
+    @fullness.setter
+    def fullness(self, new_fullness: float) -> None:
+        """Sets the fullness of the bar."""
+        # Check if new_fullness if valid
```
pushfoo, line 119: I would add a blank line between the docstring and the comment to increase legibility.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
-    def on_draw(self):
-        """
-        Render the screen.
-        """
+        self.background_color = arcade.color.AMAZON
 
-        # This command has to happen before we start drawing
+    def on_draw(self) -> None:
+        """Render the screen."""
+        # Clear the screen. This command has to happen before we start drawing
```
pushfoo, line 174: I would add a blank line between the docstring and the comment to increase legibility.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
 
         # Loop through each bullet
         for bullet in self.bullet_list:
-
-            # Check this bullet to see if it hit a coin
-            hit_list = arcade.check_for_collision_with_list(bullet, self.coin_list)
-
-            # If it did, get rid of the bullet
-            if len(hit_list) > 0:
+            # Check if the bullet has gone off-screen. If so, kill it
```
pushfoo: This comment is low priority if you're crunched for time.

I know we have kill as an alias for `remove_from_sprite_lists`, but I think that phrasing might confuse new players. Maybe delete would be better since it's not alive?

doc/examples/sprite_health.rst
```diff
@@ -8,7 +8,10 @@ Hit Points and Health Bars
 .. image:: sprite_health.png
     :width: 600px
     :align: center
-    :alt: Screenshot of using sprites to shoot things
+    :alt: Screenshot of an enemy shooting at a player with an indicator bar
```
pushfoo: consider "health indicator bar" instead of "indicator bar"

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,249 @@
...
+        self,
+        owner: Player,
+        sprite_list: arcade.SpriteList,
+        position: tuple[float, float] = (0, 0),
+        full_color: arcade.Color = arcade.color.GREEN,
+        background_color: arcade.Color = arcade.color.BLACK,
+        width: int = 100,
+        height: int = 4,
+        border_size: int = 4,
+    ) -> None:
```
pushfoo, line 99: Consider adding `arcade.Text` instances with brief play instructions.

Top text:
```
Dodge the bullets by moving the mouse!
```

Bottom text:
```
When your health bar reaches zero, you lose!
```

pushfoo (changes requested):
I played through it locally, and I think it's pretty good! :smile:

I have the following major concerns:

**Frantic Pace Mixes Poorly With Touchpads**
The current constant values feel frantic and hard to understand on first play, especially on a touchpad. Remember, this may end up being used as reference during a PyCon workshop, so most attendees will be on laptops. Not all of those will have attached mice or good touchpads!  More on this in the file comments.

**Line Length**
Visualizing line length can be annoying. If you haven''t already, it's very helpful to build and view the embedded html version of the doc.
1. Make sure dev requirements are installed: from the root of the repo, run `pip install -e .[dev]`
2. Move into the doc dir with `cd doc`
3. build the html: `make html`
4. Serve the built html locally: `python -m http.server -d build/html`
5. Use your browser to read through the embedded code at http://localhost:8000/examples/sprite_health.html#sprite-health

arcade/examples/sprite_health.py
pushfoo: Take einarf's advice from discord over my initial suggestion on this: the parentheses are probably the more elegant option.

pushfoo:
I built the recent commits locally. Aside from a minor issue on line 49, the line lengths look great.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,297 @@
...
 
-        # Add the bullet to the appropriate lists
-        self.bullet_list.append(bullet)
+        # Call updates on bullet sprites
+        self.bullet_list.on_update(delta_time)
 
-    def on_update(self, delta_time):
-        """ Movement and game logic """
+        # Check if the enemy can attack. If so, shoot a bullet from the
+        # enemy to the player
```
pushfoo: This change would be more grammatically correct and clearer.
```suggestion
        # enemy toward the player
```
Although `to` is shorter, it could be read as there being some certainty the bullet will hit.

doc/examples/sprite_health.rst
```diff
@@ -8,7 +8,14 @@ Hit Points and Health Bars
 .. image:: sprite_health.png
     :width: 600px
     :align: center
-    :alt: Screenshot of using sprites to shoot things
+    :alt: Screenshot of an enemy shooting at a player with an health indicator bar
+
+This example demonstrates a reasonably efficient way of drawing a health
+bar above a character. The enemy at the center of the screen shoots
```
pushfoo: I would add a blank line after the first sentence to separate it from the rest of the paragraph below.

arcade/examples/sprite_health.py
```diff
@@ -1,230 +1,297 @@
...
+        sprite.top < 0
+        or sprite.bottom > screen_height
+        or sprite.right < 0
+        or sprite.left > screen_width
+    )
+
+
+class Player(arcade.Sprite):
+    def __init__(self, bar_list: arcade.SpriteList) -> None:
+        super().__init__(filename=image_female_person_idle, scale=SPRITE_SCALING_PLAYER)
```
pushfoo: When examples are built as html doc, this line currently extends beyond the right boundary of the embedded view. Splitting the line up would help.
```suggestion
        super().__init__(
            filename=image_female_person_idle,
            scale=SPRITE_SCALING_PLAYER
        )
```

pushfoo (changes requested):
I hid resolved comments and deleted ones that aren't crucial. If these changes are made, this will be ready for merge.

pushfoo (approved):
Played, built, and viewed locally. Imo, this is ready to merge! :tada:  Thank you for getting this done so quickly!

Any remaining style issues aren't important enough to deal with right now given the PyCon RPG workshop are coming up.

Issue #1198, opened 2022-04-25
Supporting for setting pitch of sounds


Cleptomania:
Pyglet has support for setting the pitch of a sound by adjust the `pitch` attribute of a Player object. We can easily add this support to Arcade by adding a pitch argument to `arcade.sound.play()` with a default of 1.0 and passing it through to the player.

iansedano:
I'll try this one

Cleptomania:
This has been implemented with #1208

pushfoo:
I don't quite understand how it works, so beginners will probably be confused as well. Maybe we could clarify the documentation and add a tiny piano example that bends the pitch of a single sample to cover 1 octave?

pushfoo:
The [pyglet doc](https://pyglet.readthedocs.io/en/latest/modules/media.html#pyglet.media.player.Player.pitch) makes the underlying behavior much clearer:
> The pitch shift to apply to the sound.
>
>The nominal pitch is 1.0. A pitch of 2.0 will sound one octave higher, and play twice as fast. A pitch of 0.5 will sound one octave lower, and play twice as slow. A pitch of 0.0 is not permitted.

Something like `play_speed` might be clearer for our users, even if it isn't a 1:1 match to pyglet's argument name.

iansedano:
I agree that `pitch` is not such an intuitive attribute. As a beginner, you might expect the pitch to be shifted but the length to remain the same. Likewise, if you change the speed, you might assume that the pitch would stay the same.

So, perhaps we could implement some additional methods to preserve pitch when changing speed and preserving speed (or length) when changing pitch. That also has the additional advantage of keeping the same argument names, just adding a bit of user-facing functionality to pyglet.

If we did this, then we might have an additional argument to bypass this "corrective" default behavior. In the case of `pitch` it could be `compensate_speed=True`, for instance.

However, looking over the pyglet docs, I can't immediately see an obvious way to achieve this. I'll look into it a bit further, but I think `play_speed` is a good term for what is happening for now. The pitch shifting behavior can be clarified in the docs.

pushfoo:
Although a a few audio tools tools implement pitch shifting while preserving playback speed, it's lossy and can sound very strange. I think we might want to do the following:

1. rename the argument to one of the following before a release gets cut:`play_speed`, `playback_speed`, or `speed`
2. put off further changes to the playback until we understand the problem better
3. leave this issue open

[...]

PR #1212, opened 2022-05-04
Add a color fill classmethod to Texture

pushfoo:
## tl;dr
1. Added `arcade.Texture.create_filled`, which is like `create_empty` but takes a color
2. Made `arcade.SpriteSolidColor` use it
3. Tested affected examples and building the doc, all looks ok

### 1. Easier prototyping for end users
Some users may begin their project by displaying game state through`Sprite.color` changes applied to solid white `SpriteCircle` or `SpriteSolidColor` instances. However, it can be annoying to rewrite such code to use texture data.

Merging this PR will make it easier for users to generate filler textures from the start. Switching to loading texture data from files is easy, regardless of whether they use [`load_textures`](https://api.arcade.academy/en/latest/api/texture.html#arcade-load-textures), [`load_spritesheet`](https://api.arcade.academy/en/latest/api/texture.html#arcade-load-spritesheet), or another method of their choice.

The code for projects that expand on #1204 could also be cleaner as a result. For example, some RTS games use a convention of red for low health, yellow for medium health, and green for full health. Users will no longer need to understand Pillow's API or image channels to achieve this.

### 2. Make writing examples easier
The piano example for #1198 will be easier to write cleanly with this merged.

### 3. Avoid memory leaks
It's unclear whether this comment is still relevant, but [`Sprite.color` may create memory leaks](https://github.com/pythonarcade/arcade/pull/974#issuecomment-931342070):
> [pvcraven](https://github.com/pvcraven) commented [on Sep 30, 2021](https://github.com/pythonarcade/arcade/pull/974#issuecomment-931342070)
My initial, quick-look-through, makes me think this is a good change.
>
>If someone color-cycles we'll have a memory leak, but that will mess up the atlas anyways. Makes me wonder if there should be docs somewhere that explain to use .color or otherwise explain the internals a bit.

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -1375,8 +1375,7 @@ class SpriteSolidColor(Sprite):
 
         # otherwise, generate a filler sprite and add it to the cache
         else:
-            image = PIL.Image.new("RGBA", (width, height), color)
-            texture = Texture(cache_name, image)
+            texture = Texture.create_filled(cache_name, (width, height), color)
             load_texture.texture_cache[cache_name] = texture  # type: ignore
 
         # apply chosen texture to the current sprite
--- a/arcade/texture.py
+++ b/arcade/texture.py
@@ -128,7 +128,7 @@ class Texture:
         background color, you can save CPU time and RAM by calling this
         function once, then passing the ``image`` attribute of the
         resulting Texture object to the class constructor for each
-        additional blank Texture instance you would like to create.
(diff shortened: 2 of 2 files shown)


pushfoo:
A few questions for reviewers:
1. Should I add a series creation method that returns `List[Texture]` from a parameter sequence?
2. Is the current method name ok?
3. Should `color` have a default value?

Issue #1206, opened 2022-04-29
Add row and column span in UIGridLayout


Ibrahim2750mi:
## Enhancement request:

Column span and row span should be added to UIGridLayout.
Basic concept is that the center of the widget which has a row span _n_ should have the y-coordinate as the center of sum of each widget's max height of the previous column till row number _n_. For column span the center of the widget which has a column span _n_ should have the x-coordinate as the center of sum of each widget's max width of the previous column till columns number _n_. If there is no widgets to compare for then it will behave as if like it has its col/row span = 1.

### Examples

Example for column span = 2 aligned by center:

![column span = 2](https://i.imgur.com/UlRtP9o.png)

Example for row span = 2 aligned by center:

![row span = 2](https://i.imgur.com/j0eMXke.png)

eruvanos:
merged

PR #1214, opened 2022-05-05
UIGridLayout column and row span

Ibrahim2750mi:
#1206


eruvanos:
Hi could you also add a test, which checks that the dummies get the right position and size?

PR #1216, opened 2022-05-07
Rename pitch to speed, added example, and tweaked existing sound example

iansedano:
Mainly renaming the `pitch` parameter to `speed` as per #1198. I also reverted the change to the `asteroid_smasher.py` example and created a standalone `sound_speed_demo.py` example.

I also tweaked the original `sound_demo.py` to make it a bit more concise before I created the speed demo from that template. I thought about renaming `sound_demo.py`, since it is mainly demonstrating panning. Further, I'd probably move the streaming button (in `sound_demo.py`) into it's own example, because it seems to be demonstrating something different, and it also crashes if you press it twice in quick succession.

I didn't yet, though, because that would probably mess with the diff, and that could be done in another PR.

### DIFF
--- a/arcade/examples/sound_demo.py
+++ b/arcade/examples/sound_demo.py
@@ -1,11 +1,5 @@
 """
-Starting Template
-
-Once you have learned how to use classes, you can begin your program with this
-template.
-
-If Python and Arcade are installed, this example can be run from the command
-line with:
+Sound Panning Demo
 
 python -m arcade.examples.sound_demo
 
@@ -23,8 +17,11 @@ import arcade
 
 SCREEN_WIDTH = 800
 SCREEN_HEIGHT = 600
-SCREEN_TITLE = "Starting Template"
+SCREEN_TITLE = "Sound Panning Demo"
 BUTTON_SIZE = 30
+
+SOUND_PANNING = [-1.0, -0.5, 0.0, 0.5, 1.0]
+assert len(SOUND_PANNING) == 5
 
 
 class SoundButton(arcade.SpriteSolidColor):
@@ -74,19 +71,9 @@ class AudioStreamButton(arcade.SpriteSolidColor):
 
 
 class MyGame(arcade.Window):
-    """
-    Main application class.
-
-    NOTE: Go ahead and delete the methods you don't need.
-    If you do need a method, delete the 'pass' and replace it
-    with your own code. Don't leave 'pass' in this program.
-    """
-
     def __init__(self, width, height, title):
         super().__init__(width, height, title)
-
         arcade.set_background_color(arcade.color.AMAZON)
-
         self.button_sprites = None
 
     def setup(self):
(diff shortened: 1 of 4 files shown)


iansedano:
I'm seeing that the tests are failing due to my "table", any ideas? If not, I'll find another way, later.

image

Ibrahim2750mi:
> I'm seeing that the tests are failing due to my "table", any ideas? If not, I'll find another way, later.

Does doing "flake8 file_name.py" return nothing? Also theres a discussion(on discord) going on whether to change the naming of  `pitch` or not.

pushfoo:
@iansedano you can click `details` on the right-hand side of any CI check to see more information:
image
For either of the runs above, doing should automatically scroll you to the code-inspection section, which will list issues:
image
The way you have your arguments formatted isn't something that python likes. I'll try to make some suggestions on a more pythonic way to format your code in the next 24 hours.

arcade/examples/sound_demo.py
```diff
@@ -93,86 +101,30 @@ def setup(self):
...
-        self.button_sprites.append(button)
-
-        button = SoundButton(":resources:sounds/upgrade4.wav", pan=1, volume=volume)
-        button.center_x = SCREEN_WIDTH - BUTTON_SIZE
-        button.center_y = y
-        self.button_sprites.append(button)
+        upgrade_sound = ":resources:sounds/upgrade4.wav"
+
+        # fmt: off
+        button_params = [
```
pushfoo: Even if we think it looks nice, python style guides strongly disagree with this sort of formatting. :frowning: 

One way to deal with it is to span this across two for loops: the outer one handles setting the row y position, and the inner one passes updated x position, `pan`, and `vol` values.

If you want to be explicit about the panning volumes, maybe you can put those in a tuple constant stored at the top?

pushfoo: Even if we think it looks nice, python style guides strongly disagree with this sort of formatting.

One way to deal with it is to span this across two for loops: the outer one handles setting the row y position, and the inner one passes the `pan` and `vol` columns in.

If you want to be explicit about the panning volumes, maybe you can put those in a tuple constant stored at the top?

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def draw(self):
+        super().draw(self)
+
+
+class MyGame(arcade.Window):
+    """
+    Main application class.
+
+    NOTE: Go ahead and delete the methods you don't need.
```
pushfoo: I would delete this comment, as well as the methods that are stubbed out with `pass`.

pushfoo: I would delete this comment, as well as the methods that are stubbed out with `pass`.

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+        self.button_sprites = None
+
+    def setup(self):
+        self.button_sprites = arcade.SpriteList()
+
+        upgrade_sound = ":resources:sounds/gameover3.wav"
+
+        # fmt: off
+        button_params = [
+        #    sound        , speed, vol   , x                          , y
```
pushfoo: You could probably use the same nested for loop for loop approach suggested above here.

pushfoo: You could probably use the same nested for loop for loop approach suggested above here.

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+            ):  # right click to increase volume on currently playing sound
+                if not button_sprite.sound.is_complete():
+                    button_sprite.volume_up()
+                    button_sprite.stream_position()
+            elif button == arcade.MOUSE_BUTTON_MIDDLE:
+                if not button_sprite.sound.is_complete():
+                    button_sprite.volume_down()
+                    button_sprite.stream_position()
+
+    def on_mouse_release(self, x, y, button, key_modifiers):
```
pushfoo: probably best to delete this if we're not using it.

pushfoo: probably best to delete this if we're not using it.

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def on_update(self, delta_time):
+        """
+        All the logic to move, and the game logic goes here.
+        Normally, you'll call update() on the sprite lists that
+        need it.
+        """
+        self.button_sprites.update()
+
+    def on_key_press(self, key, key_modifiers):
```
pushfoo: delete this method if it's just a stub

pushfoo: delete this method if it's just a stub

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        Called whenever a key on the keyboard is pressed.
+
+        For a full list of keys, see:
+        https://api.arcade.academy/en/latest/arcade.key.html
+        """
+        pass
+
+    def on_key_release(self, key, key_modifiers):
```
pushfoo: delete this method if it's just a stub

pushfoo: delete this method if it's just a stub

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        Called whenever the user lets off a previously pressed key.
+        """
+        pass
+
+    def on_mouse_motion(self, x, y, delta_x, delta_y):
```
pushfoo: delete this method if it's just a stub

pushfoo: delete this method if it's just a stub

arcade/examples/text_loc_example_done.py
```diff
@@ -56,7 +56,11 @@ def on_draw(self):
         start_y = 450
         arcade.draw_point(start_x, start_y, arcade.color.BLUE, 5)
         arcade.draw_text(
-            _("Simple line of text in 12 point"), start_x, start_y, arcade.color.BLACK, 12
+            _("Simple line of text in 12 point"),
```
pushfoo: Although this change makes sense, why is it in this PR? I thought this was supposed to be sounds-specific.

pushfoo: Although this change makes sense, why is it in this PR? I thought this was supposed to be sounds-specific.

arcade/sound.py
```diff
@@ -58,7 +58,7 @@ def play(
         player: media.Player = media.Player()
         player.volume = volume
         player.position = (pan, 0.0, math.sqrt(1 - math.pow(pan, 2)))  # used to mimic panning with 3D audio
-        player.pitch = pitch
+        player.pitch = speed # note that the underlying attribute is pitch but "speed" is used because it better describes the behavior see #1198
```
pushfoo: It's probably best to break this up as a multiple comment lines above setting the pitch on the player. The linters won't like this as-is.

pushfoo: It's probably best to break this up as a multiple comment lines above setting the pitch on the player. The linters won't like this as-is.

pushfoo (changes requested):
I think there's room for changes beyond ones to make the linters happy.

iansedano:
Thanks for the review @pushfoo!

I've made the changes you mentioned, mainly changing the way the buttons were created and general cleaning up.

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,113 @@
...
+
+SCREEN_WIDTH = 800
+SCREEN_HEIGHT = 600
+SCREEN_TITLE = "Sound Speed Demo"
+BUTTON_SIZE = 30
+
+SPEED_VARIATION = [0.1, 0.5, 1.0, 2.0, 4.0]
+assert len(SPEED_VARIATION) == 5
+for speed in SPEED_VARIATION:
+    assert speed > 0
```
pushfoo, line 26: If you're trying to ensure the length of the array is the same for the x positions and the speeds, it might make more sense to generate the values dynamically in one place.

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,113 @@
...
+        self.button_sprites = None
+
+    def setup(self):
+        self.button_sprites = arcade.SpriteList()
+
+        game_over_sound = ":resources:sounds/gameover3.wav"
+
+        volume_variation = [0.1, 0.5, 1]
+        y_offset = [50, 0, -50]
+        button_x_spread = [
```
pushfoo, line 63: This could be either a constant at the top of the file or dynamically generated.

arcade/examples/sound_demo.py
```diff
@@ -93,140 +88,43 @@ def setup(self):
...
+        button_x_spread = [
+            BUTTON_SIZE,
+            SCREEN_WIDTH / 4,
+            SCREEN_WIDTH / 2,
+            SCREEN_WIDTH / 4 * 3,
+            SCREEN_WIDTH - BUTTON_SIZE,
+        ]
+
+        for vol, y_offset in zip(volume_variation, y_offset):
+            for pan_setting, x_pos in zip(SOUND_PANNING, button_x_spread):
```
pushfoo, line 103: Although this `zip`-based approach might confuse some beginners, it's very concise and I think I like the idea. :+1:

It'd be helpful for beginners if you add some comments with the following:
1. A few words explaining what `zip` does
2. A link to the python doc for `zip`: https://docs.python.org/3/library/functions.html#zip

Maybe something like this?
```py
# Arrange the buttons in a grid
# The zip function puts pieces of two iterables together
# For more information, see the python doc:
# https://docs.python.org/3/library/functions.html#zip
```

arcade/examples/sound_speed_demo.py
```diff
@@ -0,0 +1,113 @@
...
+            BUTTON_SIZE,
+            SCREEN_WIDTH / 4,
+            SCREEN_WIDTH / 2,
+            SCREEN_WIDTH / 4 * 3,
+            SCREEN_WIDTH - BUTTON_SIZE,
+        ]
+
+        for vol, y_offset in zip(volume_variation, y_offset):
+            for speed_setting, x_pos in zip(SPEED_VARIATION, button_x_spread):
```
pushfoo, line 72: Same comment as for the other example: it's unexpected but it's concise :+1:. Add a brief comment to explain what `zip` does, possibly with a link to the python doc. https://docs.python.org/3/library/functions.html#zip

arcade/examples/sound_demo.py
```diff
@@ -93,140 +88,43 @@ def setup(self):
...
-        self.button_sprites.append(button)
-
-        button = SoundButton(":resources:sounds/upgrade4.wav", pan=1, volume=volume)
-        button.center_x = SCREEN_WIDTH - BUTTON_SIZE
-        button.center_y = y
-        self.button_sprites.append(button)
+        upgrade_sound = ":resources:sounds/upgrade4.wav"
+        volume_variation = [0.1, 0.5, 1]
+        y_offset = [50, 0, -50]
+        button_x_spread = [
```
pushfoo, line 94: In its current form, this would be better put in a constant at the top of the file.

arcade/examples/sound_demo.py
```diff
@@ -1,45 +1,50 @@
...
 import arcade
 
 SCREEN_WIDTH = 800
 SCREEN_HEIGHT = 600
-SCREEN_TITLE = "Starting Template"
+SCREEN_TITLE = "Sound Panning Demo"
 BUTTON_SIZE = 30
 
+SOUND_PANNING = [-1.0, -0.5, 0.0, 0.5, 1.0]
+assert len(SOUND_PANNING) == 5
```
pushfoo, line 24: This might not make sense to keep. It would be better to move related constants here with a comment, or dynamically generate this.

pushfoo (approved):
@iansedano There are some minor stylistic issues where things look a little JS and less python, but this is otherwise good! Thank you!
My understanding is that we might merge it with some style fixes down the line since there may be a minor release getting made soon.

einarf:
Let's just merge this in and do some final tweaks.

PR #1229, opened 2022-06-02
Improve clarity & fix typos in build_mipmaps doc

pushfoo:
Built locally & viewed in browser.

### DIFF
--- a/arcade/gl/texture.py
+++ b/arcade/gl/texture.py
@@ -695,22 +695,21 @@ class Texture:
 
         Mipmaps are successively smaller versions of an original
         texture with special filtering applied. Using mipmaps allows
-        OpenGL to render scaled versions of original textures faster
-        and with far fewer scaling artifacts.
+        OpenGL to render scaled versions of original textures with fewer
+        scaling artifacts.
 
-        Mipmaps can be made for textures of any size. Each mipmap step
-        is 1/2 the size of the previous one (256 x 256, 128 x 128, 64 x
-        64, etc) down to a minimum size of 1 x 1.
+        Mipmaps can be made for textures of any size. Each mipmap
+        version halves the width and height of the previous one (e.g.
+        256 x 256, 128 x 128, 64 x 64, etc) down to a minimum of 1 x 1.
 
         .. note:: Mipmaps will only be used if a texture's filter is
                   configured with a mipmap-type minification::
 
-                    # Set up linear interpolating minification filter
-                    texture.filter = ctx.LINEAR_MIPMAP_LINEAR,\
- ctx.LINEAR
+                   # Set up linear interpolating minification filter
+                   texture.filter = ctx.LINEAR_MIPMAP_LINEAR, ctx.LINEAR
 
         :param int base: Level the mipmaps start at (usually 0)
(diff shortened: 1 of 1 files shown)


arcade/gl/texture.py
```diff
@@ -689,17 +689,25 @@ def write(self, data: Union[bytes, Buffer, array], level: int = 0, viewport=None
...
-        Note that mipmaps will only be used if the texture filter is
-        configured with a mipmap-type minification::
+        The default values usually work well.
 
-            # Set up linear interpolating minification filter
-            texture.filter = ctx.LINEAR_MIPMAP_LINEAR, ctx.LINEAR
+        Mipmaps are successively smaller versions of an original
+        texture with special filtering applied. Using mipmaps allows
+        OpenGL to render scaled versions of original textures faster
+        and with far fewer scaling artifacts.
```
einarf: It's definitely not faster, but it's pretty fast. Mipmap filtering will kick in when the texture is rendered in smaller sizes and OpenGL will read from multiple layers to produce a nicer interpolated result (fewer scaling artifacts is a decent way to say it)

arcade/gl/texture.py
```diff
@@ -689,17 +689,25 @@ def write(self, data: Union[bytes, Buffer, array], level: int = 0, viewport=None
...
+
+        Mipmaps can be made for textures of any size. Each mipmap step
+        is 1/2 the size of the previous one (256 x 256, 128 x 128, 64 x
```
einarf: They are actually 1/4th of the size. with and height divided by 2

arcade/gl/texture.py
```diff
@@ -689,17 +689,25 @@ def write(self, data: Union[bytes, Buffer, array], level: int = 0, viewport=None
...
+        64, etc) down to a minimum size of 1 x 1.
+
+        .. note:: Mipmaps will only be used if a texture's filter is
+                  configured with a mipmap-type minification::
+
+                    # Set up linear interpolating minification filter
+                    texture.filter = ctx.LINEAR_MIPMAP_LINEAR,\
+ ctx.LINEAR
```
einarf: You can reduce indentation here to avoid braking the line. 4 spaces is usually enough.

einarf (changes requested):

arcade/gl/texture.py
pushfoo: Good catch on the phrasing. :+1:

arcade/gl/texture.py
pushfoo: I broke it to keep line length at 72. In code blocks, breaking the line seems to cause the indent to get handled badly.

I can try to make this block long if it'll look nicer.

arcade/gl/texture.py
pushfoo: Some of the sources I was reading indicated that it can be used to provide speed increases, but that may have referred to other past implementations. I'll include this change in a bit. :+1:

einarf: Speed might depend on the situation. As a baseline mipmaps filtering will read more than one texel from the texture when reading from it. Probably just drop claims about speed 😄

PR #1230, opened 2022-06-03
Match pyglet version in release notes to the one in setup.py

pushfoo:
s/dev16/dev18/

Built doc & tested locally in browser.


pushfoo:
The CI failure is the same one that's present at the tip of development at time of writing: https://github.com/pythonarcade/arcade/runs/6719340390?check_suite_focus=true

PR #1246, opened 2022-06-19
Added UIDisappearingInfoBox construct and example.

JackAshwell11:


arcade/gui/constructs.py
```diff
@@ -72,3 +73,76 @@ def on_ok(self, event):
...
+
+
+class UIDisappearingInfoBox(UIMouseFilterMixin, UIAnchorLayout):
+    """
+    Represents a simple dialog box that pops up with a message and disappears after a
+    certain amount of time.
+
+    Parameters
+    ----------
+    width: float
```
pushfoo: This style of constructor documentation won't properly render when generating doc.

See this class for an example in the same file:
https://github.com/pythonarcade/arcade/blob/f41ae2666029427940e03d4f2d6d3df839a4efea/arcade/gui/constructs.py#L13-L19

doc/examples/gui_disappearing_info_box.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Disappearing Info Box
+==============
+
+For an introduction the GUI system, see :ref:`gui_concepts`.
+
+This example shows how to display an info box that will disappearing after a given
+time using the :class:`arcade.gui.UIDisappearingInfoBox` class.
+
+.. image:: gui_disappearing_info_box.png
```
pushfoo: 1. This image should probably be moved to after the heading.
2. You should use a literal include to list the source here. See the example linked below.

https://github.com/pythonarcade/arcade/blob/development/doc/examples/gui_scrollable_text.rst#L15-L17

pushfoo (changes requested):
I can't speak about conformity with UI design principles yet, but I've made some documentation suggestions.

arcade/gui/constructs.py
JackAshwell11: Yea, I just caught that error. I've made a fix. Kinda forgot to change it from my personal one.

JackAshwell11:
Adding a fade out effect as well as a fit parameter (where the box changes its size based on the text inside) would be nice, however, I'm not quite sure how to implement them right now.

PR #1247, opened 2022-06-19
UI message box enhance

JackAshwell11:


arcade/gui/constructs.py
```diff
@@ -25,50 +31,66 @@ def __init__(
         width: float,
         height: float,
         message_text: str,
-        buttons=("Ok",),
-        callback=None
-    ):
+        background: arcade.Texture = arcade.load_texture(
+            ":resources:gui_basic_assets/window/grey_panel.png"
+        ),
```
pushfoo: I don't think it's a good idea to load a texture inside this kind of module. It can lead to strange behavior and make debugging more difficult.

However, I liked the options you provided for setting backgrounds in the earlier drafts. I think we should support both colors and textures by providing arguments for both using the [same names that are on widgets as properties](https://github.com/pythonarcade/arcade/blob/118722555fbe27ad727cb800b72ff3d3560b8f94/arcade/gui/widgets/__init__.py#L199-L200):
```suggestion
        bg_color: arcade.Color = None,
        bg_texture: arcade.Texture = None,
```

In the constructor, you can then use both arguments:
```py
       message_box.with_background(bg_color, bg_texture)
```
Because the texture is applied after the color, using both lets you do things like this:
image
I used these arguments to do it:
```py
            bg_color=arcade.color.RED,
            bg_texture=arcade.load_texture("jewel.png"),
```
And this transparent 90s-inspired texture:
jewel

arcade/gui/constructs.py
```diff
@@ -25,50 +31,66 @@ def __init__(
...
-
-        self._bg_tex = arcade.load_texture(
-            ":resources:gui_basic_assets/window/grey_panel.png"
-        )
-        frame.with_background(texture=self._bg_tex)
+        # Store a few variables needed for this class
+        self._callback: Callable = callback  # type: ignore
+        self._disappear: bool = disappear
+        self._disappear_time_counter: float = disappear_time
+        padding_size = 10
```
pushfoo: This should probably be turned into another constructor kwarg with a default value.

arcade/gui/constructs.py
```diff
@@ -10,13 +13,16 @@
...
-    :param message_text:
-    :param buttons: List of strings, which are shown as buttons
-    :param callback: Callback function, will receive the text of the clicked button
+    :param width float: The width of the message box.
+    :param height float: The height of the message box.
+    :param message_text str: The message text to display.
+    :param buttons tuple[str, ...]: List of strings, which are shown as buttons.
+    :param disappear bool: Whether the box should disappear after a set time or not.
+    :param disappear_time float: The time before the box should disappear.
+    :param callback Callable: The callback function which will receive the text of the
```
pushfoo: Better than the previous drafts, but the syntax for type annotated params is the other way around: `:param float width:`

arcade/examples/gui_disappearing_info_box.py
```diff
@@ -0,0 +1,83 @@
...
+        # the info box which disappears after a certain amount of time
+        info_box = arcade.gui.UIMessageBox(
+            width=BOX_WIDTH,
+            height=BOX_HEIGHT,
+            message_text="This box will disappear in 3 seconds!",
+            background=arcade.Texture.create_filled(
+                "background color",
+                (int(BOX_WIDTH), int(BOX_HEIGHT)),
+                arcade.color.BABY_BLUE,
+            ),
```
pushfoo: We should use both `bg_color` and `bg_texture` as the arguments on this widget to comply with the currently unfinished standard:
https://github.com/pythonarcade/arcade/blob/118722555fbe27ad727cb800b72ff3d3560b8f94/arcade/gui/widgets/__init__.py#L199-L200
See my later comment for more info on why.

arcade/examples/gui_disappearing_info_box.py
```diff
@@ -0,0 +1,83 @@
+"""
+Example code showing how to use the DisappearingInfoBox
```
pushfoo:
```suggestion
Disappearing Info Box

An example of how to make info boxes vanish after some amount of time.
```
Having the user friendly title in the header is the convention for the doc and marginally helpful for spot checking examples from the terminal in vim.

arcade/gui/constructs.py
```diff
@@ -25,50 +31,66 @@ def __init__(
...
         message_text: str,
-        buttons=("Ok",),
-        callback=None
-    ):
+        background: arcade.Texture = arcade.load_texture(
+            ":resources:gui_basic_assets/window/grey_panel.png"
+        ),
+        buttons: Tuple[str, ...] = ("Ok",),
+        disappear: bool = False,
+        disappear_time: float = 3,
```
pushfoo, line 41: having `disappear` as an argument might be redundant if you have `None` as the default value for `disappear_Time`. I'm not sure about this yet. It might be worth leaving as-is and bringing it up with Eruvanos later.

pushfoo (changes requested):
This is much closer to ready!

There are a few things we should think about more, but we can bring them up with Eruvanos when he has time after cleaning up outstanding issues. One of them is having the timer behavior in the message box itself.

JackAshwell11:
I would also like to have that fade out effect and the fit box to text thing that I was talking about earlier. But for now its good until Eruvanos can chime in.

JackAshwell11:
The timer countdown code doesn't look that clean but it works as a proof of concept for now and can be redone later.

pushfoo (approved):
@eruvanos This mostly looks good enough to me from a stylistic standpoint. Can you give it a look through to make sure it's consistent with current design goals and concepts?

My remaining concerns are:
1. Is it a good idea to make disappearing a core feature of the message box?
2. Is `disappear` is redundant as an argument?

If we want to keep it disappearing, we could add support for `None` as a value for `disappear_time` and achieve the same effect without an extra parameter.

eruvanos:
I still need some time to think about this feature. Same goes for PR #1246

JackAshwell11:
I think we want to try and prefer this over #1246.

eruvanos:
Hi,
thank you for the idea, after a second review. I would like to keep the MessageBox like it is and not push it further.
The functionality will also be covered with the transitions, which come after the 3.0 release.

PR #1258, opened 2022-06-23
Improve rescaling & add rescale_xy_relative_to_point

### LINKED ISSUE #1187: Optimize Sprite.rescale_relative_to_point
Each line of `Sprite.rescale_relative_to_point` currently performs a sprite list update when it should do so only once at the end of the function:
https://github.com/pythonarcade/arcade/blob/a454fad5452d5d71343e53092527baa10796d193/arcade/sprite.py#L657-L661

https://github.com/pythonarcade/arcade/blob/a454fad5452d5d71343e53092527baa10796d193/arcade/sprite.py#L642-L656
https://github.com/pythonarcade/arcade/blob/a454fad5452d5d71343e53092527baa10796d193/arcade/sprite.py#L668-L678
https://github.com/pythonarcade/arcade/blob/a454fad5452d5d71343e53092527baa10796d193/arcade/sprite.py#L685-L695
### What would it help with?
1. #1164 , if done in Python rather than GL shaders
2. `SpriteList.rescale` would be much more efficient

pushfoo:
Closes #1187

The areas that need closest attention are:
1. Whether the change to the original `scale` setter is correct
2. Whether the optimized version looks good

~~If the core logic is good, I can add a `rescale_xy_relative_to_point` method as well before moving this out of draft form.~~

After discussion in chat, the current state of the optimized `rescale_relative_to_point` method and its new sibling are now marked ready for review as-is.

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -739,7 +739,7 @@ class Sprite:
             self._width = self._texture.width * self._scale[0]
             self._height = self._texture.height * self._scale[1]
 
-        # be lazy about math; only do it if we have to
+        # detect the edge case where the distance to multiply is 0
         position_changed = point != self._position
 
         # be lazy about math; only do it if we have to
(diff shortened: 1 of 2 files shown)


einarf:
As discussed the xy variant might be changed into the default one, Let's just merge for now.

Issue #1168, opened 2022-04-05
Automate broken link detection for documentation


pushfoo:
### What should be added/changed?

A way to detect 404 links in doc. 

Recap of [today's discord discussion after I brought the topic up](https://discord.com/channels/458662222697070613/458662458198982676/960683382684127232):
1. A [BeautifulSoup](https://beautiful-soup-4.readthedocs.io/en/latest/) script against generated files should be enough to get started
2. Selenium is overkill given we don't use much JS in the doc
3. 404 detection could be run as either a unit test or as part of the sphinx build

### What would it help with?
Finding broken links. We have multiple recent examples of broken links (#1165 , #1130 , #998), and there are likely more undetected. Upcoming doc reorganization will create even more, so we need a way to find them all.

pushfoo:
For now, developers can run `make linkcheck` locally from inside the doc directory. This sphinx functionality will attempt to contact outside sites in addition to checking internal links, and you may get rate limited. It will also be very slow as it decides to sleep sometimes.

For the future, I'll look into ways of specifying behavior such as conf.py overrides and CLI flags for running the linkcheck module on its own. If anyone beats me to it, comment on here and/or make a PR.

einarf:
So we should run `make linkcheck` in github actions?

pushfoo:
Not exactly. Default `make linkecheck` behavior does some things we don't want:

1. it detects all broken links everywhere, including the release notes
2.  if an external page is up but missing a specified anchor, linkcheck treats the link as broken

That's why I was asking about splitting release notes into different files on discord. It could help us choose where we link check more easily, ie only the the most recent releases. Old release notes have broken links to old page locations or deleted github accounts. We might not want to prioritize those, so they'd be noise if detected by CI.

PR #1260, opened 2022-06-28
Make example thumbnails clickable with sphinx workaround

### LINKED ISSUE #1181: Make example thumbnail images clickable parts of links
Example listing thumbnails

### Where is it located?
[doc/examples/index.rst](https://github.com/pythonarcade/arcade/blob/141c291c6ae78d7641c61894d0a5e5cb21df9e3f/doc/examples/index.rst)

Clickable or tappable thumbnails are the norm in the eyes of younger users who grew up with touch interfaces. Older users might not be put off, but the examples listing may seem broken to students at the moment.

pushfoo:
Closes #1181 the ugly way as a workaround for https://github.com/sphinx-doc/sphinx/issues/4351. Built & tested locally, the thumbnails work and nothing seems to break.

### DIFF
--- a/doc/example_code/how_to_examples/index.rst
+++ b/doc/example_code/how_to_examples/index.rst
@@ -738,6 +738,7 @@ Concept Games
 
 .. figure:: https://raw.githubusercontent.com/pythonarcade/asteroids/main/screenshot.png
    :figwidth: 170px
+   :target: https://github.com/pythonarcade/asteroids
 
    `Asteroids with Shaders <https://github.com/pythonarcade/asteroids>`_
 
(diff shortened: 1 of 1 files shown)


pvcraven:
The reason I haven't done this, is that it makes breaking links much easier. I think adding in a Python script to yell if a link is broken might be needed.

pushfoo:
That's already filed as a separate issue: #1168. [As I commented on the issue](https://github.com/pythonarcade/arcade/issues/1168#issuecomment-1092367516), it's more complicated to solve elegantly than initially expected.

PR #1261, opened 2022-06-29
Bump pyglet version to 2.0.dev20

pushfoo:
This release fixes issues a user reported on a 32 bit python version. There appears to be no difference in pytest output between dev19 and dev20 versions.


einarf:
You beat me to it 😄

PR #1265, opened 2022-07-06
Optimization & bug fixes for PerfGraph

pushfoo:
**tl;dr ~7% faster, fixes bugs, backward compatible**

These changes seem to run at least 7% faster and with far more FPS stability than the current development branch. Most of the improvement is likely from the removal of `arcade.draw_text`.

Major changes:
* Switch to `arcade.Text` instances instead of `arcade.draw_text` calls
* Use batched mutable `pyglet.graphics.Line` instances for non-data lines
* Convert multiple attributes into backward-compatible properties that update mutable pyglet shapes
* Fix rendering bugs when PerfGraph instances are scaled

## Performance comparison:
These screen shots are toward the maximum improvement (52 FPS is the minimum on development for my system). 7% improvement seems to be the average. Getting exact numbers doesn't seem worth the time at the moment, but the improved FPS stability of the PR version is worth noting.

Before:
image
After:
image

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -135,6 +135,7 @@ from .utils import rand_on_line
 from .utils import rand_vec_magnitude
 from .utils import rand_vec_spread_deg
 from .utils import generate_uuid_from_kwargs
+from .utils import opacity_of_color
 
 from .geometry_generic import get_distance
 from .geometry_generic import rotate_point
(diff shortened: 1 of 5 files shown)


einarf:
Let's just merge this. It will be be first feature using pyglet batches in arcade 👍 (Since this is now possible in 2.7)

PR #1266, opened 2022-07-06
Added lerp_angle to utils.

shadow7412:


shadow7412:
Where in the pipelines do the tests run?

einarf:
Right now we mainly run code inspection with flake8 and mypy in github actions. Just run `pytest` locally for now.

`pip install -e .[dev]` to get all the development dependencies.

Test should ideally run in github actions, but we need to do some cleanup for that to work properly.

arcade/utils.py
```diff
@@ -16,6 +16,17 @@ def lerp_vec(v1: Vector, v2: Vector, u: float) -> Vector:
     )
 
 
+def lerp_angle(start_angle: float, end_angle: float, u: float):
+    """Linearly interpolate between two angles in degrees, following the shortest path."""
+    while start_angle - end_angle > 180:
+        end_angle += 360
+
+    while start_angle - end_angle < -180:
+        end_angle -= 360
```
einarf, line 28: Is there a better/faster way doig this as in not using a while loop?

arcade/utils.py
```diff
@@ -16,6 +16,17 @@ def lerp_vec(v1: Vector, v2: Vector, u: float) -> Vector:
     )
 
 
+def lerp_angle(start_angle: float, end_angle: float, u: float):
+    """Linearly interpolate between two angles in degrees, following the shortest path."""
```
einarf, line 20: Considering this method specifically lerps using the shortest path in a circle, the method should probably be named to relfect this? It could cause confusion.

shadow7412: I thought it was pretty clear, as if you were lerping an angle, this is pretty likely the behaviour you'd want. Happy to take suggestions though.

arcade/utils.py
shadow7412: Probably - but I figured it was ok since this came from the existing codebase anyway.

Open to ideas, but the only one that comes to my mind is using some form of division, which is presumably going to be less performant than this solution (unless you have a excessively large angle where this actually needs to execute multiple times)

Most of the time, this will effectively be an if check.

arcade/utils.py
einarf: What about `lerp_angle_shortest`? I just fear people will use this for angles in general.

shadow7412: I'm happy to change it to that - though I'm not sure why people using this for angles in general would necessarily be a bad thing...

arcade/utils.py
einarf: A suggetions is to make a variant with division and use `timeit` to compare. When putting a new method in a library we don't really know how people will end up using it. For example: Users might accumulate rotation values over time.

shadow7412:
> `pip install -e .[dev]` to get all the development dependencies.

That failed for me - but quoting `".[dev]"` did the trick. I guess shell was trying to do substitutions.

For what it's worth, these two tests (unrelated to my changes) are failing for me locally.
```
FAILED tests/unit2/test_opengl_query.py::test_create - assert 480000 >= (3840 * 2160)
FAILED tests/unit2/test_sprite_collision.py::test_sprite_collides_with_point - assert True is False
```

arcade/utils.py
einarf: Then let's just keep it as is.

arcade/utils.py
shadow7412: Ok, I'm modding both angles as they come in now. This forces them to be in a respectable range before the while loop runs.

I did also try checking if the angle was outside of a reasonable range before bothering to mod - but I don't think doing so gave enough benefit to be worth it, or rather, somehow it was worse.

Code:
```
print("Normal:", timeit.timeit(lambda: arcade.lerp_angle_shortest(50, 180, 0.5), number=50000))
print("Mod:", timeit.timeit(lambda: arcade.lerp_angle_shortest_mod(50, 180, 0.5), number=50000))
print("Check Mod:", timeit.timeit(lambda: arcade.lerp_angle_shortest_check_mod(50, 180, 0.5), number=50000))
print()
print("Normal (large angle):", timeit.timeit(lambda: arcade.lerp_angle_shortest(50, 50 + 360*10000, 0.5), number=50000))
print("Mod (large angle):", timeit.timeit(lambda: arcade.lerp_angle_shortest_mod(50, 50 + 360*10000, 0.5), number=50000))
print("Check Mod (large angle):", timeit.timeit(lambda: arcade.lerp_angle_shortest_check_mod(50, 50 + 360*10000, 0.5), number=50000))
```

>Normal: 0.01713432699943951
Mod: 0.01952442699985113
Check Mod: 0.02082333600083075
>
>Normal (large angle): 29.170401744999253
Mod (large angle): 0.019441309001194895
Check Mod (large angle): 0.020343105999927502

einarf:
I think this so looking good now 👍. Will merge this unless you have anything else to add.

For the two tests: Think I fixed the query test. For the collision tests we know that the test fails if you don't have shapely installed. It's a known issue in the dev branch, so don't worry about it 😄

shadow7412:
> It's a known issue in the dev branch, so don't worry about it

Cool - as long as it's known.

> Will merge this unless you have anything else to add.

I don't think so. Merge away :) What's your rough release cycle? I'll need to notice the next update so I can take out my local version of this method :P

einarf:
We don't really have a deadline on 2.7, but I would believe we could be ready late summer.
You could set up the development branch in arcade as the dependecy for your project?

shadow7412:
Nah, there's no rush at this stage. My local version is literally the same anyway :P

PR #1269, opened 2022-07-08
Expand SpriteList.extend typing to include Iterables of Sprites

pushfoo:
It seems reasonable to support this, and it's easier than bringing SpriteList into full compliance with the MutableSequence ABC. It also seems more elegant to use generators of various sorts to create sprites instead of creating a second list, then copying from it to the SpriteList's internal one.


einarf:
Much better.

PR #1278, opened 2022-07-15
Fix capitalization in title

eschan145:


pvcraven:
This is a correct capitalization, but oddly enough I kind of prefer the incorrect one. Thanks for the PR, but I think we'll keep it as-is.

PR #1280, opened 2022-07-20
Nine patch renderer without GUI abstraction.

DragonMoffon:
Keeping this in a branch until we create the abstraction may be a good idea. We may also want to add the ability to change the texture, start, or end values before merging. It is also not typed because I know one of the supported Python versions is funky about that, and I don't know what to do for it. The renderer is heavily commented but is missing a doc string.


eruvanos:
@einarf could you have a look on the shader part. We could merge it and I can start to build a GUI integration.
Maybe we need a method on the Surface to draw the 9patch within the GUI buffer?

eruvanos:
Thank you very much for the contribution. We moved your changes with a few adjustments into a new branch.
We will do some refactoring and adjustments and merge it into development after that.

https://github.com/pythonarcade/arcade/tree/gui/ninepatch

PR #1289, opened 2022-07-31
Draw Rate Addition and Update Event Cleanup

Cleptomania:
## Overview

Throughout Arcade 2.6, it has been not possible to increase the frequency of dispatching the `on_draw` function from a target of 60 FPS. This is because of a change in Pyglet 2.0 from the legacy versions. Previously in Pyglet, the `on_draw` event was dispatched whenever any other event was. In Pyglet 2.0 this behavior changed such that it is dispatched at a set interval using the clock, like we do for the `on_update` event. The rate for this dispatch is set as an argument to the `pyglet.app.run()` function, and defaults to `1 / 60`. 

The first thing this PR does is add the ability to set a `draw_rate` parameter in the Window's constructor, and that value will then be passed down to `pyglet.app.run()`. This alone doesn't make any breaking change, the value defaults to `1 / 60` like it always has, you now just have the ability to set it. The `update_rate` parameter has also been changed to no longer be able to be None. If it is None an error will be introduced. There is almost certainly no usage of making this value None, but it was technically possible before. You don't actually have to set this, it will default to 1 / 60 same as before, it just can't be explicitly set to None anymore.

The next thing in this PR is the removal of the `update` event. This has been discussed many times previously, but with a shift to more things being able to separately control draw rate vs update rate, it is more important that we stress the available of the `delta_time` value that you get with the `on_update` function. This has been the sort of soft recommended one for quite some time, but with this the old `update` function is explicitly removed and you are forced to use the newer `on_update` one. This applies to the Window, View, and Section classes.

Another thing here is the removal of the `arcade.run()` function. There has existed a `Window.run()` method for some time now, which has just been an alias to the main `arcade.run()`. This PR removes the direct call, and you now have to call it as a method on the Window class. This mostly just serves to unify the API around the object-oriented approach and using less static functions, and makes the `run()` function not be dependent on a singleton global Window value since the `run()` function now needs to know the `draw_rate` value from the Window. There are other solutions to this that don't introduce a breaking change, but with already removing the `update` function, this is a fairly light change that will introduce at most a one line change for any applications and create a more unified API so it is probably worth it.

## Task Breakdown

- [x] Implement draw_rate
- [x] Remove `update` events from Window, View, and Section classes
- [ ] Add delta time value to update function of various classes which may need it(Sprite, Particle, Emitter, Camera, etc). Even if we don't make use of it internally, the function should allow passing it, for future use or for anyone who subclasses and overrides so that it can have a matching signature. Some of these may also need setup to accept kwargs or something as well, but that is probably out of scope of this PR. 
- [x] Update all documentation/example references to the old `update` event
- [ ] Update release notes, and specifically add a section for 2.7 Breaking Changes(this may be extended by other work before 2.7 is released)


Cleptomania:
We are no longer removing the `arcade.run()` function. This will still function exactly as before, it will grab the draw rate from the global Window singleton.

PR #1292, opened 2022-08-02
Fix type hint in `arcade.Window.__init__()`

### LINKED ISSUE #1291: Missing `Optional[]` in `arcade.Window.__init__()`
## Bug Report
The type hint for the `screen` parameter in `arcade.Window.__init__()` is missing an `Optional[]` hint, and the default value is `None`. This causes type checkers like Pyright to underline `super().__init__()` in a subclass of `arcade.Window` with default arguments as an error. If possible I would like to work on this issue myself.

https://github.com/pythonarcade/arcade/blob/71807836c5359a0d0feddd7ffb276c03abd42e49/arcade/application.py#L85-L89

image

## System Info
Arcade 2.6.15

vendor: Intel
renderer: Intel(R) UHD Graphics
version: (4, 6)
python: 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
platform: win32

### Actual behavior:
Pyright underlines the `__init__()` call with a [type error](https://github.com/microsoft/pylance-release/blob/main/DIAGNOSTIC_SEVERITY_RULES.md#diagnostic-severity-rules).

### Expected behavior:
There should be no type errors for the default arguments, especially when other parameters use the `Optional[]` hint.

In VS Code, enable type checking by setting the `Python > Analysis > Type Checking Mode` option to `basic` or `strict`.
Paste the following code into a Python file.
```python
import arcade


class MyGame(arcade.Window):
    def __init__(self):
        super().__init__()
```

## Enhancement request:
Fix the typing issue.

Add `Optional[]` around the `pyglet.canvas.Screen` type hint.
```python
screen: pyglet.canvas.Screen = None,
```
Changed to:
```python
screen: Optional[pyglet.canvas.Screen] = None,
```

### What would it help with?
It would remove the Pyright error.

lemonyte:
Added `Optional[]` to the `screen` parameter in `arcade.Window.__init__()` because the default value is `None`. Fixes #1291.

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -85,7 +85,7 @@ class Window(pyglet.window.Window):
         update_rate: Optional[float] = 1 / 60,
         antialiasing: bool = True,
         gl_version: Tuple[int, int] = (3, 3),
-        screen: pyglet.canvas.Screen = None,
+        screen: Optional[pyglet.canvas.Screen] = None,
         style: Optional[str] = pyglet.window.Window.WINDOW_STYLE_DEFAULT,
         visible: bool = True,
         vsync: bool = False,


einarf:
Thanks. There seems to be some other issues with this screen parameter. Missing docstring and it's not always used in the initializer.

PR #1294, opened 2022-08-03
Integer UI

DragonMoffon:
While working on the Hit Box tool I found that the UI elements often did not have integer position causing 1-2 pixel flickering and unpleasant artefacts. I have currently just updated gui.Rect to always floor values to int.


eruvanos:
Hi,
thanks for the idea and your contribution/work.

While it makes sense to only use integers within the GUI, this would reduce the reuse of Rect.

Furthermore Rect provides a lot of type hints, which should be adjusted. It might be a better idea, to add a method, which would return an ensured int tuple or adjust the type hints and use Rect in a consistent way with integers.

I am up for discussions.

DragonMoffon:
Thats a very fair point. I believe keeping the rects ability to store floating points would be useful. The two options i can think off would be to give the rect a series properties which return integer values, or  returning a correctly typed tuple/new rect. I am hesitant with the second option as it requires creating another rect every time the rect is updated (which i am aware already happens but I mean twice). I dont have enough experience with named tuples to know if the creation cost is cheaper than casting to int on property call.

PR #1304, opened 2022-08-19
Sprite update hitbox docs

eschan145:


einarf:
Remove the claims about caching. This is no longer true. There is also a lingering whitespace in the PR?

arcade/sprite.py
```diff
@@ -1084,14 +1084,15 @@ def draw(self, *, filter=None, pixelated=None, blend_function=None) -> None:
 
     def draw_hit_box(self, color: Color = BLACK, line_thickness: float = 1) -> None:
         """
-        Draw a sprite's hit-box.
+        Draw a sprite's hit box.
 
-        The 'hit box' drawing is cached, so if you change the color/line thickness
-        later, it won't take.
+        The hit box drawing is cached, so if you change the color or line thickness
+        later, it won't take more time.
```
pvcraven, line 1090: The original phrase was to let the user know, if they later changed the color from red to blue, the caching would prevent the update from happening at all. Not that it wouldn't take more time. And from einarf's comment, I don't think this is the case so this can just be removed.

arcade/sprite.py
```diff
@@ -1084,14 +1084,15 @@ def draw(self, *, filter=None, pixelated=None, blend_function=None) -> None:
...
 
-        :param color: Color of box
-        :param line_thickness: How thick the box should be
+        :param color: Color of hitbox outline
+        :param line_thickness: How thick the hitbox outline should be
         """
+
```
pvcraven, line 1095: This new blank line contains whitespace and is preventing checks from passing.

pvcraven (changes requested):

PR #1306, opened 2022-08-20
Add documentation of profiling

eschan145:


pvcraven:
Looks like there are errors in the docs build.

doc/programming_guide/performance_tips.rst
```diff
@@ -90,4 +90,44 @@ Text drawing performance
...
+removing stuff and seeing if your game runs faster, you can use a profiler. A
+profiler is for measuring parts of your code and seeing how fast they run. This
+is useful so you can know what is taking up most of the time and causing your
+game to slow down.
+
+There are many great profilers out there, but one we recommend is called
+`cProfile`. You can make it test some code like this:
+
+.. code:: python
+    import cProfile
```
pushfoo:
```suggestion
.. code-block:: python

    import cProfile
```
Try this. The newlines are important in sphinx, and using the full `code-block` directive gets picked up better by PyCharm.

doc/programming_guide/performance_tips.rst
```diff
@@ -90,4 +90,44 @@ Text drawing performance
...
+    import time
+
+    cProfile.run("time.sleep(5)")
+
+It would return something like this:
+
+.. code:: python
+    4 function calls in 10.009 seconds
```
pushfoo, line 116:
```suggestion
.. code-block:: python

    4 function calls in 10.009 seconds
```
Same change as the last comment.

pushfoo (changes requested):
Try making the changes I suggested to fix the errors specific to your PR. They seem to work locally. There may be some other build warnings remaining afterward, but I think those are from other commits.

doc/programming_guide/performance_tips.rst
pvcraven: This is also not python code, and needs to be formatted plain-text. I think that's 'text'? I forget.

einarf:
We do actually have a profiler wrapper class in `experimental` as well.

doc/programming_guide/performance_tips.rst
pushfoo: I think you commented on the wrong block, but your point is valid.

doc/programming_guide/performance_tips.rst
pushfoo: 1. `cProfile.run` prints, it doesn't return values
2. Paul C is correct that this isn't python code, so you should use [a literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) as below:
```suggestion
It would print something like this::

    4 function calls in 10.009 seconds
```

doc/programming_guide/performance_tips.rst
```diff
@@ -90,4 +90,44 @@ Text drawing performance
...
+        Ordered by: standard name
+
+        ncalls  tottime  percall  cumtime  percall filename:lineno(function)
+             1    0.000    0.000   10.009   10.009 <string>:1(<module>)
+             1    0.000    0.000   10.009   10.009 {built-in method builtins.exec}
+             1   10.009   10.009   10.009   10.009 {built-in method time.sleep}
+             1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects)
+
+ncalls : Number of calls made
+tottime : Total time taken by the function. Time made in calls to sub-functions are excluded.
```
pushfoo, line 127: What do you mean by "Time made in calls"? This is unclear.

pushfoo (changes requested):
I added some further comments. Hopefully fixing the second block stops formatting issues. You should also consider formatting lines 126-129 as a table.

PR #1307, opened 2022-08-20
Reformat to Pep 8

eschan145:
The line width is 84 characters, and should be 80.

### DIFF
--- a/arcade/arcade_types.py
+++ b/arcade/arcade_types.py
@@ -20,7 +20,8 @@ NamedPoint = namedtuple("NamedPoint", ["x", "y"])
 
 Sequence[int]
 PointList = Sequence[Point]
-Rect = Union[Tuple[float, float, float, float], List[float]]  # x, y, width, height
+Rect = Union[Tuple[float, float, float, float],
+             List[float]]  # x, y, width, height
 RectList = Union[Tuple[Rect, ...], List[Rect]]
 
 


Cleptomania:
I believe we are typically using an 88 character line length. Much of the code is typically formatted with black now(though we haven’t gotten to using it across the entire codebase and checking it in CI yet)

einarf:
This is project dependent. A line length of 80 is just a suggesion. A more modern apporach is 100ish. We do allow lines lengths up to 120 in arcade: https://github.com/pythonarcade/arcade/blob/288f1a1e5c50e90f52cfaf9bde564e413f319649/setup.cfg#L11

eschan145:
I always used 80 chars for maximum line lengths.

einarf:
https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

Even Raymond Hettinger (python core developers) have suggested that 95+ is usually better since 80 is a bit short. In most cases you make the code less readable when breaking lines with an 80 char limit. It's the readability that counts.

The 80 char limit is really an arcane rule simply because terminals back in the days were 80 chars wide.

einarf:
Just closing this for now. If we are changing the maximum line length it should include changes to `setup.cfg` and fixing all lines violating the new rule by running flake8. I personally don't think there is anything to gain from doing this at this point.

This could definitely be discussed on discord if you think it's important. `setup.cfg` also contains other options ignoring some pep8 rules that we consider too strict. I don't think many projects follow pep8 by the book and that was not the intention of pep8 in the first place.

PR #1310, opened 2022-08-21
Remove remaining references to pillow text functions

pushfoo:
Remove leftover references to pillow text functions not already removed by 72790c2.

Doc built & tested locally.


einarf:
Thanks! The question is if we should have a new `create_text` that crates a pillow image using pyglet text. The downside is that this is a lot slower than what the text sprite method is doing currently (allocates space in atlas and renders into it).

pushfoo:
I think the answer will depend on how much backward compatibility we want to have. I don't have a good big picture view of what the API was like in 2.6.X and what we've changed in `development` branch.

pushfoo:
I looked a bit closer, and we seem to still have a `create_text_sprite` method despite there being no tests for it. If my understanding is correct, this is a very useful method because drawing a sprite is 1 quad while Text objects draw a quad per glyph.

If backward compatibility is a priority, maybe we can use `create_text_sprite`'s output and return `text_sprite.texture.image`? I think reference counting should allow this to work even if the original sprite is destroyed. However, if we want to cut edge case library features, we can describe this technique somewhere in the programming guide.

PR #1319, opened 2022-09-06
Clarify SpriteSolidColor and SpriteCircle docstrings

pushfoo:
Built & tested locally.

This PR clarifies the `SpriteSolidColor` and `SpriteCircle` docstrings by making the following changes:
* Use more precise language in describing the shapes (circle instead of ellipse)
* Succinctly explain the user impact of the texture generation & caching behavior
* Explain the direction of the gradient for `SpriteCircle`
* Sync the types of arguments and their annotation in the docstrings

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -1416,12 +1416,19 @@ def load_animated_gif(resource_name) -> AnimatedTimeBasedSprite:
 
 class SpriteSolidColor(Sprite):
     """
-    This sprite is just a rectangular sprite of one solid color. No need to
-    use an image file.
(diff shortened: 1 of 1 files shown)


einarf:
I'll just merge this. We can update when we change it to only make white textures.

pushfoo:
@einarf I think we might want to set color to a keyword arg to get the best of both worlds.

einarf:
Should not be a problem. Default white then?

pushfoo:
Yes, for both of them. I'd also like to take consider changing some of the related underlying functionality after I finish my current experiment and reflect on it for a bit.

Issue #1076, opened 2022-01-31
arcade.Camera not displaying correctly when position is not 0,0 (left, bottom)


alejcas:
I would like the `arcade.Camera` left, bottom positions to be configurable. Now 0,0 its hardcoded.
That's part one.

As I talked on Discord with Clepto theoretically we just need to change:

```python
class NewCamera(arcade.Camera):

    def __init__(self, left, bottom, viewport_width: int = 0, viewport_height: int = 0):
        self.left = left
        self.bottom = bottom
        super().__init__(viewport_width, viewport_height)

    def set_projection(self):
        """
        Update the projection matrix of the camera. This creates an orthogonal
        projection based on the viewport size of the camera.
        """
        self.projection_matrix = Mat4.orthogonal_projection(
            self.scale * self.left,
            self.scale * self.viewport_width,
            self.scale * self.bottom,
            self.scale * self.viewport_height,
            self.near,
            self.far
        )

    def use(self):
        """
        Select this camera for use. Do this right before you draw.
        """
        self._window.current_camera = self
        self.update()
        self._window.ctx.viewport = int(self.left), int(self.bottom), int(self.viewport_width), int(self.viewport_height)
        self._window.ctx.projection_2d_matrix = self.combined_matrix
```

But when doing so, If I position the camera at say (600, 0) I got two issues:
1. The position of the sprites get inverted (right - left)
2. The sprites are streched

Below you a have an image showing this:
One camera set at (0,0) (left one). Sprites are 32x32 and look normal. Also, sprite at left, bottom (0,0) is yellow so you can see the effect when setting the camera away from (0,0).
However, at the right you have another camera set at (600,0). You can see that the yellow sprite is at the right (but it's position is still at left, bottom (0,0) and you can see how the 32x32 sprites are distorded.

![image](https://user-images.githubusercontent.com/7224106/151803923-b9896538-0927-4e06-84f6-98aa019821fa.png)

alejcas:
Any chances this could be somehow fixed? I've no clue why this happens so I can't fix it.

Also this is pretty needed if you want for example a minimap camera and the minimap is not at left, bottom = 0,0.

pvcraven:
I don't understand what the use-case is, or what is being attempted here.

alejcas:
There are two issues:

1) arcade.Camera is set to begin at 0,0 left bottom. So if you want the camera set at the top right corner you can’t
[...]

PR #1320, opened 2022-09-08
Camera new implementations

alejcas:
### **This pull request adds two new cameras to arcade.**

First a `BaseCamera` that is just a camera that allows to set the viewport, projection and can move around. Just that.

Second a  `AdvanceCamera` that implements a camera with more functionality.

The arcade current `Camera` remains the same. In the future a deprecation warning should be implemented to drop this camera in favor of the `AdvanceCamera` (or whatever other solution you may find appropriate). 

The changes (differences also) are:
- Both cameras implement viewport and projection independent from each other
- Both cameras allow to set the camera viewport on whatever position other than left, bottom = (0, 0)
- Both cameras allow to set the camera projection on whatever position other than left, bottom = (0, 0)
- Both cameras only calculate the matrices when it's needed. arcade.Camera calculates the matrices in every frame.
- Both cameras allow to set important properties right from the init.
- AdvanceCamera has a zoom (renamed from the more abstract keyword "scale")
- AdvanceCamera has a `center` method that just centers (moves to the center) the camera on a certain vector
- AdvanceCamera has a  `get_map_coordinates` method that returns the "map" (or whatever you call what the camera is navigating on) coordinates from the camera viewport coordinates relative to the screen.

AdvanceCamera also has rotation but that was already implemented on the development branch.

What's missing:
- Although I test them, further **testing** on the cameras is needed. Specially when using more than one camera and when cameras are positioned far away of left, bottom (0, 0)
-  **Collision detection** when using rotation or zoom doesn't work. @einarf told me on discord that this should be possible to overcome by implementing the arcade collision methods on the camera and translating the coordinates to check with the projection and view matrices. I have not achieved a success here. Also seems trivial to achieve `get_sprites_at_point` but I can image how to transform coordinates when checking a spritelist against another spritelist for example.

 
I hope this PR helps and gets accepted.

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -99,7 +99,7 @@ from .window_commands import set_window
 from .window_commands import start_render
 from .window_commands import unschedule
 
-from .camera import Camera
+from .camera import Camera, BaseCamera, AdvanceCamera
 from .sections import Section, SectionManager
 
 from .application import MOUSE_BUTTON_LEFT
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -6,7 +6,7 @@ import logging
 import os
 import sys
 import time
-from typing import Tuple, Optional
+from typing import Tuple, Optional, Union
 
 import pyglet
 
@@ -28,6 +28,8 @@ MOUSE_BUTTON_MIDDLE = 2
 MOUSE_BUTTON_RIGHT = 4
 
 _window: 'Window'
+
+ArcadeCameras = Union[arcade.Camera, arcade.BaseCamera, arcade.AdvanceCamera]  # type alias
 
 
 def get_screens():
@@ -168,7 +170,7 @@ class Window(pyglet.window.Window):
         set_window(self)
 
         self._current_view: Optional[View] = None
-        self.current_camera: Optional[arcade.Camera] = None
+        self.current_camera: Optional[ArcadeCameras] = None
         self.textbox_time = 0.0
         self.key: Optional[int] = None
         self.flip_count: int = 0
--- a/arcade/camera.py
+++ b/arcade/camera.py
@@ -2,11 +2,150 @@
 Camera class
 """
 import math
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 
 from pyglet.math import Mat4, Vec2, Vec3
 
 import arcade
+
+# type aliases
+FourIntTuple = Tuple[int, int, int, int]
+FourFloatTuple = Tuple[float, float, float, float]
+
+
+class BaseCamera:
+    """
+    A simple camera that allows to change the viewport, the projection and can move around.
+    That's it.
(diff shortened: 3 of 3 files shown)


alejcas:
The PR also changes:
-  application.py to change the type definition of window.current_camera
- __init__ to add the imports for BaseCamera and AdvanceCamera

einarf:
#1076

alejcas:
Still can't figure out  what's failing in collision detection.

This is a preliminary method based on @einarf comments on discord:

```python

# method from arcade.AdvanceCamera class

def get_sprites_at_point(self, point: tuple, sprite_list: "arcade.SpriteList"):
    # this is needed when the camera applies rotation or zoom.
    # default arcade.get_sprites_at_point will fail to get the sprites

    x, y = point
    vector = Vec4(x, y, 0, 1)

    # self._view_matrix takes the zoom into account (not rotation)
    # in theory this should translate the point to the correct position when zoom is applied
    
    translated_point = self._view_matrix @ vector
    return arcade.get_sprites_at_point(translated_point, sprite_list)
```

It is not working and this is what I think about (although I can be very wrong):
-  `camera._view_matrix` takes zoom into account but not rotation. Rotation is set inside `camera._rotation_matrix` (this is what is passed then to `window.view_matrix_2d`). But rotation matrix does not take zoom into account... so to make collision rotation work both rotation and zoom matrices should be taken into account.
- the sprite_list should also somehow translate it's coordinates?

pvcraven:
Creating a new camera and deprecating an old one, I think needs a strong reason for the switch. All examples and docs need to be updated, and users will need to update their code. Are the features of the new camera compelling enough for this? 

I'm not sure yet that I see those reasons.

alejcas:
The fact is that the current arcade camera has some things that must be changed. Mainly the fact that viewport and projection are interrelated in some ways that are incorrect. Or the fact that camera viewport and projection is hardcoded at (0,0) left bottom.

Why I didn’t just change/fix/add features to the old camera? Because to do that I needed to change the constructor method signature + change some things that will break old code. And that is a mess for current users.

As a user of arcade I find the current implementation of the Camera lacking some big big things as I mention on the changelog of this PR. And that was the rationale…

Deprecating the old camera was just a suggestion, you don’t have to do it as I tried to say with “or whatever other solution you may find appropriate”.

einarf:
Sorry for not reviewing this yet. Something definitely needs to be fixed with the camera. We are completely stuck with the current one unless we do breaking changes.

alejcas:
Hi @einarf Can I do something with this to get the functionality merged (in it's current state or other approach)?

einarf:
Let's just merge this in for now so we get some progress. There are probably some things to iron out before 2.7 release. Things I don't have the brain power and time to evaluate right now 😄 (Texture revamp stuff takes up 100% of my energy atm)

PR #1322, opened 2022-09-17
Update pymunk_physics_engine.py

MostTornBrain:
Per pymunk documentation, the pre handler is supposed to return a bool:

"Return false from the callback to make pymunk ignore the collision this step or true to process it normally. Additionally, you may override collision values using Arbiter.friction, Arbiter.elasticity or Arbiter.surfaceVelocity to provide custom friction, elasticity, or surface velocity values. See Arbiter for more info."

This addresses the error which occurs if you define a pre_handler function for collisions without this fix.:
"UserWarning: Function '_f3' should return a bool to indicate if the collision should be processed or not when used as 'begin' or 'pre_solve' collision callback.
  def _f3(arbiter, space, data):"

### DIFF
--- a/arcade/pymunk_physics_engine.py
+++ b/arcade/pymunk_physics_engine.py
@@ -308,7 +308,7 @@ class PymunkPhysicsEngine:
 
         def _f3(arbiter, space, data):
             sprite_a, sprite_b = self.get_sprites_from_arbiter(arbiter)
-            pre_handler(sprite_a, sprite_b, arbiter, space, data)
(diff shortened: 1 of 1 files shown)


pushfoo:
This might be worth backporting to 2.6.X if we plan to do another release in that branch

pushfoo (approved):
It looks good to me & works locally. I can't merge it myself, but einarf or Clepto might get to it later tonight.

einarf:
@pvcraven can you look at this?

Issue #1325, opened 2022-09-19
General Codebase Refactor


BrettskiPy:
### What should be added/changed?

I've scanned most of the Arcade repo with https://sourcery.ai/ (pro version) and I would like to implement high quality refactoring solutions without changing code behavior. 

### What would it help with?

Improving the overall code quality of the Arcade engine and ensuring modern Python practices are upheld.

PR #1326, opened 2022-09-19
General Code Refactor Using Sourcery

BrettskiPy:
Issue #1325 

I've noticed a few areas of the Arcade codebase that can be improved with the assistance of AI refactoring tools [Sourcery](https://docs.sourcery.ai/Welcome/). I spent some time to pick out a subsection of the ways I believe the code can be improved. 

If any/all of these "suggestions" are implemented, I'll be happy. The AI refactoring selections I've chosen are my bias to what "cleaner code" is. There are plenty more suggestions, but I figured I would start with this to see if this is something that is received positively. 😉 

Thank you!

### DIFF
--- a/arcade/buffered_draw_commands.py
+++ b/arcade/buffered_draw_commands.py
@@ -200,7 +200,7 @@ def create_lines_with_colors(point_list: PointList,
 
     if line_width == 1:
         return create_line_generic_with_colors(point_list, color_list, gl.GL_LINES, line_width)
-    
+
     triangle_point_list: List[Point] = []
     new_color_list: List[Color] = []
     for i in range(1, len(point_list), 2):
--- a/arcade/buffered_draw_commands.py
+++ b/arcade/buffered_draw_commands.py
@@ -66,8 +66,7 @@ def create_line(start_x: float, start_y: float, end_x: float, end_y: float,
     points = get_points_for_thick_line(start_x, start_y, end_x, end_y, line_width)
     color_list = [color, color, color, color]
     triangle_point_list = points[1], points[0], points[2], points[3]
-    shape = create_triangles_filled_with_colors(triangle_point_list, color_list)
-    return shape
+    return create_triangles_filled_with_colors(triangle_point_list, color_list)
 
 
 def create_line_generic_with_colors(point_list: PointList,
@@ -128,13 +127,7 @@ def create_line_generic(point_list: PointList,
     just changing the OpenGL type for the line drawing.
     """
     colors = [get_four_byte_color(color)] * len(point_list)
-    shape = create_line_generic_with_colors(
-        point_list,
-        colors,
-        shape_mode,
-        line_width)
-
-    return shape
+    return create_line_generic_with_colors(point_list, colors, shape_mode, line_width)
 
 
 def create_line_strip(point_list: PointList,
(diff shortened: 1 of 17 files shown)


arcade/buffered_draw_commands.py
```diff
@@ -406,15 +391,9 @@ def create_rectangle(center_x: float, center_y: float, width: float,
             data = point_list_2
 
         border_width = 1
-        shape_mode = gl.GL_TRIANGLE_STRIP
 
-        # _generic_draw_line_strip(point_list, color, gl.GL_TRIANGLE_STRIP)
```
BrettskiPy, line 411: Do we need these 3 removed comments?

arcade/draw_commands.py
```diff
@@ -125,13 +121,9 @@ def draw_arc_outline(center_x: float, center_y: float, width: float,
...
     else:
-        uncentered_point_list = []
-        for point in unrotated_point_list:
-            uncentered_point_list.append(rotate_point(point[0], point[1], 0, 0, tilt_angle))
+        uncentered_point_list = [rotate_point(point[0], point[1], 0, 0, tilt_angle) for point in unrotated_point_list]
 
-    point_list = []
-    for point in uncentered_point_list:
-        point_list.append((point[0] + center_x, point[1] + center_y))
+    point_list = [(point[0] + center_x, point[1] + center_y) for point in uncentered_point_list]
```
pvcraven, line 126: This is probably faster

arcade/drawing_support.py
```diff
@@ -108,7 +107,7 @@ def get_three_float_color(color: Color) -> Tuple[float, float, float]:
     :param Color color: Three or four byte tuple
     :return: Three floats as a RGB tuple
     """
-    if len(color) in (3, 4):
+    if len(color) in {3, 4}:
```
pvcraven: This seems like a weird refactor.

arcade/geometry_python.py
```diff
@@ -78,10 +78,7 @@ def _orientation(p: tuple, q: tuple, r: tuple) -> int:
 
     if val == 0:
         return 0
-    if val > 0:
-        return 1  # Collinear
-    else:
-        return 2  # Clock or counterclock
+    return 1 if val > 0 else 2
```
pvcraven: Losing the comments is a bummer. Prefer keeping those in.

arcade/paths.py
```diff
@@ -264,10 +256,4 @@ def astar_calculate_path(start_point: Point,
...
     result = _AStarSearch(mod_start, mod_end, graph)
 
-    if result is None:
-        return None
-
-    # Currently 'result' is in grid locations. We need to convert them to pixel
-    # locations.
-    revised_result = [_expand(p, grid_size) for p in result]
-    return revised_result
+    return None if result is None else [_expand(p, grid_size) for p in result]
```
pvcraven: I wonder if this is faster, but it certainly is not an improvement in readability.

arcade/perf_info.py
```diff
@@ -179,8 +179,7 @@ def get_fps(frame_count: int = 60) -> float:
 
     start_time = _frame_times[-frame_count]
     total_time = cur_time - start_time
-    fps = frame_count / total_time
-    return fps
+    return frame_count / total_time
```
pvcraven: I like the original better, as it is clear what it is doing.

arcade/sprite.py
```diff
@@ -1511,8 +1507,4 @@ def get_distance_between_sprites(sprite1: Sprite, sprite2: Sprite) -> float:
     :return: Distance
     :rtype: float
     """
-    distance = math.sqrt(
-        (sprite1.center_x - sprite2.center_x) ** 2
-        + (sprite1.center_y - sprite2.center_y) ** 2
-    )
-    return distance
+    return math.sqrt((sprite1.center_x - sprite2.center_x) ** 2 + (sprite1.center_y - sprite2.center_y) ** 2)
```
pvcraven: Original is easier to read. Cases like this, I'm not fond of auto-formatters.

pvcraven (changes requested):
There are some good refactors here. Not super-wild about all the ternary usage. Some is ok when the code is short, but I think it makes things hard to read.

arcade/draw_commands.py
alejcas: It is faster indeed

arcade/drawing_support.py
alejcas: This is in fact slower:
https://stackoverflow.com/questions/45017774/python-set-vs-tuple-lookup-is-lookup-in-tuple-o1

Better use the tuple

arcade/paths.py
alejcas: Not faster but shorter. I like this because for me is concise enough. The list comprehension is short enough

arcade/perf_info.py
alejcas: Maybe just add a comment? like `# returning fps`

arcade/draw_commands.py
BrettskiPy: As I attempt to calibrate these kinds of changes, I appreciate the feedback! 😎

arcade/drawing_support.py
BrettskiPy: Yes, this is slower based on my tests as well. I'll change it back to a tuple. I don't know why it suggested this if a set is slower. 🤷

arcade/geometry_python.py
BrettskiPy: I agree. I'll restore these comments. Thank you.

arcade/paths.py
BrettskiPy: I'm ok with either of these. I'll revert it to the original.

arcade/drawing_support.py
alejcas: > Yes, this is slower based on my tests as well. I'll change it back to a tuple. I don't know why it suggested this if a set is slower. 🤷

If set size is dynamic and big enough it will be faster. But not as a fixed size of 2 elements

arcade/perf_info.py
BrettskiPy: I'll revert this back to the original to be consistent with established preferences 😉

arcade/sprite.py
BrettskiPy: I'll revert this thanks.

pvcraven:
I did another pass-through. I think if you update with the changes, I'm happy to merge in.

pvcraven (approved):

PR #1337, opened 2022-09-27
Remove redundant code from Sprite.set_texture

alejcas:
Just removing duplicate code in Sprite.set_texture... and also setting the method return type


alejcas:
That code was duplicated in the property setter of the `texture` property.

So to avoid duplication just call the `texture` setter from the method `set_texture`

einarf:
Not sure how long this methods wills stay around, but let's just merge this 😄

PR #1340, opened 2022-09-29
GUI: Add union method to Rect to allow to get the union of two rects

alejcas:
This adds a method to gui Rect to allow the union between to rects:

```python
# inside Rect class
    def union(self, rect: "Rect"):
        """
        Returns a new Rect that is the union of this rect and another.
        The union is the smallest rectangle that contains theses two rectangles.
        """
        x = min(self.x, rect.x)
        y = min(self.y, rect.y)
        right = max(self.right, rect.right)
        top = max(self.top, rect.top)
        return Rect(
            x=x,
            y=y,
            width=right - x,
            height=top - y
        )
```

### DIFF
--- a/arcade/gui/widgets/__init__.py
+++ b/arcade/gui/widgets/__init__.py
@@ -166,6 +166,22 @@ class Rect(NamedTuple):
 
         return Rect(self.x, self.y, w, h)
 
+    def union(self, rect: "Rect"):
+        """
+        Returns a new Rect that is the union of this rect and another.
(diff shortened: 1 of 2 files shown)


eruvanos:
Hi looks good, could you just add a small test for it in `tests/test_gui/test_rect.py`, that would be great!
Let me know if this works for you.

alejcas:
> Hi looks good, could you just add a small test for it in `tests/test_gui/test_rect.py`, that would be great! Let me know if this works for you.

Done

PR #1347, opened 2022-10-05
Gitignore env

emil-bryniel:
I added the GitHub standard ignores for python environments to the gitignore
[GitHub standard python gitignore](https://github.com/github/gitignore/blob/e5323759e387ba347a9d50f8b0ddd16502eb71d4/Python.gitignore#L122)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@
 
 # virualenvs
 env
+.env
+.venv
+ENV/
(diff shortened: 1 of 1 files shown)


einarf:
Thanks. Looks reasoanble. The local ignore can also be tweaked, but this doesn't hurt.

PR #1354, opened 2022-10-11
Camera code reorganization into SimpleCamera and Camera

alejcas:
This is a new camera approach. Changes vs the previous PR:

- AdvanceCamera is now just Camera. It’s meant to substitute the default arcade Camera. As we discussed in discord the constructor change and therefore it will break previous code with constructor params. The constructor implements a (*, params) approach, so params need to be set via keyword.
- BaseCamera is now SimpleCamera (just a camera with viewport, projection and camera movement methods.
- Camera now inherits from SimpleCamera.
- Fixed a small bug inside Camera update method
- The shake funcionality is still inside Camera. We can move that code into a “ShakeCamera” that will inherit from Camera if you think so.

If it’s ok to you, I can change all references to the old camera to comply with the new one. Also I can write some examples with camera zoom + rotation.


einarf:
Thanks! let's just get this merged.

einarf:
... but maybe look into those failing tests first

alejcas:
> ... but maybe look into those failing tests first

It’s failing the self hosted… don’t know why

einarf:
Things like this. It's not really necessary to pass in w and h here
```
/camera_platform.py", line 132, in setup
    self.camera = arcade.Camera(SCREEN_WIDTH, SCREEN_HEIGHT)
TypeError: Camera.__init__() takes 1 positional argument but 3 were given
```

alejcas:
> Things like this. It's not really necessary to pass in w and h here
> 
> ```
> /camera_platform.py", line 132, in setup
>     self.camera = arcade.Camera(SCREEN_WIDTH, SCREEN_HEIGHT)
> TypeError: Camera.__init__() takes 1 positional argument but 3 were given
> ```

Oh, my bad, let me commit here changing all references to the new camera constructor.

einarf:
Can probably just search "Camera(" and get most of the issues.

alejcas:
Work done! note that I mainly change `arcade.Camera` with `arcade.SimpleCamera` in the examples.

PR #1358, opened 2022-10-21
Add general NinePatchTexture support within all UIWidgets

eruvanos:
@einarf This removes the position and draw_size from the NinePatchTexture.
Right now the shader code does not allow to set some draw position, which is not a dealbreaker due to the scissor box used while rendering each widget.

Anything we should change before merging?


einarf:
Looks fine. I would like to have the old `draw()` back and support position so it can be used independenly of the gui.

If that didn't work before these changes we can fix that after merge.

einarf (approved):

PR #1363, opened 2022-10-26
Camera resize: resize_projection

alejcas:
`Camera.resize` method now includes a `resize_projection` option that will be False by default.
On True the method will resize the projection as well to be width / height equal to the resized viewport.

I also fix all example code to use resize_projection.

### DIFF
--- a/arcade/camera.py
+++ b/arcade/camera.py
@@ -44,7 +44,7 @@ class SimpleCamera:
         if viewport is not None and projection is None:
             # if viewport is provided but projection is not, projection
             # will match the provided viewport
-            self._projection: FourFloatTuple = (viewport[0], viewport[2], viewport[1], viewport[3])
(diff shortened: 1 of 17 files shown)


arcade/camera.py
```diff
@@ -184,16 +186,21 @@ def get_map_coordinates(self, camera_vector: Union[Vec2, tuple]) -> Vec2:
         """
         return Vec2(*self.position) + Vec2(*camera_vector)
 
-    def resize(self, viewport_width: int, viewport_height: int) -> None:
+    def resize(self, viewport_width: int, viewport_height: int, *,
+               resize_projection: bool = False) -> None:
```
einarf: Can we set this to True instead to have better backwards compatibility?

alejcas: Done

PR #1379, opened 2022-11-02
Fix for @width.setter and @height.setter

Felipe-Carlesso:
- Fix for ```@width.setter``` and ```@height.setter``` when their old values are zero

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -557,11 +557,19 @@ class Sprite:
             self.clear_spatial_hashes()
             self._point_list_cache = None
 
-            # If there is a hit box, rescale it to the new width
-            if self._points:
-                scale = new_value / self._width
-                old_points = self._points
-                self._points = [(point[0] * scale, point[1]) for point in old_points]
+            # If current width is not zero, rescale current hitbox
+            if self._width != 0.0:
+
+                # If there is a hit box, rescale it to the new width
+                if self._points:
+                    scale = new_value / self._width
+                    old_points = self._points
+                    self._points = [(point[0] * scale, point[1]) for point in old_points]
+            
+            else:
+
+                # Redefine hit box points
+                self._points = self.create_hit_box_points(new_value, self._height)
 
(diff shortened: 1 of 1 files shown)


arcade/sprite.py
```diff
@@ -557,11 +557,19 @@ def width(self, new_value: float):
             self.clear_spatial_hashes()
             self._point_list_cache = None
 
-            # If there is a hit box, rescale it to the new width
-            if self._points:
-                scale = new_value / self._width
-                old_points = self._points
-                self._points = [(point[0] * scale, point[1]) for point in old_points]
+            # If current width is not zero, rescale current hitbox
+            if self._width != 0.0:
```
einarf: `if self._width` is probably sufficient here?

arcade/sprite.py
```diff
@@ -557,11 +557,19 @@ def width(self, new_value: float):
...
+
+                # If there is a hit box, rescale it to the new width
+                if self._points:
+                    scale = new_value / self._width
+                    old_points = self._points
+                    self._points = [(point[0] * scale, point[1]) for point in old_points]
```
einarf: Can we just put the if test here?

einarf:
I'm actually not entirely sure what the best solution is here. We are modifying `_points`. This is the unique hit points for the sprite before translation and rotation. When scale is set to 0 the unqiue hit box points for the sprite is gone forever and cannot be reconstructed unless `None` hit box algo is used.

Your solution works when simple hit boxes are used, but in any other case it breaks the sprite collision.

einarf:
After a longer discussion in discord, we found a better solutuon. Since scale is supposed to be ratio of the current size and the texture size we can simply calculate that. We only update the scale values so the hit box is properly recalcualted. This have some other implications related to missing textures on sprites that we will solve in another branch before 2.7 release.

https://github.com/pythonarcade/arcade/commit/b10bdd69627011d272177ad909828e4d6f44f931

PR #1383, opened 2022-11-03
fix docstrings return types on `Sprite.collides_with_list`

alejcas:
Just fixing the return type for `Sprite.collides_with_list`


einarf:
Thanks!

PR #1385, opened 2022-11-03
Possible error in ShapeElementList.remove

alejcas:
This seems like an error in `ShapeElementList.remove`.


eschan145:
Probably from Copy and Paste.

einarf:
The group should definitely be added to the `dirties` set regadless if it's add or remove. The bug here might be that the group should be removed if it's empty?

alejcas:
> The group should definitely be added to the `dirties` set regardless if it's add or remove. The bug here might be that the group should be removed if it's empty?

Don't know... just closing this as my comment was specifically about dirties

PR #1387, opened 2022-11-04
Sections: Fix for events on_show_section and on_hide_section 

alejcas:
Fix: events `on_show_section` and `on_hide_section` were not being fired on add_section or remove_section


einarf:
Looks good!

PR #1399, opened 2022-11-07
Mypy/fix pep 484

eruvanos:


alejcas:
Hi, shouldn't methods from spatial_hash.py like `check_for_collision_with_list` also return `List[_SpriteType]` instead of `List["Sprite"]`?

Thanks

alejcas:
for example if I have a subclass of Sprite called Unit whatever`Unit.collides_with_list()` returns is always expected to be a `List[Sprite]` and not a `List[Unit]`

PR #1405, opened 2022-11-18
Fix typos, formatting errors, etc.

eschan145:
Changed several comments to capitalization and added more docs.

* _Counter-clock-wise_ → **counterclockwise**
* _Over-ride_ → **override**

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -1,5 +1,5 @@
 """
-Drawing text with pyglet label
+Draw and display text with pyglet labels and batching.
 """
 from pathlib import Path
 from typing import Any, Optional, Tuple, Union
@@ -14,7 +14,8 @@ from arcade.resources import resolve_resource_path
 
 def load_font(path: Union[str, Path]) -> None:
     """
-    Load fonts in a file (usually .ttf) adding them to a global font registry.
+    Load fonts in a file (usually ``*.ttf``) adding them to a global
+    font registry.
 
     A file can contain one or multiple fonts. Each font has a name.
     Open the font file to find the actual name(s). These names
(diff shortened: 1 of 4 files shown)


eschan145:
I might add more commits to this, so don't merge immediately.

pvcraven:
> I might add more commits to this, so don't merge immediately.

Let us know when you are ready.

eschan145:
@pvcraven should I change all of the line widths to 80 characters? It's inconsistent all over the place. Some are wrapped to 100 chars, some to 80.

arcade/text.py
```diff
@@ -234,7 +236,7 @@ def group(self, group: pyglet.graphics.Group):
     @property
     def value(self) -> str:
         """
-        Get or set the current text string to display.
+        Current text string to display.
```
einarf, line 239: It's important to communicate if the property is read/write. That's the reason for the "get or set" part.

This applies to most or all propreties here.

einarf (changes requested):

einarf:
It would probably be better/easier to make smaller PRs for these things? Something that could just be looked at briefly and merged within the same day or week. Some of these comments are already changed in other branches for example.. or the code have moved or changed in other ways.

arcade/text.py
eschan145: It's kind of repetitive, though

eschan145: If it's read only, it's usually stated that `Read-only` to clarify.

einarf: It's definitely repetative, but it's explicit not creating doubt. We've had so many examples of users not understanding the properties can actually be set. It's immediately visible in your IDE, python `help()` and in sphinx.

eschan145:
@einarf I think I'm ready to merge. There is one conflict in `window_commands`. My pull request should override it.

einarf:
There are lots of other cross branch conflicts due to this PR being so old and way too large.

Still, there are some good things in here. I suspect this PR just needs to be manually applied after the texture-revamp branch is merged.

PR #1420, opened 2022-11-30
UI: Add Transitions to arcade GUI [after 3.0 release]

eruvanos:
Adds a complete **GUI animation stack** to Arcade, shipped under `arcade.gui.experimental` so the novel per-property takeover semantics, cached overdraw and hover-zone behavior can be iterated on without deprecations. Stabilization later is purely additive (e.g. promoting `animate()` onto `UIWidget` once proven).

## High-level API

`UIAnimatedGroup` is the main entry point. It wraps a single child (or a whole subtree), renders it into a cached surface, and tweens transform properties that **do not** affect layouting:

```python
from arcade.gui.experimental import UIAnimatedGroup

group = UIAnimatedGroup(child=UIFlatButton(text="Play"))
layout.add(group)

# scale to 108% within 0.15s, with a "back out" overshoot
group.animate(scale=1.08, duration=0.15, ease=Easing.BACK_OUT)
```

Animatable transform properties: `scale`, `angle`, `alpha`, `offset_x`, `offset_y`, `tint`.

`animate()` supports:
- multiple properties per call
- sequencing via `.then(...)`, completion callbacks via `.on_finish(...)`, cancellation via `.stop()`
- repetition (`repeat`, `yoyo`), delays, easing (`arcade.anim.Easing`)
- relative targets via `rel()` (e.g. `angle=rel(-360)`)
- **per-property takeover**: starting a new animation takes over just its properties from running ones — so hover in/out is a one-liner with no cleanup.

`UIAnimatedGroup` is interactive: `hovered` / `pressed` / `on_click` are hit-tested against the group's **untransformed** rect, so the trigger zone stays stable while the visuals scale or rotate. The child subtree still receives mouse events transformed into its local coordinate system.

```python
def on_hover_change():
    if group.hovered:
        group.animate(scale=1.08, duration=0.15, ease=Easing.BACK_OUT)
    else:
        group.animate(scale=1.0, duration=0.15)

bind(group, "hovered", on_hover_change)
```

`UIRenderGroup` is the non-interactive caching primitive underneath, available for use on its own.

## Low-level transitions

`animate()` is built on a small set of composable `Transition*` primitives (also exported from `arcade.gui.experimental`), which can be combined with `+` (parallel) and `|` (sequential):

- `TransitionAttr` — tween a value over time (start → end)
- `TransitionAttrIncr` — increment a value over time
- `TransitionAttrSet` — set a value after a delay
- `TransitionParallel` — run transitions in parallel
- `TransitionChain` — run transitions sequentially
- `TransitionDelay` — pause within a chain

`Animation` implements the `TransitionBase` protocol, so animations and raw transitions compose freely.

## Examples

- `arcade.examples.gui.exp_animations` — an animated main menu: staggered pop-in entrances, pulsing/swaying idle animations, hover pops, spinning tiles, fade-out on quit.
- `arcade.examples.gui.exp_animations_2` — hoverable cards that lift and enlarge.
- `arcade.examples.gui.transitions` — a minimal low-level showcase animating `offset_x`/`offset_y`.

## Docs & tests

- New programming guide page `doc/programming_guide/gui/animations.rst` plus updates to `concepts.rst`; the experimental API is indexed under GUI Experimental Features.
- Full unit coverage in `tests/unit/gui/` for `animate`, `transition`, and the render/animated groups (parent-space hover, transform-invariant hit-testing, ticking, takeover semantics).

## Notes

- Supporting changes to the GUI render path (`surface.py` + the `surface_vs/fs.glsl` shaders), the reactive `property.py` system (adds `AliasProperty`), and `UIScrollArea` came along on this branch.
- Everything animation-related lives in `arcade.gui.experimental` — the stable `arcade.gui` namespace does not export any of it yet.


eruvanos:
Talked to @einarf :

- [ ] missing feature: cancel transition
- [ ] for Sprites, this might be a performance problem

eruvanos:
Will be added after 3.0 release. It is more important to release!

Issue #1422, opened 2022-12-02
Switch color constants to RGBA


pushfoo:
### What should be added/changed?

- [X] Make all color constants in `arcade.color` into RGBA constants
- [X] Make all color constants in `arcade.csscolor` into RGBA constants
- [x] Add a `TRANSPARENT_BLACK` constant to `arcade.color`

### What would it help with?
1. Provide better immediate-term support for sprite generation with Pillow or drawing into contexts
2. Set the stage for future sprite generation API improvements
3. Make tests for #1342 and other functionality cleaner

This is also a chore einarf and Clepto agreed should be done months ago

einarf:
I have already made them into RGBA in my branch
https://github.com/pythonarcade/arcade/blob/texture-revamp/arcade/color/__init__.py
https://github.com/pythonarcade/arcade/blob/texture-revamp/arcade/csscolor/__init__.py

pushfoo:
How long till that gets merged? I'd like to add some more constants to `arcade.color` and revamp the docstring for `arcade.csscolor` afterward.

einarf:
Think we can get away with cherry picking that commit into development. I can look at that a bit later today

einarf:
https://github.com/pythonarcade/arcade/pull/1423

Issue #1342, opened 2022-10-02
Build improved screenshot API


pushfoo:
## Enhancement request:

**tl;dr add beginner-friendly screenshot functionality with reasonable defaults**

#### The minimum

A function that grabs the image from the screen and immediately writes it to disk. It should meet the following requirements:

1. Calling with zero arguments saves a screenshot to the current working directory with a reasonable [time stamp format](https://docs.python.org/3/library/datetime.html?highlight=strftime#strftime-strptime-behavior) such as `"%windowname_%Y%m%d_%H%M_%S_%f.jpg"`
2. Accepts the same optional bounds arguments as `get_image`
3. Accepts an optional `save_path` argument that directly specifies the directory to save to or filename to use, overriding any formatting in the latter case
 4. Accepts an optional `time_format` argument to override the default timestamp format.

 #### Additional features
 
We may also want to add :
* A screenshot manager object that allows setting a default filename / timestamp format
* Queuing features that allow multiple screenshots to queued up before writing to disk at a specified time, such as on game exit to avoid slow disk access during gameplay
* Integration with the pyglet event system (example: take a screenshot whenever event or condition X happens)
* ~~Integrations with video format libraries to automatically assemble a movie of the screenshots~~ Moved to #1349 

### What would it help with?

1. Reduce the number of tools users need to install or learn to get help
2. Make the cross-platform experience much less confusing
2. Make #1338 far easier to write

pushfoo:
@einarf 

> It's possible to pipe each frame directly into ffmpeg instead. We don't have a wrapper for this, but it's not too difficult to make.

I noticed a previous commit removed ffmpeg as a windows dependency. Would we add it back in, or rely on a windows API for encoding?

einarf:
> I noticed a previous commit removed ffmpeg as a windows dependency. Would we add it back in, or rely on a windows API for encoding?

I think ffmpeg is something people would install separately. It's too much to bundle this for every platform.

pushfoo:
> It's too much to bundle this for every platform.

Which do you mean?

* ffmpeg is overkill for this problem
* managing ffmpeg as an included dependency creates too much work for the project
* including ffmpeg makes the install size too large

einarf:
Streaming raw frames into ffmpeg is very useful. A lot of other libraries do this (Wasabi2D, moderngl-window, etc)

The last option is true: including ffmpeg makes the install size too large

It's something you install on the side.

pushfoo:
[...]

Issue #1194, opened 2022-04-16
Documentation Reorganization


pvcraven:
Reorganize the documentation into the following four sections.

The programming guide and the tutorials will be made into printable/downloadable PDFs.

* Programming Guide
  * Installation on Windows, Mac, Linux, and Raspberry Pi
  * Basic drawing commands (this is where we might need to use pyglet shapes instead?)
  * Sprites
    * Using Sprites
    * Using sprite lists
    * Drawing sprites in layers
    * Textures
    * Texture Atlas
    * Sprite animation
    * Sprite movement
    * Avoid spawning sprites on walls
    * Sprite drag and drop
    * Fixing edge artifacts
  * Sprite physics
    * Sprite collisions
      * Collision detection methods (Normal, Spatial, GPU)
    * Simple physics engine
    * Pymunk
  * Working with views
    * Adding instruction screen, game over, pause screen
  * Working with the camera
    * Scrolling around
    * Full screen
    * Window resizing
    * Camera shake, panning
  * Tile map editor support
    * Loading maps
    * Animation
    * Custom hit boxes
    * Object layer
  * Sound support
    * Streaming vs. Loading
  * GUI
    * GUI stuff here
  * OpenGL
    * OpenGL and the pipeline (How data flows)
    * High level explanation of each opengl type we have (buffer, texture, geometry, etc etc)
    * A library explaining various rendering techniques to get people started
  * Distributing your game via PyInstaller
  * Performance tips
    * Arcade performance information
  * Headless Arcade
  * Vertical synchronization
  * Arcade vs. Pygame
  * Contributing to Arcade
    * Release check-list

* Tutorials
  * Simple Platformer
  * Pymunk Platformer
  * Using Views for Menus and Game Over Screens
  * Solitaire Game
  * Shadertoy Glow
  * Shadertoy Particles
  * Using the CRT filter
  * GPU particle burst
  * Ray-casting Shadows
  * Lights

* Example Code

* API Docs

Certain items would remain outside these sections, but still appear on the homepage:

* Social links
* Example games
* Links to GitHub
* Release notes
* Diverse coders section

pvcraven:
I really thing the programming guide should be organized around subjects. Maybe something like the skill tree.
https://api.arcade.academy/en/latest/get_started.html

einarf:
[...]

PR #1428, opened 2022-12-04
Fix color table rendering & RGBA constant compatibility

pushfoo:
Color table generation has been broken for a while because the API doc paths it assumed were changed by #1194. We didn't notice until it was highlighted by #1422 and #1342.

This PR does the following:

1. Fix color table build
2. Add support for rendering RGBA color constants with non-255 alpha value into proper CSS (3 unsigned bytes + normalized float)
3. Enhance standards compliance in small ways: a raw string for a regex, move some styling behavior into CSS from HTML

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -271,19 +271,24 @@ def source_read(_app, docname, source):
 
         original_text = source[0]
         append_text = "\n\n.. raw:: html\n\n"
-        append_text += "    <table>"
+        append_text += "    <table class='colorTable'><tbody>\n"
         color_file = open(filename)
 
         for line in color_file:
             match = p.match(line)
+
             if match:
-                append_text += "    <tr><td>"
-                append_text += match.group(1)
-                append_text += "</td><td>"
-                append_text += match.group(2)
-                append_text += f"<td style='width:80px;background-color:rgba{match.group(2)};'>&nbsp;</td>"
-                append_text += "    </td></tr>\n"
-        append_text += "    </table>"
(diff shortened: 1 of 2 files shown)


pvcraven:
Thanks!

doc/conf.py
```diff
@@ -259,30 +259,36 @@ def source_read(_app, docname, source):
...
         filename = "../arcade/color/__init__.py"
-    elif docname == "arcade.csscolor":
+    elif docname == "api_docs/arcade.csscolor":
         filename = "../arcade/csscolor/__init__.py"
 
     if filename:
+        # print(f"  XXX Handling color file: {filename}")
         import re
-        p = re.compile("^([A-Z_]+) = (\\(.*\\))")
+        p = re.compile(r"^([A-Z_]+) = (\(.*\))")
```
thomascookandroid, line 270: What does  this regex change do?

pushfoo: It replaces the string with a raw string, which does not process escapes inside it. This removes the need to escape backslashes, which in turn makes the regex inside it more legible. Please see [the doc for the `re` module](https://docs.python.org/3/library/re.html) for more information:
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'. So r"\n" is a two-character string containing '\' and 'n', while "\n" is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

PR #1431, opened 2022-12-07
Camera zoom changes

alejcas:
This PR changes a bit how zoom is controlled and applied to cameras.

@einarf sorry to ping but I would like you to review if this is somehow the right approach. In the end what's send to the gl context should be the same, but now the users gets correct and consistent info about the zoom applied.

**A)** Before, you could set a zoom that would change the projection matrix, but you could also change the ratio between the viewport and the projection effectively changing the zoom level again:

```python
# ...
camera = arcade.Camera(viewport=(0, 0, 800, 600))
# at this point projection and viewport matches

camera.zoom = 0.5
# now camera zoom is 0.5

camera.projection = (0, camera.viewport[2] / 2, 0, camera.viewport[3] / 2)
# now, by changing the projection I have applied a zoom indirectly,

# BUT. camera.zoom still says 0.5 when it's not:
print(camera.zoom)  # returns 0.5
```

With this PR, a change in zoom will effectively change the projection (adding or removing width and height to the projection).
Getting `camera.zoom` will show the ratio between projection and viewport (not an scalar value set by the user).
Setting `camera.zoom` will effectively change the projection like this: 

```python
left, right, bottom, top =  camera.projection
camera.projection = left, right * camera.zoom, bottom, top * camera.zoom
```

Also if the user decides to not change camera.zoom but instead change the ratio between viewport and projection the `camera.zoom` property will return a good value, effectively showing the zoom applied.

**Also, as you may probably notice by now, changing the projection so it's 1/2 of the viewport or setting `camera.zoom` to 0.5 will produce the same effect and reflect the same values across all camera attributes.**

**B)** I've added a `camera.scale` property that returns the x, y ratio between projection and viewport. And camera.zoom will return the width scale (if x, y scale is not the same, then zoom will return a result that's not completely correct).
Setting camera.zoom will set camera.scale to the same value. 

**C)** As a result of this change, zoom is now available in `SimpleCamera` (as `SimpleCamera` allows to change viewport and projection).

**D)** This also fixes an error in `camera.center`. Later the center method was not taking into account the projection to view ratio.

**E)** I've also added an example so you can see all this changes in action and also show how you can draw into the screen using cameras instead of using atlas render into.

### DIFF
--- a/arcade/examples/minimap_camera.py
+++ b/arcade/examples/minimap_camera.py
@@ -182,11 +182,6 @@ class MyGame(arcade.Window):
 
         # Center the minimap viewport to the player in the minimap
         self.camera_minimap.center(self.player_sprite.position, CAMERA_SPEED)
-        # print(self.camera_sprites.position, self.camera_sprites.get_adjusted_position(),
-        #       self.camera_minimap.position, self.camera_minimap.get_adjusted_position())
-        # print(self.camera_sprites.sync(self.camera_minimap))
-        print(self.camera_minimap.scale)
-        print(self.camera_minimap.projection)
 
     def on_resize(self, width: int, height: int):
         """
--- a/arcade/examples/minimap_camera.py
+++ b/arcade/examples/minimap_camera.py
@@ -8,10 +8,8 @@ python -m arcade.examples.minimap
 """
 
 import random
-from uuid import uuid4
 
 import arcade
-from pyglet.math import Vec2
 
 SPRITE_SCALING = 0.5
 
--- a/arcade/examples/minimap_camera.py
+++ b/arcade/examples/minimap_camera.py
@@ -1,0 +1,214 @@
+"""
+Work with a mini-map
+
+Artwork from https://kenney.nl
+
+If Python and Arcade are installed, this example can be run from the command line with:
+python -m arcade.examples.minimap
+"""
+
+import random
+from uuid import uuid4
+
+import arcade
+from pyglet.math import Vec2
+
+SPRITE_SCALING = 0.5
+
+DEFAULT_SCREEN_WIDTH = 800
+DEFAULT_SCREEN_HEIGHT = 600
+SCREEN_TITLE = "Minimap Example"
+
+# How many pixels to keep as a minimum margin between the character
+# and the edge of the screen.
+VIEWPORT_MARGIN = 220
+
+# How fast the camera pans to the player. 1.0 is instant.
+CAMERA_SPEED = 0.1
+
+# How fast the character moves
+PLAYER_MOVEMENT_SPEED = 7
+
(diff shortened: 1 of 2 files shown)


alejcas:
Btw I don't know what's failing in self hosted

arcade/examples/minimap_camera.py
```diff
@@ -0,0 +1,212 @@
+"""
+Work with a mini-map
+
+Artwork from https://kenney.nl
+
+If Python and Arcade are installed, this example can be run from the command line with:
+python -m arcade.examples.minimap
```
pushfoo: One of the test run errors appears to be because the run command in the docstring does not match the name of the file exactly: Try this:
```suggestion
python -m arcade.examples.minimap_camera
```

arcade/examples/minimap_camera.py
```diff
@@ -0,0 +1,212 @@
...
+        # Center the screen to the player
+        self.camera_sprites.center(self.player_sprite.position, CAMERA_SPEED)
+
+        # Center the minimap viewport to the player in the minimap
+        self.camera_minimap.center(self.player_sprite.position, CAMERA_SPEED)
+        # print(self.camera_sprites.position, self.camera_sprites.get_adjusted_position(),
+        #       self.camera_minimap.position, self.camera_minimap.get_adjusted_position())
+        # print(self.camera_sprites.sync(self.camera_minimap))
+        print(self.camera_minimap.scale)
+        print(self.camera_minimap.projection)
```
pushfoo: If I read the test output correctly, the second error is due to a recently added check to ensure examples do not print to the console. Try replacing these print statements with Text objects that are displayed inside the window.

arcade/examples/minimap_camera.py
alejcas: There is a minimap.py and now a minimap_camera.py. Both are correct

arcade/examples/minimap_camera.py
alejcas: I’ll change this!

alejcas: Thanks!

arcade/examples/minimap_camera.py
pushfoo: I think there is a misunderstanding somewhere. To my knowledge, the tests expect the module name in an example's run instructions to match its filename exactly. This particular example (`minimap_camera.py`) is separate from `minimap.py`, so it must use the run instruction line I included above.

alejcas: Ok! I understand now!thanks

arcade/examples/minimap_camera.py
alejcas: Done

einarf:
Let's just merge this

PR #1436, opened 2022-12-15
Sprite: Change velocity into a tuple

einarf:
This solves a a few problems
* `velocity` is now a publicly documented member of Sprite
* Users setting `sprite.velocity = 1, 1` will no longer get weird crashes
* Even if the users supplies a list they will be fine


einarf:
I don't know how this will affect persomance considering we're adding some extra tuple creations.

It can be changed into two separate float value members if needed, but it should definitely not be a list.

PR #1454, opened 2023-01-11
adding contex manager to camera

DragonMoffon:


einarf:
Did you also test this locally?

DragonMoffon:
@einarf I just tested it, and It has an issue that `window.current_camera` is None until the first camera is used. We should have a default camera. We could even add a viewport argument to the window which modifies the default camera (this does have the added confusion of arcade using viewport in multiple different ways, so maybe not).

einarf:
Yup. We need to figure out something here. We have the old style `set_viewport` that sets projection and viewport assuming a they are both the same size (no zooming) and setting the raw viewport. Then pyglet also have its one viewport and projection that we already wrap.

There could be a lot of confusion around this. Would setting viewport and projection update the active camera?

einarf:
I'd say this is something we can expore later in 2.7. The topic around how to handle a default camrea is fairly complex and can introduce breaking changes. Still, its probably a good idea in the long run since the default state is preserved.

einarf:
let's close this for now until we figure out what to do with default camera etc

PR #1467, opened 2023-01-19
Build is platform dependent, not independent.

codiophile:
Just saw this contradiction on the website and thought I could quickly fix it.


einarf:
Thanks. Looks like we need to pin the mypy version in that branch. I will just merge this.

PR #1478, opened 2023-01-24
Added size hinting implementation in UIGridLayout

Ibrahim2750mi:
#1101

### DIFF
--- a/arcade/gui/widgets/layout.py
+++ b/arcade/gui/widgets/layout.py
@@ -654,4 +654,8 @@ class UIGridLayout(UILayout):
         :param height: Base height on which the widget is to be resized.
         """
         for child, _ in self._children:
-            child.resize(width=width*child.size_hint[0], height=height*child.size_hint[1])
+            size_hint_width = 0
+            size_hint_height = 0
+            if child.size_hint:
+                size_hint_width, size_hint_height = (child.size_hint[0] or 0), (child.size_hint[1] or 0)
+            child.resize(width=(width*size_hint_width) or child.width, height=(height*size_hint_height) or child.height)
--- a/arcade/gui/widgets/layout.py
+++ b/arcade/gui/widgets/layout.py
@@ -458,6 +458,16 @@ class UIGridLayout(UILayout):
         # initially update size hints
         self._update_size_hints()
 
+    @staticmethod
+    def _layouting_allowed(child: UIWidget) -> Tuple[bool, bool]:
+        """
+        Checks if size_hint is given for the dimension, which would allow the layout to resize this widget
+
+        :return: horizontal, vertical
+        """
+        sh_w, sh_h = child.size_hint or (None, None)
+        return sh_w is not None, sh_h is not None
+
     def _update_size_hints(self):
 
         child_sorted_row_wise = [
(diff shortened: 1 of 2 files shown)


eruvanos:
Actually, the implementation is not what I expected, but I will check, if I was missing some tests. I thought more into the direction, that the do_layout follows the behaviour from:  https://github.com/pythonarcade/arcade/blob/development/doc/programming_guide/gui/concept.rst#uiwidget

UIWidget:
- **size_hint**: tuple of two floats, defines how much of the parents space it would like to occupy (range: 0.0-1.0). For maximal vertical and horizontal expansion, define size_hint of 1 for the axis.
- **size_hint_min** - tuple of two ints, defines minimal size of the widget. If set, changing the size of a widget to a lower values will use this size instead.
- **size_hint_max** - tuple of two ints, defines maximum size of the widget. If set, changing the size of a widget to a higher values will use this size instead.

size_hint, size_hint_min, and size_hint_max are values that are additional information of a widget, but do not effect the widget on its own. [:class:`UILayout`](https://github.com/pythonarcade/arcade/blob/development/doc/programming_guide/gui/concept.rst#id16) may use these information to place or resize a widget.

`UILayout` are widgets, which reserve the option to move or resize children. They might respect special properties of a widget like size_hint, size_hint_min, or size_hint_max.

The `UILayouts` only resize a child widget (x or y axis) if size_hint provides a value for the axis which is not None for the dimension.

Ibrahim2750mi:
> The `UILayouts` only resize a child widget (x or y axis) if size_hint provides a value for the axis which is not None for the dimension.

Ah, I will add a check if the widget's size hints are not None before resizing them. Other than that I think it is following the arcade gui concept right?

Ibrahim2750mi:
> 1. UIGridLayout should cover the content by default and not take more space.
Which results in the default of size_hint=(0,0) and the requirement, that size_hint_min is always set to the minimal size it requires to cover all children.
> 2. If a user wants to use UIGridLayout to place something evenly on the screen, it could be stretched to the full parent size by setting size_hint=(1,1).
Which should result in more space than required within each cell (ignoring the edge case where the parent is not big enough)
> 3. If a user wants to define the size of the UIGridLayout,  setting the size_hint=None should fix the UIGridLayout to the current size. (This is respected in other UILayouts). So the resize function or setting the rect directly can be used to set the size. 
(Ignoring again the edge case, that the size might not be big enough for the children) This is basically a special case of 2.
Maybe constructor args (width, height) would not be a bad idea? Not sure.
> How to deal with children size hints.
>size_hint = None means the same like size_hint=(None, None) which states that the widget should not be resized in any dimension. 
This always applies only to the axis: size_hint=(1, None) would mean, grow the child to maximal available width, but do not touch the height at all.
If the size_hint for an axis is 0, it should be shrunk to 0 or to the given size_hint_min value for the axis. The other way around with size_hint_max as the maximal growth value.
> The UIGridLayout size_hint_min[0] should be the sum of:
> - (child.size_hint_min[0] or 0) if child has size_hint[0] or
> - child.width
> Same applies to size_hint_min[1] referring the child's height.
> Combining this with the cases from above:
> Children are allowed to take grow within the bounds of their cell. E.g:
>- size_hint[0]==0.5  -> half of the available cell width
> - size_hint[0]==1  -> full available cell width

New changes in the code for size hinting is according to this text.

eruvanos:
Great work, thank you!

PR #1486, opened 2023-01-31
Gui/issue 1482

eruvanos:


eruvanos:
Still does not feel right, how the width is handled.

PR #1514, opened 2023-02-12
UI: Fix bug of size hints not updating when setting border

Ibrahim2750mi:
The bug: WIdgets in Layout ate up the border.
image

After fix:


eruvanos:
Great work!

PR #1516, opened 2023-02-13
add constants for F17-F24

DigiDuncan:
Tested by binding a key in Razer Synpase to F17-F24, printing the value returned from `on_key_press()`, and setting the constant value to what was returned.


arcade/key/__init__.py
```diff
@@ -122,6 +122,14 @@
...
 F15 = 65484
 F16 = 65485
+F17 = 549755813888
+F18 = 554050781184
+F19 = 558345748480
+F20 = 562640715776
+F21 = 566935683072
+F22 = 571230650368
+F23 = 575525617664
+F24 = 579820584960
```
pushfoo: Change the 206 in this file to 214 and the build should pass:
https://github.com/pythonarcade/arcade/blob/2c4fd952223fe86dc8fbba958e90f3103cbd9c99/tests/unit/test_key.py#L2-L5

pushfoo (changes requested):
This isn't a spurious failure from connectivity issues. There's a CI check that needs to be updated.

einarf:
These are window only then? It would pobably be better doing this change in pyglet? I'm not entirely sure.

Also there are unit tests failing.

DigiDuncan:
Updated the constants to match pyglet's implementation. These do not currently work, because pyglet's backend does not support F17-F24 yet, but will in a future update (https://github.com/pyglet/pyglet/pull/773).

pvcraven:
Are you going to update the unit test so this passes? Far as I can tell, that's the only thing keeping it from being merged.

DigiDuncan:
> Are you going to update the unit test so this passes? Far as I can tell, that's the only thing keeping it from being merged.

Done and done, apologies for the wait time.

pushfoo (approved):
It looks ready to merge now.

PR #1522, opened 2023-02-15
1502 create lines fix

### LINKED ISSUE #1502: create_lines_with_colors doesn't work properly for width > 1
## Bug Report
When invoking create_lines_with_colors with width > 1 the output looks like (but not exact) call to line_strip functions:
all segments of lines are connected.
When invoking create_lines_with_colors with width 1 everything is as expected (lines segments are separated)

## System Info
Run this and paste the output here: python -m arcade
Arcade 2.6.17
-------------
vendor: NVIDIA Corporation
renderer: NVIDIA GeForce GTX 960M/PCIe/SSE2
version: (4, 6)
python: 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 16:30:00) [MSC v.1900 64 bit (AMD64)]
platform: win32

### Actual behavior:
with width > 1 (thick lines code) all points in points array are connected, colors are gradients in every other connection:
```
(0, 1) color: (c0, c1)
(1, 2) color: (c1, c2) <==== wrong segment
(2, 3) color: (c2, c3)
...
```
### Expected behavior:
Connected only pairs of points i.e. (0,1), (2,3), etc..
```
(0, 1) color: (c0, c1)
(2, 3) color: (c2, c3)
...
```

```
        self.shape_list = arcade.ShapeElementList()
        point_list = ((0, 50),
                      (10, 10),
                      (50, 0),
                      (10, -10))
        color = [random.choice(colors) for i in range(len(point_list))]
        my_lines = arcade.create_lines_with_colors(points, color, 4)
```

## Enhancement request:
fix 
1. modify create_lines_with_colors so that 2 triangles are configured instead of 1 rectangle
```
-            new_color_list += color1, color1, color2, color2
-            triangle_point_list += points[1], points[0], points[2], points[3]
+           new_color_list += color1, color1, color2, color1, color2, color2
+           triangle_point_list += points[0], points[1], points[2], points[0], points[2], points[3]
```
2. modify create_triangles_filled_with_colors so that separate triangles are rendered:
```
-    shape_mode = gl.GL_TRIANGLE_STRIP
+    shape_mode = gl.GL_TRIANGLES
```
3. 
### What would it help with?
bring the expected behavior

kosvitko:
1. added fix to create_lines_with_colors
2. created triangles_strip wrapper to handle other functions which use that option
3. added some comments to documentation section of changed functions

sidenote regarding related code: 
create_line_strip function in original version of code causes twice more triangles to be rendered due to call to triangle_strip function (one triangle per vertex, so 4 triangles per line segment).
This behavior could be changed (2 triangles would be rendered for one line segment) by using no strip triangles function, but vertex array would increase 1.5 times (6 vertex for one line segment vs 4)

Built and tested locally

Closes #1502

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -217,6 +217,7 @@ from .buffered_draw_commands import create_rectangle_filled_with_colors
 from .buffered_draw_commands import create_rectangle_outline
 from .buffered_draw_commands import create_rectangles_filled_with_colors
 from .buffered_draw_commands import create_triangles_filled_with_colors
+from .buffered_draw_commands import create_triangles_strip_filled_with_colors
 from .buffered_draw_commands import get_rectangle_points
 
 from .draw_commands import draw_arc_filled
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -454,6 +454,7 @@ __all__ = [
     'create_rectangles_filled_with_colors',
     'create_text_sprite',
     'create_triangles_filled_with_colors',
+    'create_triangles_strip_filled_with_colors',
     'disable_timings',
     'draw_arc_filled',
     'draw_arc_outline',
--- a/arcade/buffered_draw_commands.py
+++ b/arcade/buffered_draw_commands.py
@@ -481,6 +481,7 @@ def create_triangles_filled_with_colors(point_list, color_list) -> Shape:
     shape_mode = gl.GL_TRIANGLES
     return create_line_generic_with_colors(point_list, color_list, shape_mode)
 
+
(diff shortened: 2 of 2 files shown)


kosvitko:
* make full is performed locally

pvcraven:
Hi, looks like the auto-code check found some minor formatting issues:
https://github.com/pythonarcade/arcade/actions/runs/4183291764/jobs/7258906425
Can you fix and make the formatters happy?

kosvitko:
definitely. just for the first time the workflow was disabled, so I wan't able to run checks ;)

kosvitko:
ok. one more fix to complete.

PR #1504, opened 2023-02-10
Proposed benchmark infra, adds collision detection benchmark, experimental performance improvements

cspotcode:
This is a draft meant to accompany Discord conversation, but is not meant to be merged.  Individual commits will, I hope, be useful and can be cherry-picked into separate pull requests.

The reports are confusing, I need to clean them up.  But basically, every commit on this branch was checked out, and then the benchmark was run.  The results show most recent commits first, older commits at the bottom.  If you see a noticeable change in performance between one commit and the previous commit, then you know it made things faster/slower, and you can look at the changes within that specific commit.  We can choose the commits that were most beneficial and extract them into a separate pull request.

If you try to run these benchmarks yourself, you'll need to tweak `bench.sh` and install `hyperfine`.  When I come back to this, I'll add proper instructions.

Over two benchmark runs, these are my results.

The big improvements came from:
- [d5cbb312c4c8a83388acfb96f83a683a527d87fc](https://github.com/pythonarcade/arcade/commit/d5cbb312c4c8a83388acfb96f83a683a527d87fc)
- [055c42413039aa40c458e93663c60053f1dd66e9](https://github.com/pythonarcade/arcade/commit/055c42413039aa40c458e93663c60053f1dd66e9)

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `python -m benchmarks.collisions.bench d93ba54ace7feb49cd949856a28d4e88977b1265` | 17.273 ± 0.263 | 17.087 | 17.458 | 1.12 ± 0.02 |
| `python -m benchmarks.collisions.bench 055c42413039aa40c458e93663c60053f1dd66e9` | 15.485 ± 0.091 | 15.421 | 15.550 | 1.00 |
| `python -m benchmarks.collisions.bench 8d5d889a0f0eb9cfcf1d61122b902b61bfdc7d46` | 18.247 ± 0.281 | 18.048 | 18.445 | 1.18 ± 0.02 |
| `python -m benchmarks.collisions.bench 39633527f4a2d3ab8454a74dd284a89b9ceb2bef` | 18.573 ± 0.059 | 18.531 | 18.615 | 1.20 ± 0.01 |
| `python -m benchmarks.collisions.bench 68a5c838dad6afd5e996ee8ea3a7d2c96d243333` | 18.911 ± 0.214 | 18.759 | 19.063 | 1.22 ± 0.02 |
[...]

PR #1524, opened 2023-02-16
Collision detection optimizations, benchmark runner, and collision detection benchmark

cspotcode:
This is related to #1504.  Where the other PR is a draft which experimented with many ideas, this PR includes *only* the changes that had the biggest positive improvements.  It also adds the benchmark runner and collision benchmark.

I think this PR is suitable to be merged as-is.

Here are my benchmark results.  I ran it against each commit twice, I'm not sure if that was necessary but I was worried that my laptop would always run the first commit fastest because it was not hot (overheating) yet.  That may have been wrong or unnecessary.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[0066a6ae _check_for_collision: optimize broadphase](https://github.com/pythonarcade/arcade/commit/0066a6aed1618b42b16c38cd540cbbe345acb0a0) | 30.941 ± 0.864 | 30.330 | 31.552 | 1.01 ± 0.03 |
[6c023813 get_adjusted_hit_box: list comprehension into tuple is faster than generator into tuple](https://github.com/pythonarcade/arcade/commit/6c023813f282d8e6d81cbaeacdaf31164060db87) | 36.530 ± 0.386 | 36.257 | 36.803 | 1.19 ± 0.01 |
[5748dec0 get_adjusted_hit_box: inline rotate_point() call, avoid attribute lookups on self and math](https://github.com/pythonarcade/arcade/commit/5748dec038ea522ee57797fabab9838d484ac7ca) | 36.603 ± 0.262 | 36.418 | 36.789 | 1.19 ± 0.01 |
[0066a6ae _check_for_collision: optimize broadphase](https://github.com/pythonarcade/arcade/commit/0066a6aed1618b42b16c38cd540cbbe345acb0a0) | 30.664 ± 0.028 | 30.644 | 30.683 | 1.00 |
[6c023813 get_adjusted_hit_box: list comprehension into tuple is faster than generator into tuple](https://github.com/pythonarcade/arcade/commit/6c023813f282d8e6d81cbaeacdaf31164060db87) | 36.045 ± 0.347 | 35.800 | 36.291 | 1.18 ± 0.01 |
[5748dec0 get_adjusted_hit_box: inline rotate_point() call, avoid attribute lookups on self and math](https://github.com/pythonarcade/arcade/commit/5748dec038ea522ee57797fabab9838d484ac7ca) | 40.900 ± 5.103 | 37.291 | 44.508 | 1.33 ± 0.17 |

### DIFF
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -411,9 +411,11 @@ class Sprite:
             y *= scale_y
 
             # Rotate the point if needed
-            if radians:
-                x = x * rad_cos - y * rad_sin
-                y = x * rad_sin + y * rad_cos
+            if rad:
+                rot_x = x * rad_cos - y * rad_sin
+                rot_y = x * rad_sin + y * rad_cos
+                x = rot_x
+                y = rot_y
 
             # Apply position
             return (
--- a/arcade/sprite.py
+++ b/arcade/sprite.py
@@ -6,6 +6,7 @@ https://www.gamedev.net/articles/programming/general-and-gameplay-programming/sp
 """
 
 import math
+from math import sin, cos, radians
 import dataclasses
 from typing import (
     Any,
@@ -397,20 +398,27 @@ class Sprite:
         if self._point_list_cache is not None:
             return self._point_list_cache
 
+        rad = radians(self._angle)
+        scale_x, scale_y = self._scale
+        position_x, position_y = self._position
+        rad_cos = cos(rad)
+        rad_sin = sin(rad)
         def _adjust_point(point) -> Point:
+            x, y = point
+
+            # Apply scaling
+            x *= scale_x
+            y *= scale_y
+
             # Rotate the point if needed
-            if self._angle:
-                # Rotate with scaling to not distort it if scale x and y is different
-                point = rotate_point(point[0] * self._scale[0], point[1] * self._scale[1], 0, 0, self._angle)
-                # Apply position
-                return (
-                    point[0] + self._position[0],
-                    point[1] + self._position[1],
-                )
-            # Apply position and scale
+            if radians:
+                x = x * rad_cos - y * rad_sin
+                y = x * rad_sin + y * rad_cos
+
+            # Apply position
             return (
-                point[0] * self._scale[0] + self._position[0],
-                point[1] * self._scale[1] + self._position[1],
+                x + position_x,
+                y + position_y,
             )
 
         # Cache the results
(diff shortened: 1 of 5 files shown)


arcade/sprite.py
```diff
@@ -397,24 +398,33 @@ def get_adjusted_hit_box(self) -> PointList:
...
             return (
-                point[0] * self._scale[0] + self._position[0],
-                point[1] * self._scale[1] + self._position[1],
+                x + position_x,
+                y + position_y,
             )
 
         # Cache the results
-        self._point_list_cache = tuple(_adjust_point(point) for point in self.hit_box)
+        self._point_list_cache = tuple([_adjust_point(point) for point in self.hit_box])
```
einarf, line 427: can't the tuple just take the generator expression without paying for the list creation?

arcade/sprite.py
```diff
@@ -397,24 +398,33 @@ def get_adjusted_hit_box(self) -> PointList:
...
-                return (
-                    point[0] + self._position[0],
-                    point[1] + self._position[1],
-                )
-            # Apply position and scale
+            if rad:
+                rot_x = x * rad_cos - y * rad_sin
+                rot_y = x * rad_sin + y * rad_cos
+                x = rot_x
+                y = rot_y
```
einarf, line 418: Note that we are planning to add a pivot point / anchor to sprites.

arcade/sprite.py
cspotcode: Benchmarking showed that this is faster.  I think the problem is that generator expressions are slower than list comprehensions.  I also tried `*(),` syntax, but it was slower.

The fastest option was a plain list comprehension without converting to a tuple.  But this code is the second-fastest.

arcade/sprite.py
cspotcode: Ok, that makes sense.  If I'm understanding correctly, we will add the pivot point in a future PR?

pvcraven:
I think the speed-ups would be good. I'm inclined to merge those in.

I'm not sure about leaving the benchmarks directory in. 

* Either leave it in and add a readme documenting how it works, using it, sample output
* Or just drop it if documenting it is too burdensome.

arcade/sprite.py
pvcraven: Until we add the pivot point I'm cool with improving existing code. Just realize it might change soon.

einarf:
As discussed on discord we merge this as is. The contents in the benchmarking directory we can change to using standard cProfile using the build in profiler in arcade. Optionally just use simple timeit. Instead of bringing in large examples we can test individual functions only.

PR #1527, opened 2023-02-17
Allow adding empty sprite lists to a scene

ccntrq:
Checking for existence of the list with `not sprite_list` causes a new list to be created if the caller passes an empty sprite list. The new check with `sprite_list is None` ensures that the given list is used, even if empty.


pvcraven:
Thanks!

PR #1535, opened 2023-02-17
Benchmark with and without shapely

cspotcode:
I'm creating a PR to ask someone else to run this benchmark, too.  Because I'm confused by the results.

I'm running on Windows 11, python 3.11.2:
```python
& 'C:\Program Files\Git\bin\bash.exe' ./benchmarks/bench-shapely.sh
```

But the important bit is really that it's running `benchmarks/collisions/bench.py` with shapely enabled and then disabled.  I enable and disable using an environment variable.  In both cases, shapely is imported -- we pay the startup cost -- but it's only used for collision detection in one of the benchmark cases.

Here are my surprising results:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bash ./benchmarks/bench-shapely-helper.sh enabled` | 90.945 ± 2.481 | 89.190 | 92.699 | 3.54 ± 0.10 |
| `bash ./benchmarks/bench-shapely-helper.sh disabled` | 25.715 ± 0.029 | 25.694 | 25.736 | 1.00 |

### DIFF
--- a/arcade/geometry.py
+++ b/arcade/geometry.py
@@ -1,3 +1,5 @@
+import os
+
 shapely_exists = False
 
 try:
@@ -6,8 +8,10 @@ try:
 except ImportError:
     pass
 
+use_shapely = shapely_exists and not os.environ.get("DISABLE_SHAPELY")
+print("Use shapely: " + str(use_shapely))
 
-if shapely_exists:
+if use_shapely:
     from .geometry_shapely import are_polygons_intersecting  # noqa: F401
     from .geometry_shapely import is_point_in_polygon  # noqa: F401
 else:
--- a/arcade/paths.py
+++ b/arcade/paths.py
@@ -2,11 +2,20 @@
 Classic A-star algorithm for path finding.
 """
 import sys
+import os
 from arcade.types import Point
 from arcade import check_for_collision_with_list, SpriteList, Sprite
 from typing import Union, List, Tuple, Set, Optional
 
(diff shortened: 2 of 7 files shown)


cspotcode:
Closing to keep the PR queue tidy.  This PR is linked in #1536

cspotcode:
I added a line-of-sight benchmark.

On Windows you can run it like this, assuming hyperfine and git are installed.  Or you can run it via `bench-shapely-helper.sh` directly, or using `timeit`.

```
& 'C:\Program Files\Git\bin\bash.exe' .\benchmarks\bench-shapely.sh line-of-sight
```

Here are my results:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bash ./benchmarks/bench-shapely-helper.sh line-of-sight enabled` | 86.526 ± 1.641 | 85.366 | 87.686 | 1.00 |
| `bash ./benchmarks/bench-shapely-helper.sh line-of-sight disabled` | 89.144 ± 0.628 | 88.700 | 89.588 | 1.03 ± 0.02 |

Cleptomania:
We have decided to remove Shapely entirely with #1646. Much of the pure python implementations were faster or only slightly better with shapely, and we are providing better acceleration options for that functionality via `arcade-accelerate` coming up with Arcade 3.0.

PR #1536, opened 2023-02-17
Disable shapely for `are_polygons_intersecting` and `is_point_in_polygon`

cspotcode:
We benchmarked shapely in #1535 and found that it's way slower when used in these two functions.  We discussed on Discord that we can disable shapely now, and leave the code in-place for someone to investigate in the future.

This does not disable using shapely in other parts of arcade, such as pathfinding.  Shapely may still be faster in those other areas, so we should not disable it.

### DIFF
--- a/arcade/geometry/__init__.py
+++ b/arcade/geometry/__init__.py
@@ -8,19 +8,27 @@ from .geometry_python import (
     are_lines_intersecting,
     is_point_in_box,
 )
-try:
+
+# Shapely is disabled due to https://github.com/pythonarcade/arcade/pull/1535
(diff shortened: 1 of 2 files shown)


einarf:
Geometry module was moved, so might as well redo this one. It's a small thing anyway

cspotcode:
Rebased onto the latest `development`, should be good to go now.

einarf:
Might want to check if we're mentioning shapely anywhere in docs or other places in the codebase. We don't want to cause confusion.

cspotcode:
Good call, searching the codebase shows it's not mentioned in docs, only in the changelog for 3.0.0.

Issue #1137, opened 2022-03-27
Particles don't care about frame time


einarf:
On some systems particles are :

* Not moving smoothly
* Have a longer or shorter lifespan based on framerate
* Fades faster or slower based on framerate

Looking at this briefly it's clearly realted to `update()` methods in sprite and spritelist.

Frametime was probable hardcoded because update don't have frametime parameter:

https://github.com/pythonarcade/arcade/blob/a0213acc263dd3d89da9ffb4abd19f3d7e43e79a/arcade/particle.py#L101-L104

We do have an `on_update` method now that takes a frame time. The whole `update` vs `on_update` siutation should probably be resolved in the next major version of arcade. For now I think moving the logic to `on_update` and calling that from `update` with hardcoded delta time is a possible fix. The `Emitter` should be changed to call `on_update` and we need to somehow find a delta time to pass in without breaking anything.

pushfoo:
This is still a valid ticket as of 2023.

einarf:
Yes. It's a general frame time issue in arcade. I put it on 3.0 milestone for a reason. We need to solve this before release I think. It's very imporant.

Arcade currently works with ticks assuming 60fps. This is how sprite and sprite velocity also works now. We should make a thread about this on discord and figure something out.

DragonMoffon:
could be fixed with issue #1795.

einarf:
Fixed in https://github.com/pythonarcade/arcade/pull/2288

PR #1543, opened 2023-02-18
Clamp delta time in debugger, or other ways to control delta_time

cspotcode:
There are a few different cases where it might be helpful to clamp or control `delta_time`.

## Debugger

When pausing in the debugger, the next frame's `delta_time` will be massive.  Makes sense, if you pause the debugger for a minute, then the delta is 60s.
But engines like Unity clamp delta time for this reason.  When you pause and then resume the game in Unity, it will not generate a massive `delta_time`.
Should we add something like that to arcade?

It could be as simple as `my_window.set_max_delta_time(1/30)` or `set_fixed_delta_time` or maybe there's something more auto-magic that would allow detecting when the debugger has paused? `if sys.gettrace() is not None`

## AI

I can't speak to this one, but people want to use arcade for AI simulation or visualization.  In these cases, I think the simulation and render loop should run in fast-forward, as fast as the CPU can handle.  We can update docs to explain how to do this easily.  Maybe we explain how to set a fixed `delta_time` but for arcade to tick as fast as possible, no sleeping.

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -4,6 +4,7 @@ derive from.
 """
 import logging
 import os
+import sys
 import time
 from typing import Tuple, Optional
 
@@ -31,6 +32,7 @@ MOUSE_BUTTON_RIGHT = 4
 
 _window: 'Window'
 
+_in_debugger = sys.gettrace() is not None
 
 def get_screens():
     """
@@ -323,6 +325,10 @@ class Window(pyglet.window.Window):
         Internal function that is scheduled with Pyglet's clock, this function gets run by the clock, and
         dispatches the on_update events.
         """
+        # Clamp delta_time in debugger, since long debugger pauses should not
+        # cause sky-high `delta_time` on the next frame.
+        if _in_debugger:
+            delta_time = min(delta_time, self._update_rate * 2.)
         self.dispatch_event('on_update', delta_time)
 
     def set_update_rate(self, rate: float):


gran4:
Maybe the AI section could be a separate issue?

pushfoo:
**TL;DR: The root issue is still the same: customizing frame / event loops**

@FengchiW  Since you mentioned being interested in dealing with #1137, i'm tagging you here.

> But engines like Unity clamp delta time for this reason. When you pause and then resume the game in Unity, it will not generate a massive delta_time.

Arcade (and pyglet) having poor delineations between system time and game time contexts have hurt us before. One example is a crash when debugging under PulseAudio (#1900) due to an issue with upstream (https://github.com/pyglet/pyglet/issues/952).

What I know at the moment:

* @DragonMoffon started on some experimental clock-related work over in [`arcade/experimental/clock`](https://github.com/pythonarcade/arcade/tree/development/arcade/experimental/clock), but it's in a very rough state.
* There's some capacity for custom event loops in pyglet which I haven't had time to fully investigate yet

DragonMoffon:
They don't do any clamping, they don't call the updated functions of inactive scene elements.

DragonMoffon:
We have the new Clock API now, i will just add the ability to set a clamp on the clock's  delta_time

that way, it needs to check only once for debug and can be changed at run time (in response to a lag spike, for example)

PR #1560, opened 2023-02-20
Fix #1558 - avoid imports to be marked as private

### LINKED ISSUE #1558: Avoid marking non-aliased imports as private
## Bug Report
It is standard (see PEP 484 stub file section) to mark redundantly
imported items as exported. 

Pyright users currently see a linting error when using arcade.color and other imports from arcade in the global namespace.

## System Info

Run this and paste the output here: python -m arcade

Arcade 3.0.0.dev13
------------------
vendor: Intel
renderer: Intel(R) Iris(R) Xe Graphics
version: (3, 3)
python: 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
platform: win32
pyglet version: 2.0.4
PIL version: 9.4.0

### Actual behavior:
Pyright displays an error 
image

### Expected behavior:
No linting for these symbols

any use of arcade.color, arcade.key, arcade.experimental etc. while using pyright

Imports from the arcade __init__ file should be redundantly aliased as themselves
`from arcade import color as color` etc

### What would it help with?
Linting and static type checking for pyright users.

MrWardKKHS:
It is standard (see PEP 484 stub file section) to mark redundantly imported items as exported. This PR adds this redundant exporting to avoid accidentally indicating they should be private to type checkers.

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -220,14 +220,14 @@ from .perf_info import disable_timings
 from .perf_graph import PerfGraph
 
 # Module imports
-from arcade import color
-from arcade import csscolor
-from arcade import key
-from arcade import resources
-from arcade import types
-from arcade import math
-from arcade import shape_list
-from arcade import experimental
+from arcade import color as color
+from arcade import csscolor as csscolor
+from arcade import key as key
+from arcade import resources as resources
+from arcade import types as types
+from arcade import math as math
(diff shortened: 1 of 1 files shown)


einarf:
Thanks!

PR #1569, opened 2023-02-22
Adding a random.glsl for use with glsl code injection

DragonMoffon:
Adding a random.glsl for use with glsl code injection, which can be included using the common argument of `ArcadeContext,load_program` and `Context.program`.

Created so all who want to do stuff with random white noise inside shaders can.  This noise is focused on generating visually good-looking noise, and there are no assurances on the mathematical quality.

(NOTE: it is pseudo-random noise).


arcade/examples/gl/using_random_glsl.py
```diff
@@ -0,0 +1,100 @@
+from time import time
+from array import array
+from random import random
+
+from arcade import Window
+from arcade.gl import BufferDescription
+from arcade.resources import resolve_resource_path
+
+"""
```
pushfoo: This docstring needs to be moved to the top of the file before imports for CI to pick it up.

pushfoo (changes requested):
Aside from needing to move the docstring to the top of the file, this looks good.

PR #1572, opened 2023-02-22
Remove typing.cast() in hot code path

cspotcode:
Since this is such a hot code path, avoiding the cost of a function call gives a noticeable improvement in performance.


einarf:
With the updated type annotations we did earlier to this module this should no longer be needed.

einarf:
The cast is only cosmetic for the type checked, so we might as well do this.

PR #1573, opened 2023-02-23
Switch to pyproject.toml and Ruff for PEP 621/660

Cleptomania:
PEP [621](https://peps.python.org/pep-0621/)/[660](https://peps.python.org/pep-0660/) describes the switch to pyproject.toml for centralized project metadata, configuration, and build systems.

This PR removes `setup.cfg` and `setup.py` in favor of `pyproject.toml`.

One problem with completely consolidating everything in `pyproject.toml` is flake8's lack of support for the format. To remedy this, this PR also replaces flake8 with ruff. They are largely compatible, and most everyone shouldn't really notice much of a difference in using them. If we wanted to continue using flake8 instead, we would either need to keep the `setup.cfg` with the flake8 configuration, or add a `.flake8` config file.


pyproject.toml
```diff
@@ -1,7 +1,140 @@
+[project]
+name = "arcade"
+description = "Arcade Game Development Library"
+readme = "README.rst"
+authors = [
+    {name="Paul Vincent Craven", email="paul.craven@simpson.edu"}
```
pvcraven: That e-mail isn't valid anymore. Use paul@cravenfamily.com

PR #1578, opened 2023-02-24
Gui/fix labels

eruvanos:


eruvanos:
Windows worker renders text differently and so the sizes are different -.-
I guess I can make the tests more fuzzy to just check for a range.

PR #1596, opened 2023-02-28
Add intro text to shader tutorial index

pushfoo:
This is preparation for the improved SpriteList doc. This page will be linked as a better alternative to techniques mentioned in new the SpriteList doc.


einarf:
I'm not sure if this is the right place, but at least it's a start!

pushfoo:
We'll definitely move or expand upon the intro text in future commits once we get the style guidelines ironed out. :+1:

PR #1598, opened 2023-03-02
Optimization: only rebuild the spatial hash when necessary

pushfoo:
This PR separates the changes below from the larger SpriteList doc rewrite:
* Add int type annotation to function signature
* Make the spatial hash recreate only when it doesn't exist or has different cell size

### DIFF
--- a/arcade/sprite_list/sprite_list.py
+++ b/arcade/sprite_list/sprite_list.py
@@ -790,13 +790,15 @@ class SpriteList(Generic[SpriteType]):
         """
         self.spatial_hash = None
 
-    def enable_spatial_hashing(self, spatial_hash_cell_size=128):
+    def enable_spatial_hashing(self, spatial_hash_cell_size: int = 128):
         """Turn on spatial hashing."""
-        LOG.debug("Enable spatial hashing with cell size %s", spatial_hash_cell_size)
-        from .spatial_hash import SpatialHash
-
-        self.spatial_hash = SpatialHash(cell_size=spatial_hash_cell_size)
-        self._recalculate_spatial_hashes()
+        if self.spatial_hash is None or self.spatial_hash.cell_size != spatial_hash_cell_size:
+            LOG.debug("Enabled spatial hashing with cell size %s", spatial_hash_cell_size)
+            from .spatial_hash import SpatialHash
+            self.spatial_hash = SpatialHash(cell_size=spatial_hash_cell_size)
+            self._recalculate_spatial_hashes()
+        else:
+            LOG.debug("Spatial hashing is already enabled with size %s", spatial_hash_cell_size)
 
     def _recalculate_spatial_hashes(self):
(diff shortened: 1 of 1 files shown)


arcade/sprite_list/sprite_list.py
```diff
@@ -790,13 +790,15 @@ def disable_spatial_hashing(self) -> None:
         """
         self.spatial_hash = None
 
-    def enable_spatial_hashing(self, spatial_hash_cell_size=128):
+    def enable_spatial_hashing(self, spatial_hash_cell_size: int = 128):
         """Turn on spatial hashing."""
```
einarf, line 794: Docsting should probalby be improved here? Parameters. What happens if you already have a spatial hash?

arcade/sprite_list/sprite_list.py
```diff
@@ -790,13 +790,15 @@ def disable_spatial_hashing(self) -> None:
...
-        self._recalculate_spatial_hashes()
+        if self.spatial_hash is None or self.spatial_hash.cell_size != spatial_hash_cell_size:
+            LOG.debug("Enabled spatial hashing with cell size %s", spatial_hash_cell_size)
+            from .spatial_hash import SpatialHash
+            self.spatial_hash = SpatialHash(cell_size=spatial_hash_cell_size)
+            self._recalculate_spatial_hashes()
+        else:
+            LOG.debug("Spatial hashing is already enabled with size %s", spatial_hash_cell_size)
 
     def _recalculate_spatial_hashes(self):
```
einarf, line 803: If enabling spatial hash no longer recalculates them, this method should probably be public with some docstring? or the enable method can take a `recalculate` bool param that is default true? Are we using any of these methods in arcade currently?

arcade/sprite_list/sprite_list.py
pushfoo: My intent was to do the docstring in a subsequent PR since I'm working on separating the Programming Guide content out of them.

arcade/sprite_list/sprite_list.py
pushfoo: > with some docstring

As per my other response, I wanted to add all the comments in one PR, and then discuss an API rework around hashing separately.

>  Are we using any of these methods in arcade currently?

Only in unit tests.

> or the enable method can take a recalculate bool param that is default true?

This is why I kept this PR simple. There are a lot of questions raised by trying to polish this further, such as whether all spatial hash objects will really use cell size as their underlying concepts.

Imo, the stable core of the spatial hash API is the following:
1. add
2. remove
3. reset
4. move

The other methods may need to be rethought to accommodate different underlying implementations. For example, someone could conceivably write a spatial hash backed by quad trees or another technique.

pushfoo: I also think the better approach here is to move the recalculation method onto the SpatialHash itself.

PR #1604, opened 2023-03-04
GPU Collision Experiment

einarf:
Experiment supporting GPU collision checking in the Spritelist. This has nothing to do with the spatial hash. Ignore the branch name.

* The idea is to write the original untransformed hit box into a floating point texture.
* The shader will have access to all the sprite properties for free and can (on the fly) transform the hit box points and do polygon intersection (really just line intersection checks).
* Running this collision check is of course expensive, but it completely eliminates the need for sprites to calculate adjusted hit box points and dealing with a spatial hash.
* Spritelists with this collision type will have a much lower capacity. Possibly 1024 as a cap.
* Another challenge is that the user needs to pre-determine the max number of hit box points the sprites have. We can assume 4-8 points if the simple hit box algo is used for now.
* This collision type is better suited for spritelist vs spritelist collision (collision with self or other list)

### DIFF
--- a/arcade/experimental/spritelist_gpu_collision.py
+++ b/arcade/experimental/spritelist_gpu_collision.py
@@ -1,0 +1,45 @@
+import arcade
+from pyglet import gl
+
+coin_path = ":resources:images/items/coinGold.png"
+
+class Test(arcade.Window):
+    def __init__(self):
+        super().__init__(800, 600)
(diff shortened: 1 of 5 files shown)


einarf:
Will rremporarily retire this experiment in favor of the rust acceleration. We can experiment more with this later in 3.0.

PR #1609, opened 2023-03-04
Add test for column-row-span with spacing

Ibrahim2750mi:


Ibrahim2750mi:
No changes were made to `layout.py` it was initially made because I calculated wrong assertion value.

einarf:
Merging this since I'm refactoring tests

Issue #1493, opened 2023-02-02
Camera scrolls in screenspace as opposed to projection-space [2.7.1.dev12]


DigiDuncan:
## System Info
```
Arcade 2.7.1.dev12
------------------
vendor: NVIDIA Corporation
renderer: NVIDIA GeForce GTX 1080/PCIe/SSE2
version: (3, 3)
python: 3.11.0 | packaged by conda-forge | (main, Jan 16 2023, 14:12:30) [MSC v.1916 64 bit (AMD64)]
platform: win32
```

### Actual behavior:
All cameras move at the same speed when give the same `move()` command, despite their differing zoom levels.

### Expected behavior:
Cameras `move()` amount is scaled relative to it's current `scale`. (as it was pre-2.7)

### Steps to reproduce/example code:
https://gist.github.com/DigiDuncan/2f194def17db0c1d471f69da9179ee9a

einarf:
@janscas if you wanted to take a look

einarf:
The multiplication order was reversed in the new pyglet version. My hunch is that translation is multiplied wrong.

alejcas:
Didn’t have the time today. I’ll try tomorrow

alejcas:
Pyglet version?

DigiDuncan:
> Pyglet version?

2.0.3

alejcas:
@DigiDuncan can you provide any example in the arcade examples that is also not working? I can't run your code.

Error: ' 'ParallaxView' object has no attribute 'camera'

alejcas:
Also I think parallax.py example works as expected...

alejcas:
Besides what @einarf explained about pyglet changing the matrix multiplication orden, the last camera PR I submitted changed the way zoom works.

Maybe I broke it....

Before, the zoom was modifying the right and top of the projection matrix and also the view matrix by just setting a param called zoom.

**The problem with this is that changing the projection to view ratio will also produce a zoom. So setting the projection to view ratio to something like 0.5 (the projection is 1/2 smaller than the viewport) and also setting a zoom of 0.5 WILL NOT output a zoom of 0.5 but a much larger zoom... but the user was setting zoom=0.5 producing NOT the desired effect **

So I tried to correlate scale (effectively zoom) and the projection to view ratio so changing one or the other will output the same zoom effect.

The before code looked like this:

```python
# simplified code...
def _set_projection_matrix():
    right *= zoom
    top *= zoom
[...]

PR #1626, opened 2023-03-07
Camera rollback

alejcas:
This just rollback the camera implementation.

Zoom works as expected but is disconnected from the projection to view ratio.

Solves issue #1493 

Note that SimpleCamera no longer have zoom (as was planned initialy)


einarf:
Thanks. I will be looking at the cameras again soon. I think we can refactor them a bit since we are now doing 3.0. The design is based on the previous versions and that is fairly limiting.

PR #1639, opened 2023-03-15
Refactor color handling into a new Color type

pushfoo:
- [x] Add backwards-compatible `Color` tuple subclass
- [x] Convert `arcade.color` and `arcade.csscolor` modules to use the new type
- [x] Port float normalization behavior to a property & deleted redundant `get_three_three_float_color` function
- [x] Port creation from uint24 values to a class method
- [x] Port creation from uint32 values to a class method
- [x] Port creation from normalized floats to a class method
- [x] Port creation from hex strings to a class method
- [x] Consolidate typing around the code base
- [x]  Replace `get_four_byte_color` with `Color.from_iterable`
    - [x] `arcade.draw_commands`
    - [x] `arcade.perf_graph`
    - [x] `arcade.shape_list`
    - [x] `arcade.text`
    - [x] `arcade.examples`
    - [x] `arcade.gui.widgets`
    - [x] `arcade.sprite`
    - [x] `doc/tutorials`
    - [x] `tests/`
- [X] Fix any lingering linting issues


arcade/types.py
```diff
@@ -14,15 +14,255 @@
...
+MAX_UINT32 = 0xFFFFFFFF
+
+
 RGB = Tuple[int, int, int]
 RGBA = Tuple[int, int, int, int]
-Color = Union[RGB, RGBA]
+
+
+# Color = Union[RGB, RGBA]
+class Color(tuple):
```
Ibrahim2750mi: Is there a problem it being a NamedTuple? Might make things more informative when using it.

pushfoo: tl;dr I didn't see a good way to do validation when subclassing `NamedTuple`, and `Color(*iterable)` was intended to replace `get_four_byte_color`

I may be wrong, but to my understanding:
1. Direct subclassing of `NamedTuple` generates `__new__` and `__init__` methods which preclude boundary checking efficiently:
```console
ImportError while loading conftest '/home/user/src/arcade/tests/conftest.py'.
tests/conftest.py:3: in <module>
    import arcade
arcade/__init__.py:69: in <module>
    from .drawing_support import get_four_byte_color
arcade/drawing_support.py:8: in <module>
    from arcade.types import Color, RGBA, ColorLike
arcade/types.py:36: in <module>
    class Color(NamedTuple):
/usr/lib/python3.9/typing.py:1836: in __new__
    raise AttributeError("Cannot overwrite NamedTuple attribute " + key)
E   AttributeError: Cannot overwrite NamedTuple attribute __new__
```
2. We can subclass with generic subscripts (`Color(tuple[int, int, int, int])`) which may be friendlier to type checkers

If one or both of these are wrong, we might be able to drop the use of `# type: ignore` in `__new__` as a workaround for https://github.com/python/mypy/issues/8541.

einarf: `NamedTuple` is also very slow

arcade/experimental/lights.py
```diff
@@ -15,7 +16,7 @@ def __init__(
         center_x: float,
         center_y: float,
         radius: float = 50.0,
-        color: Color = (255, 255, 255),
+        color: RGBOrA255 = WHITE,
```
einarf, line 19: Should just be `RGBA`?

arcade/experimental/lights.py
```diff
@@ -158,7 +159,7 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self._prev_target.use()
 
-    def draw(self, position: Tuple[float, float] = (0, 0), target=None, ambient_color: Color = (64, 64, 64)):
+    def draw(self, position: Tuple[float, float] = (0, 0), target=None, ambient_color: RGBOrA255 = (64, 64, 64)):
```
einarf, line 162: `RGBA255`

arcade/experimental/lights.py
pushfoo: To my understanding, passing to the GL shader discards the alpha channel:

```py
        self._combine_program['ambient'] = ambient_color[:3]
```
I can make it RGBA if you'd like, but this is an advanced example. A nudge toward RGBA may annoy advanced users here.

arcade/experimental/lights.py
pushfoo: Same concern as the prior comment: the lights demo seems advanced enough that nudging users toward RGBA might annoy users here.

PR #1650, opened 2023-03-19
Deprecate `lrtb` (left, right, top, bottom)

Ibrahim2750mi:
Relevant discussion: https://discord.com/channels/458662222697070613/1087007892575617086/1087015976157073428

Mentions for the world `lrtb` in the whole arcade package after deprecation: 
```shell
$ grep -nr -I --exclude-dir='.[^.]*' 'lrtb' .
./arcade/draw_commands.py:622:    message="Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_outline.",
./arcade/draw_commands.py:625:def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
./arcade/draw_commands.py:744:    message="Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
./arcade/draw_commands.py:747:def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
./arcade/__init__.py:128:from .draw_commands import draw_lrtb_rectangle_filled
./arcade/__init__.py:130:from .draw_commands import draw_lrtb_rectangle_outline
./arcade/__init__.py:305:    'draw_lrtb_rectangle_filled',
./arcade/__init__.py:307:    'draw_lrtb_rectangle_outline',
```

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -620,7 +620,7 @@ def draw_triangle_outline(x1: float, y1: float,
 
 @warning(
     message=\
-        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_outline.",
+        "`draw_lrtb_rectangle_outline` is deprecated. Use `draw_lrbt_rectangle_outline` instead.",
     warning_type=NameChangeWarning,
 )
 def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
@@ -743,7 +743,7 @@ def draw_rectangle_outline(center_x: float, center_y: float, width: float,
 
 @warning(
     message=\
-        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
+        "`draw_lrtb_rectangle_filled` is deprecated. Use `draw_lrbt_rectangle_filled` instead.",
     warning_type =NameChangeWarning,
 )
 def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -619,8 +619,9 @@ def draw_triangle_outline(x1: float, y1: float,
 
 
 @warning(
-    message="Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_outline.",
-    warning_type=NameChangeWarning
+    message=\
+        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_outline.",
+    warning_type=NameChangeWarning,
 )
 def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
                                 bottom: float, color: Color,
@@ -741,8 +742,9 @@ def draw_rectangle_outline(center_x: float, center_y: float, width: float,
 
 
 @warning(
-    message="Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
-    warning_type=NameChangeWarning
+    message=\
+        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
+    warning_type =NameChangeWarning,
 )
 def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
                                bottom: float, color: Color):
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -753,7 +753,7 @@ def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
     Draw a rectangle by specifying left, right, top and bottom edges.
 
     .. deprecated:: 3.0
-        Use :py:func:`draw_lrbt_rectangle_filled` instead!
+       Use :py:func:`draw_lrbt_rectangle_filled` instead!
 
     :param float left: The x coordinate of the left edge of the rectangle.
     :param float right: The x coordinate of the right edge of the rectangle.
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -656,12 +656,12 @@ def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
 def draw_lrbt_rectangle_outline(left: float, right: float, bottom: float, top: float, color: Color,
                                 border_width: float = 1):
     """
-    Draw a rectangle by specifying left, right, top, and bottom edges.
+    Draw a rectangle by specifying left, right, bottom, and top edges.
 
     :param float left: The x coordinate of the left edge of the rectangle.
     :param float right: The x coordinate of the right edge of the rectangle.
-    :param float top: The y coordinate of the top of the rectangle.
     :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
     :param Color color: The color of the rectangle.
     :param float border_width: The width of the border in pixels. Defaults to one.
     :Raises AttributeError: Raised if left > right or top < bottom.
@@ -771,12 +771,12 @@ def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
 
 def draw_lrbt_rectangle_filled(left: float, right: float, bottom: float, top: float, color: Color):
     """
-    Draw a rectangle by specifying left, right, top, and bottom edges.
+    Draw a rectangle by specifying left, right, bottom, and top edges.
 
     :param float left: The x coordinate of the left edge of the rectangle.
     :param float right: The x coordinate of the right edge of the rectangle.
-    :param float top: The y coordinate of the top of the rectangle.
     :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
     :param Color color: The color of the rectangle.
     :Raises AttributeError: Raised if left > right or top < bottom.
     """
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -25,7 +25,7 @@ from arcade import (
     Texture,
     get_window,
 )
-from arcade.utils import warning, NameChangeWarning
+from arcade.utils import warning, ReplacementWarning
 
 # --- BEGIN ARC FUNCTIONS # # #
 
@@ -619,7 +619,7 @@ def draw_triangle_outline(x1: float, y1: float,
 
 
 @warning(
-    warning_type=NameChangeWarning,
+    warning_type=ReplacementWarning,
     new_name="draw_lrbt_rectangle_outline"
 )
 def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
@@ -637,15 +637,15 @@ def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
     :param float bottom: The y coordinate of the rectangle bottom.
     :param Color color: The color of the rectangle.
     :param float border_width: The width of the border in pixels. Defaults to one.
-    :Raises AttributeError: Raised if left > right or top < bottom.
+    :Raises ValueError: Raised if left > right or top < bottom.
 
     """
     if left > right:
-        raise AttributeError("Left coordinate must be less than or equal to "
+        raise ValueError("Left coordinate must be less than or equal to "
                              "the right coordinate")
 
     if bottom > top:
-        raise AttributeError("Bottom coordinate must be less than or equal to "
+        raise ValueError("Bottom coordinate must be less than or equal to "
                              "the top coordinate")
 
     center_x = (left + right) / 2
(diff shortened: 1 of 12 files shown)


arcade/draw_commands.py
```diff
@@ -617,11 +618,16 @@ def draw_triangle_outline(x1: float, y1: float,
 # --- BEGIN RECTANGLE FUNCTIONS # # #
 
 
+@warning(
+    message=\
+        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_outline.",
+    warning_type=NameChangeWarning,
+)
 def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
```
einarf: Maybe make the message a bit shorter?
`draw_lrtb_rectangle_outline is deprecated. Use draw_lrbt_rectangle_outline instead.`

arcade/draw_commands.py
```diff
@@ -706,10 +741,15 @@ def draw_rectangle_outline(center_x: float, center_y: float, width: float,
     _generic_draw_line_strip(point_list, color, gl.GL_TRIANGLE_STRIP)
 
 
+@warning(
+    message=\
+        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
+    warning_type =NameChangeWarning,
+)
```
einarf: Same applies here. Somewhat shorter and more explicit message is probably good.

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/commit/f421cb9a1ed17d82a3c817eaac4fb60fca2a223e

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/commit/f421cb9a1ed17d82a3c817eaac4fb60fca2a223e

arcade/draw_commands.py
```diff
@@ -743,7 +743,7 @@ def draw_rectangle_outline(center_x: float, center_y: float, width: float,
 
 @warning(
     message=\
-        "Drawing with lrtb(left, right, top, bottom) is deprecated use lrbt instead i.e draw_lrbt_rectangle_filled.",
+        "`draw_lrtb_rectangle_filled` is deprecated. Use `draw_lrbt_rectangle_filled` instead.",
```
pushfoo: To my understanding, the enclosing backticks are only processed in Sphinx RST, not in ordinary terminal output. Does the warning decorator render them somehow?

Ibrahim2750mi: Doesn't matter if they do or don't do, it still kind of shows that we are mentioning about a specific function.

Ibrahim2750mi: Speaking of sphinx shall I add the warning in the docstring too?

einarf: Probably not a bad idea

pushfoo: Yes, thank you :+1: . A [deperecation directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-warning) would probably do, something like the following:

```py
def deprecated_func(param_name):
    """
    Short summary line

    .. deprecated:: 3.0
                    Use :py:func:`better_func_name` instead!
 
    Additional description, if any.
 
    :param param_name:
    """
```

einarf: Back to the decorator itself. I do agree backticks are not needed in terminal output and it can probably be expressed cleaner like this?
```py
@warning(
    message=((
        "draw_lrtb_rectangle_filled is deprecated. "
        "Use draw_lrbt_rectangle_filled instead."
    )),
    warning_type=NameChangeWarning,
)
```

Ibrahim2750mi: OK! Will do these after lunch.

einarf: Could also be a better way to design the warning system so you could pass "old_name" and "new_name" somewhere that would generate a generic message from that, but I'm not too familiar with this. It's at least good enough for now and it can easily be tweaked in the future.

Ibrahim2750mi: Will also do that.

einarf: The decorator also has the function reference, so the name can be obtained from that meaning you only need one argument for the new function that should be used instead?

Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/d9cdfbe7fe31b201acdedf9f8bdf256abb005bbd

arcade/draw_commands.py
```diff
@@ -617,11 +618,18 @@ def draw_triangle_outline(x1: float, y1: float,
...
+)
 def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
                                 bottom: float, color: Color,
                                 border_width: float = 1):
     """
-    Draw a rectangle by specifying left, right, top, and bottom edges.
+    Draw a rectangle by specifying left, right, top and bottom edges.
+
+    .. deprecated:: 3.0
+        Use :py:func:`draw_lrbt_rectangle_outline` instead!
```
pushfoo: Although this may build without perfect alignment, it's a good habit to be precise. Sphinx is picky about whitespace like Python is, just slightly less so.
```suggestion
       Use :py:func:`draw_lrbt_rectangle_outline` instead!
```

arcade/draw_commands.py
```diff
@@ -706,10 +743,17 @@ def draw_rectangle_outline(center_x: float, center_y: float, width: float,
...
+    new_name="draw_lrbt_rectangle_filled"
+)
 def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
                                bottom: float, color: Color):
     """
-    Draw a rectangle by specifying left, right, top, and bottom edges.
+    Draw a rectangle by specifying left, right, top and bottom edges.
+
+    .. deprecated:: 3.0
+        Use :py:func:`draw_lrbt_rectangle_filled` instead!
```
pushfoo: Same as the previous comment.
```suggestion
       Use :py:func:`draw_lrbt_rectangle_filled` instead!
```

arcade/utils.py
```diff
@@ -17,8 +17,16 @@ class PerformanceWarning(Warning):
     pass
 
 
-def warning(message: str, warning_type: Type[Warning]):
+class NameChangeWarning(Warning):
```
pushfoo: `NameChangeWarning` may be inappropriate for this PR since the new functions will have different argument order.

 [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) already exists as a built-in, but comes with some restrictions on filtering. 
 
 If we can't or don't want to use this builtin, maybe `ReplacementWarning` would be better?

pushfoo (changes requested):
Looks pretty good. I'll spare you formatting minute and request 3 changes which seem to be actually important.

arcade/utils.py
Ibrahim2750mi: As you said the built-in one gets filtered, thats why a custom warning. I don't think `ReplacementWarning` kind of sounds right, like it kind of means that we have changed the functionality too, I guess? Although if you, I or anyone can't think of a better name we will kind of go with `ReplacementWarning` or just won't change name.

pushfoo: > like it kind of means that we have changed the functionality too

I think we have since the argument order is different and passing the old one will raise exceptions.

arcade/draw_commands.py
```diff
@@ -731,6 +775,30 @@ def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
...
+
+    :param float left: The x coordinate of the left edge of the rectangle.
+    :param float right: The x coordinate of the right edge of the rectangle.
+    :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
+    :param Color color: The color of the rectangle.
+    :Raises AttributeError: Raised if left > right or top < bottom.
+    """
+    if left > right:
+        raise AttributeError(f"Left coordinate {left} must be less than or equal to the right coordinate {right}")
```
pushfoo: These should be `ValueError`s in the new functions since we're breaking compatibility anyway.

arcade/utils.py
Ibrahim2750mi: Oh right 👍🏽

arcade/draw_commands.py
```diff
@@ -731,6 +775,30 @@ def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
...
+
+    if bottom > top:
+        raise AttributeError(f"Bottom coordinate {bottom} must be less than or equal to the top coordinate {top}")
```
pushfoo: Ditto here, `s/Attribute/Value`.

arcade/draw_commands.py
```diff
@@ -632,7 +640,36 @@ def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
...
+    :param float right: The x coordinate of the right edge of the rectangle.
+    :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
+    :param Color color: The color of the rectangle.
+    :param float border_width: The width of the border in pixels. Defaults to one.
+    :Raises AttributeError: Raised if left > right or top < bottom.
+
+    """
     if left > right:
         raise AttributeError("Left coordinate must be less than or equal to "
```
pushfoo: `s/Attribute/Value/`

arcade/draw_commands.py
```diff
@@ -731,6 +775,30 @@ def draw_lrtb_rectangle_filled(left: float, right: float, top: float,
...
+def draw_lrbt_rectangle_filled(left: float, right: float, bottom: float, top: float, color: Color):
+    """
+    Draw a rectangle by specifying left, right, bottom and top edges.
+
+    :param float left: The x coordinate of the left edge of the rectangle.
+    :param float right: The x coordinate of the right edge of the rectangle.
+    :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
+    :param Color color: The color of the rectangle.
+    :Raises AttributeError: Raised if left > right or top < bottom.
```
pushfoo: `s/Attribute/Value/`

arcade/draw_commands.py
```diff
@@ -632,7 +640,36 @@ def draw_lrtb_rectangle_outline(left: float, right: float, top: float,
...
+    """
+    Draw a rectangle by specifying left, right, bottom and top edges.
+
+    :param float left: The x coordinate of the left edge of the rectangle.
+    :param float right: The x coordinate of the right edge of the rectangle.
+    :param float bottom: The y coordinate of the rectangle bottom.
+    :param float top: The y coordinate of the top of the rectangle.
+    :param Color color: The color of the rectangle.
+    :param float border_width: The width of the border in pixels. Defaults to one.
+    :Raises AttributeError: Raised if left > right or top < bottom.
```
pushfoo: `s/Attribute/Value/`

pushfoo (changes requested):
I apologize for leaving a separate review, but I think the new functions should raise `ValueError` instead of `AttributeError` since there's no attribute access involved.

arcade/utils.py
pushfoo: See the new reviews, btw. This point will be even truer after the exception type is changed for the new functions.

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

arcade/draw_commands.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

arcade/utils.py
Ibrahim2750mi: Updated in https://github.com/pythonarcade/arcade/pull/1650/commits/0af3ba990f9eacd0a13e188aa300615db4226460

pushfoo (approved):
Thank you for making the requested changes! It looks good! :+1: 

Please mark the earlier comment threads as resolved & we should be ready to merge.

PR #1653, opened 2023-03-21
Unbreak broken tutorials and pytest for tutorials

Ibrahim2750mi:
Unbreak(if broken) the following tutorials:
- [x] card game
- [x] compute shader
- [x] crt filter
- [x] framebuffer
- [x] gpu_particle_burst
- [x] lights
- [x] pymunk platformer
- [x] raycasting
- [x] shader inputer
- [x] shader toy glow
- [x] shader toy particles
- [x] views

Skipped tutorials:
+ pyinstaller
+ compiling with nutika
+ platfrom tutorials (source code present in examples to be looked over by Clepto)


Ibrahim2750mi:
Currently blocked by #1656

pushfoo (approved):
Looks good. The use of `SpriteList.swap` is a nice touch. Good attention to detail. :+1:

doc/tutorials/card_game/drag_drop_cards_full.py
```diff
@@ -218,9 +216,9 @@ def pull_to_top(self, card):
         index = self.card_list.index(card)
         # Loop and pull all the other cards down towards the zero end
         for i in range(index, len(self.card_list) - 1):
-            self.card_list[i] = self.card_list[i + 1]
+            self.card_list.swap(i, i+1)
         # Put this card at the right-side/top/size of list
-        self.card_list[len(self.card_list) - 1] = card
+        # self.card_list[len(self.card_list) - 1] = card
```
pvcraven: This code doesn't seem to be functionally equivalent?

doc/tutorials/pymunk_platformer/pymunk_demo_platformer_09.py
```diff
@@ -107,9 +107,6 @@ def __init__(self):
         # Set the initial texture
         self.texture = self.idle_texture_pair[0]
 
-        # Hit box will be set based on the first image used.
-        self.hit_box = self.texture.hit_box_points
-
```
pvcraven, line 112: The hitbox stays stable because if we switch textures and the new hitbox suddenly appears "inside" a solid object the engine doesn't have a good way to resolve the collision. Should it go left? Right? It doesn't know the original direction and how it got there.

pvcraven (changes requested):

doc/tutorials/pymunk_platformer/pymunk_demo_platformer_09.py
Ibrahim2750mi: Setting the `hit_box` with `self.texture.hit_box_points` throws an error. Because the hit box expects a `RotatableHitBox` and also `Texture` doesn't has a method which returns `RotatableHitBox`.

doc/tutorials/card_game/drag_drop_cards_full.py
Ibrahim2750mi: It is equivalent run the 12th step, this file had some other logic bug iirc.

doc/tutorials/pymunk_platformer/pymunk_demo_platformer_09.py
Ibrahim2750mi: Wait I think `self.set_hit_box` should do it.

doc/tutorials/card_game/drag_drop_cards_full.py
pvcraven: Right, I take that back. I'd remove lines 220, 221 then. Maybe add a comment on how the other one will float to the top.

doc/tutorials/compute_shader/main.py
```diff
@@ -16,6 +17,8 @@
 GRAPH_HEIGHT = 120
 GRAPH_MARGIN = 5
 
+CURRENT_DIR = Path(__file__).parent.resolve()
+
```
pvcraven: How necessary is this? 

At one point we had something similar in examples, and I was happy to rip it out. 

From an example perspective, the more you can simplify the example, the better. You want people to concentrate on the concept you are presenting. A new programmer might be confused by the line above, and never get to the topic you are trying to present.

I'd prefer to leave this out if possible. If we have to have it in, it needs to be commented. Because line 20 is up top, and is totally not understandable to someone that hasn't seen it before.

doc/tutorials/crt_filter/bloom_filter_example.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, width, height, title):
 
         for y in range(10, self.height, 50):
             color = random.choice([RED, YELLOW, ORANGE, GREEN, BLUEBERRY, AMETHYST])
-            my_sprite = arcade.SpriteCircle(random.randrange(40), color)
+            my_sprite = arcade.SpriteCircle(random.randrange(1, 40), color)
```
pvcraven, line 30: Nice catch.

doc/tutorials/compute_shader/main.py
Ibrahim2750mi: The tests will not run if the full path is not used, thats why I added the use of `pathlib.Path`.

gran4: Add a comment that either says don't worry about it or gets the files location.

doc/tutorials/card_game/drag_drop_cards_full.py
pushfoo: This should be addressed in https://github.com/Ibrahim2750mi/arcade/pull/6

Ibrahim2750mi:
@pushfoo Thanks for the co-author.

doc/tutorials/compute_shader/main.py
einarf: Change the `cwd` before running the example in unit tests? That's the least intusive.

I would order solutions like this (best to worst)
* Change `cwd` in unit test before importing/running the module (No code changes needed)
* Use `pathlib` (The code change with the least amount of consequences)
* Change `cwd` in the example (This is not any less "magical" than the pathlib solution, but it can create other problems for the user)

A 4th solution was implementing a resouce handle `:this:` (or `:here:`) that would use introspection to figure out what module the user attepmted to load the resource from.

pvcraven (approved):

pvcraven:
Thanks!!!

PR #1608, opened 2023-03-04
Tutorial for Menu

Ibrahim2750mi:
Please Review
What to review:
- [x] Check any grammar or spell errors.
- [x] Check that each diff hunk in menu_0[0-9].py is mentioned in the respective step [0-9].
- [x] Check any blank lines at the end of the codeblocks
- [x] Missing punctuation marks and using 'I' instead of 'we'

~~Also currently waiting for @eruvanos to propose a fix for UIDropdown 🙏🏽 in `menu_05.py`.~~ ~~The code errors now when operation the UISlider, its because of a bug/wrong code style.~~

PR #1659, opened 2023-03-24
Clarify & speed up basic views tutorial

pushfoo:
* Reduce coin count from 50 to 25 (focus is on views, not the game)


Ibrahim2750mi (approved):
LGTM!

pushfoo:
@pvcraven for context, the direction I'm leaning towards for this would be to move this under a "Screens & Menus with Views" headings akin to the shaders index we currently have.

I may want to have #1608 expand on it, but I'm not sure yet. What I am sure of is that 50 coins takes far too long to play through for a tutorial which is about the screens, not the gameplay.

doc/tutorials/views/01_views.py
```diff
@@ -4,7 +4,7 @@
 # --- Constants ---
 SPRITE_SCALING_PLAYER = 0.5
 SPRITE_SCALING_COIN = .25
-COIN_COUNT = 50
+COIN_COUNT = 3
```
pvcraven: I kind of like keeping the 50 coins. Yes, this is quicker but 50 coins doesn't take long to gather. Otherwise you don't get much of a game at all. And if the three coins spawn on top of the player? You'll switch to end game before you know what is happening.

doc/tutorials/views/index.rst
```diff
@@ -2,7 +2,7 @@
 
 .. _view-tutorial:
 
-Using Views for Start/End Screens
+Basic Non-Game Screens with Views
```
pvcraven: I get the generic wording is more accurate, but in my experience teaching, students tended to ask for start/end screens most often. When a student wants a "game over" screen, I'm not sure she'd know to look up "Non-game screens with views." I prefer the original in this case as well.

pvcraven (changes requested):

doc/tutorials/views/index.rst
pushfoo: reverted :+1:

doc/tutorials/views/01_views.py
pushfoo: Would 25 be ok as a middle ground, or is the goal to get users to adjust the value themselves?

pushfoo:
@pvcraven any objections to replacing code like this:
```py
sprite.center_x = 10
sprite.center_y = 10
```
with more efficient code which doesn't update the SpriteList twice?
```py
sprite.position = 10, 10
```
 
 Also sorry for the close / re-open. I had a touchpad issue.

pushfoo:
te

pvcraven:
> @pvcraven any objections to replacing code like this:
> 
> ```python
> sprite.center_x = 10
> sprite.center_y = 10
> ```
> 
> with a more efficient code which doesn't update the SpriteList twice?
> 
> ```python
> sprite.position = 10, 10
> ```

No. Some examples that are super-basic I avoid that because students might not be introduced to tuples yet. But by the time they use screens, shouldn't be an issue.

doc/tutorials/views/01_views.py
pvcraven: 25 doesn't bother me.

pvcraven (approved):

PR #1663, opened 2023-03-27
Made video players reusable by the user

Ibrahim2750mi:
Relevant discussion: https://discord.com/channels/458662222697070613/1089998750392537209/1090000815021887690

### DIFF
--- a/arcade/experimental/shadertoy_video_cv2.py
+++ b/arcade/experimental/shadertoy_video_cv2.py
@@ -41,10 +41,7 @@ class ShadertoyVideo(arcade.View):
                 }
             """,
         )
-        # Used this because it will throw SIGSEGV when passed a Path like object, which is not very descriptive.
-        if not issubclass(type(path), str):
-            raise TypeError(f"The path is required to be a str object and not a {type(path)} object")
-        self.video = cv2.VideoCapture(path)
+        self.video = cv2.VideoCapture(str(arcade.resources.resolve_resource_path(path)))
         width, height = (
             int(self.video.get(cv2.CAP_PROP_FRAME_WIDTH)),
             int(self.video.get(cv2.CAP_PROP_FRAME_HEIGHT)),
--- a/arcade/experimental/shadertoy_video_cv2.py
+++ b/arcade/experimental/shadertoy_video_cv2.py
@@ -14,10 +14,14 @@ SCREEN_HEIGHT = 300
 SCREEN_TITLE = "ShaderToy Video"
 
 
-class ShadertoyVideo(arcade.Window):
+class ShadertoyVideo(arcade.View):
+    """
+    Can be used to add effects like rain to the background of the game.
+    Make sure to inherit this view and call super for `__init__`, `on_draw`, `on_update` and `on_resize`.
+    """
 
-    def __init__(self, width, height, title):
-        super().__init__(width, height, title, resizable=True)
+    def __init__(self, path: str):
+        super().__init__()
         self.shadertoy = Shadertoy(
             self.get_framebuffer_size(),
             """
@@ -37,8 +41,10 @@ class ShadertoyVideo(arcade.Window):
                 }
             """,
         )
-        # INSERT YOUR OWN VIDEO HERE
-        self.video = cv2.VideoCapture("C:/Users/efors/Desktop/BigBuckBunny.mp4")
+        # Used this because it will throw SIGSEGV when passed a Path like object, which is not very descriptive.
+        if not issubclass(type(path), str):
+            raise TypeError(f"The path is required to be a str object and not a {type(path)} object")
+        self.video = cv2.VideoCapture(path)
         width, height = (
             int(self.video.get(cv2.CAP_PROP_FRAME_WIDTH)),
             int(self.video.get(cv2.CAP_PROP_FRAME_HEIGHT)),
@@ -67,8 +73,3 @@ class ShadertoyVideo(arcade.Window):
         frame = cv2.flip(frame, 0)
         if exists:
             self.video_texture.write(frame)
-
-
-if __name__ == "__main__":
-    ShadertoyVideo(SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_TITLE)
-    arcade.run()
--- a/arcade/experimental/shadertoy_video_cv2.py
+++ b/arcade/experimental/shadertoy_video_cv2.py
@@ -23,7 +23,7 @@ class ShadertoyVideo(arcade.View):
     def __init__(self, path: str):
         super().__init__()
         self.shadertoy = Shadertoy(
-            self.get_framebuffer_size(),
+            self.window.get_framebuffer_size(),
             """
                 void mainImage( out vec4 fragColor, in vec2 fragCoord )
                 {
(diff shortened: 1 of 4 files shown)


arcade/experimental/video_player.py
```diff
@@ -9,14 +9,20 @@
...
-        super().__init__(800, 600, "Video Player", resizable=True)
+    def __init__(self, path: str) -> None:
+        super().__init__()
 
         self.player = pyglet.media.Player()
-        # self.player.queue(pyglet.media.load("C:/Users/efors/Desktop/file_example_MP4_480_1_5MG.mp4"))
-        self.player.queue(pyglet.media.load("C:/Users/efors/Desktop/BigBuckBunny.mp4"))
+        # Used this because it will throw SIGSEGV when passed a Path like object, which is not very descriptive.
+        if not issubclass(type(path), str):
+            raise TypeError(f"The path is required to be a str object and not a {type(path)} object")
```
einarf: Just resolve the resource usng `arcade.resources.resolve_resource_path` and allow str and Path.

arcade/experimental/video_cv2.py
```diff
@@ -47,8 +51,10 @@ def __init__(self):
         # Configure videoFrame sampler to read from texture channel 0
         self.program["videoFrame"] = 0
 
-        # Open the video (can also read from webcam)
-        self.video = cv2.VideoCapture("C:/Users/efors/Desktop/BigBuckBunny.mp4")
+        # Used this because it will throw SIGSEGV when passed a Path like object, which is not very descriptive.
+        if not issubclass(type(path), str):
+            raise TypeError(f"The path is required to be a str object and not a {type(path)} object")
```
einarf: Just resolve the resource usng arcade.resources.resolve_resource_path and allow str and Path.

Ibrahim2750mi: It shouldn't return a Path object, `arcade.resources.resolve_resource_path` returns a Path, although we can do str() conversion for the return object

arcade/experimental/video_cv2.py
```diff
@@ -14,12 +14,16 @@
 import cv2  # type: ignore
 
 
-class CV2Player(arcade.Window):
-
-    def __init__(self):
-        super().__init__(800, 600, "OpenCV Video Player", resizable=True)
+class CV2Player(arcade.View):
```
einarf: In a perfect world we should separate the player from the view. You don't have to do that, but it will surely make things more reusable.

arcade/experimental/video_player.py
```diff
@@ -9,14 +9,20 @@
 import arcade
 
 
-class VideoPlayer(arcade.Window):
```
einarf, line 12: It would be a good idea to make a separate videoplayer class here that is used by the view itself?

Ibrahim2750mi: Yes on it.

arcade/experimental/video_cv2.py
```diff
@@ -44,40 +52,53 @@ def __init__(self):
...
+
+    @property
+    def width(self):
+        """Video width."""
+        return int(self.video.get(cv2.CAP_PROP_FRAME_WIDTH))
+
+    @property
+    def height(self):
+        """Video height."""
+        return int(self.video.get(cv2.CAP_PROP_FRAME_HEIGHT))
```
einarf: Should use the width and height already queried in the initializer.

arcade/experimental/video_cv2.py
```diff
@@ -44,40 +52,53 @@ def __init__(self):
...
 
-        # Open the video (can also read from webcam)
-        self.video = cv2.VideoCapture("C:/Users/efors/Desktop/BigBuckBunny.mp4")
+        self.video = cv2.VideoCapture(str(arcade.resources.resolve_resource_path(path)))
+
         # Query video size
         width, height = (
             int(self.video.get(cv2.CAP_PROP_FRAME_WIDTH)),
             int(self.video.get(cv2.CAP_PROP_FRAME_HEIGHT)),
         )
```
einarf: `self._width, self._height = ` and return these in the properties.

arcade/experimental/video_cv2.py
```diff
@@ -9,17 +9,25 @@
...
+    """
+    Primitive video player for arcade with cv2.
+    Can be used to add effects like rain in the background.
+
+    :param path: Path of the video that is to be played.
+    """
 
-    def __init__(self):
-        super().__init__(800, 600, "OpenCV Video Player", resizable=True)
+    def __init__(self, path: Union[str, Path], ctx: arcade.ArcadeContext):
```
einarf: We can just use `get_window().ctx` to store the context internally?

arcade/experimental/video_cv2.py
```diff
@@ -9,17 +9,25 @@
...
 import arcade
 from arcade.gl.geometry import quad_2d_fs
 import cv2  # type: ignore
 
 
-class CV2Player(arcade.Window):
+class VideoPlayerCV2:
+    """
+    Primitive video player for arcade with cv2.
+    Can be used to add effects like rain in the background.
```
einarf: Maybe this example is a bit too specific? I don't think we need mention example usage in the player.

arcade/experimental/video_cv2.py
```diff
@@ -90,4 +111,30 @@ def on_update(self, delta_time: float):
                 self.texture.write(frame)
 
 
-CV2Player().run()
+class CV2PlayerView(arcade.View):
+    """
+    Can be used to add effects like rain to the background of the game.
+    Make sure to inherit this view and call super for `__init__`, `on_draw` and `on_update`.
+    """
```
einarf: I think this docstring is unnecessary. It's really jist meant as an example.

arcade/experimental/video_player.py
```diff
@@ -51,5 +60,20 @@ def get_video_size(self):
...
+
+    def __init__(self, path) -> None:
+        super().__init__()
+
+        self.video_player = VideoPlayer(path)
+
+    def on_draw(self):
+        self.clear()
+
+        self.video_player.draw(self.window.ctx, self.window.width, self.window.height)
```
einarf: Maybe store the context in the video player instead. Make `width`, `height` and `size` property for both player types and support a `size: Tuple[int, int]` parameter for both draw methods? It can be optional using the view size if not supplied. Later we can add support for size in the cv2 video shader. No stress about that now.

arcade/experimental/video_cv2.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/video_cv2.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/video_cv2.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/video_cv2.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/video_cv2.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/video_player.py
Ibrahim2750mi: Updated https://github.com/pythonarcade/arcade/pull/1663/commits/a9f5133090d122f77571a90f14558a427a1d8d5a

arcade/experimental/shadertoy_video_cv2.py
```diff
@@ -37,18 +41,17 @@ def __init__(self, width, height, title):
...
             int(self.video.get(cv2.CAP_PROP_FRAME_HEIGHT)),
         )
-        self.video_texture = self.ctx.texture((width, height), components=3)
-        self.video_texture.wrap_x = self.ctx.CLAMP_TO_EDGE
-        self.video_texture.wrap_y = self.ctx.CLAMP_TO_EDGE
+        self.video_texture = self.window.ctx.texture((width, height), components=3)
+        self.video_texture.wrap_x = self.window.ctx.CLAMP_TO_EDGE
+        self.video_texture.wrap_y = self.window.ctx.CLAMP_TO_EDGE
         self.video_texture.swizzle = "BGR1"
         self.shadertoy.channel_0 = self.video_texture
```
einarf, line 53: Could we just use the CV2 video player here?

einarf:
It's not perfect, but this is a good start for now.

Issue #1183, opened 2022-04-10
PymunkPhysicsEngine creating invisible collision boxes with some maps


einarf:
During the pyweek game jam the PymunkPhysicsEngine was creating invisible collision areas for the tmx map in this project : https://github.com/aphkyle/xn--m28h

Had to swap to PhysicsEngineSimple. It might be something with the map or we have a bug somewhere. I remember there was a bug with collision in the tilmap not long ago, so I suspect the bug was compensated for in the pymunk version, but broke again when it was fixed in the map loading?

gran4:
@einarf I think you can close this.

PR #1666, opened 2023-04-01
Fixes collisions(#1183).

gran4:
Before, if min_a, min_b, max_a, max_b weren't changed by the code, it would break(make odd collisions) as https://github.com/pythonarcade/arcade/issues/1183 notes. Now, it always creates the collisions


pvcraven:
Thank you!

einarf:
Do we have some sample data of some hit box points that triggered this issue? It would be useful adding this to unit tests so we can cover future implementations including the rust version.

Issue #1651, opened 2023-03-20
Scene improvements


einarf:
The current `sprite_lists` and `name_mapping` members are "public". Some users are modifying these directly causing issues that are hard to debug and understand. We should make them private and add additional tools if needed.

https://github.com/pythonarcade/arcade/blob/98d4caca12768d712cdcf7e04799f6e57ebb9d47/arcade/scene.py#L30-L32

Things like `__len__` and `__delitem__` are candidates here.

On a few occations users have been deleting spritelists from `name_mapping` thinking that this will remove the spritelist. For more dynamic worlds this means you end up with hundres of spritelists in `spritelists` and performance tanks.

einarf:
Another source of bugs is that `add_sprite_list` doesn't yell if you are adding a spritelist with an existing nane. It will overwrite the existing dict entry and append the spritelist to `spritelists`. This means games can easily lose track of spritelists and the game will grind to a halt over time.

We should also cover all these cases in unit tests.

gran4:
You can't make things in python private. I think the best way to go about this is to give warnings and comments in the code. I am also trying to getting to detect when it is happening.

einarf:
> You can't make things in python private. I think the best way to go about this is to give warnings and comments in the code. I am also trying to getting to detect when it is happening.

I know. I guess non-public is a better word. Just starting the name with an underscore. Most programmers know this means you shouldn't really access it or that there might be some bad consequences for doing so. Trying to avoid this happening is usually not worth it. (PEP 8)

We do this everywhere in the arcade codebase.

gran4:
@einarf I think you can close this.

pushfoo:
> Things like `__len__` and `__delitem__` are candidates here.

I'd like to take another look at this before closing.

Cleptomania:
I think adding `__len__` and `__delitem__` would be good. They should be fairly easy to add so no reason not to.

tiffanyxiao:
@Cleptomania are you all at PyCon? Would love to try tackling this issue if someone hasn't already claimed it. I tried finding the Arcade group in 250D but was unsuccessful.

Cleptomania:
@tiffanyxiao Just missed us! We went out to lunch but will be back this afternoon. I’ll also be around for Tuesday and Wednesday

pushfoo:
> Would love to try tackling this issue if someone hasn't already claimed it.

Your help would be appreciated!

gran4:
[...]

PR #1676, opened 2023-04-02
Scene improvements

gran4:
improved scene, made it 2 variables non-public(issue #1651)
https://github.com/pythonarcade/arcade/issues/1651


Cleptomania:
Looks good to me

PR #1679, opened 2023-04-03
Broke up create_text_sprite into 2 functions [WIP][HELP]

### LINKED ISSUE #1410: Break up arcade.create_text_sprite into several useful functions

In 2.6.16, there was a function called arcade.create_text_image. In 2.7.1.dev3, it is gone. Instead, there is arcade.create_text_sprite.
This function first creates arcade.Text, then renders that into an arcade.Texture, then puts that into arcade.Sprite and returns it. It seems like it would be possible to have arcade.create_text_image put back, but using the new rending method instead. Also, there could be an arcade.create_text_texture. All that is needed is to break out parts of arcade.create_text_sprite into functions which return texture and/or image.

### What would it help with?
I want to be able to create instances of arcade.Texture which contain text, which I can use in my game. To do this now, I am replacing calls to arcade.create_text_image with calls to arcade.create_text_sprite and then take .texture on it, which means Sprite is needlessly instantiated.

gran4:
Fixed https://github.com/pythonarcade/arcade/issues/1410

Help me fix multiline issue(everything but multiline works).

### DIFF
--- a/HI.py
+++ b/HI.py
@@ -1,0 +1,84 @@
+import arcade, pyglet
+
+def _attempt_font_name_resolution(font_name):
+    """
+    Attempt to resolve a tuple of font names.
+
+    Preserves the original logic of this section, even though it
+    doesn't seem to make sense entirely. Comments are an attempt
+    to make sense of the original code.
+
+    If it can't resolve a definite path, it will return the original
+    argument for pyglet to attempt to resolve. This is consistent with
+    the original behavior of this code before it was encapsulated.
+
+    :param Union[str, Tuple[str, ...]] font_name:
+    :return: Either a resolved path or the original tuple
+    """
+    if font_name:
+
+        # ensure
+        if isinstance(font_name, str):
+            font_list: Tuple[str, ...] = (font_name,)
+        elif isinstance(font_name, tuple):
+            font_list = font_name
+        else:
+            raise TypeError("font_name parameter must be a string, or a tuple of strings that specify a font name.")
+
+        for font in font_list:
+            try:
+                path = resolve(font)
+                # print(f"Font path: {path=}")
+
+                # found a font successfully!
+                return path.name
+
+            except FileNotFoundError:
+                pass
+
+    # failed to find it ourselves, hope pyglet can make sense of it
+    return font_name
+
+
+class UsageAttempt(arcade.Window):
+
+    def __init__(self, width: int = 320, height: int = 240):
+        super().__init__(width=width, height=height)
+
+        self.sprites = arcade.SpriteList()
+        text_sprite = arcade.create_text_sprite(
+            "First line\nsecond line",
+            multiline=True,
+            width=200,
+        )
+        text_sprite.position = self.width // 2, self.height // 2
+        self.sprites.append(text_sprite)
+
+        self._label = pyglet.text.Label(
+            text="First line\nsecond line\nTHIRD NIGER",
+            x = 200,
+            y = 200,
+            font_name="Arial",
+            font_size=12,
+            anchor_x="left",
+            width=100,
+            align="baseline",
+            bold=True,
+            italic=True,
+            multiline=True,
+        )
+
+    def on_draw(self):
+        self.clear()
+        self.sprites.draw()
+
+        window = arcade.get_window()
+        with window.ctx.pyglet_rendering():
+            self._label.draw()
+
+
+
+
+if __name__ == "__main__":
+    w = UsageAttempt()
+    arcade.run()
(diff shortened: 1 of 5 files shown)


einarf:
There was some challenges related to implementing these changes I can't remember at the moment. Will have to look closer at this one.

Also: Seems your unit tests are failing. Did you run them locally before pushing?

gran4:
I had to add create_text_texture to __all__.

gran4:
I think you guys can review. @einarf

arcade/text.py
```diff
@@ -580,17 +579,13 @@ def create_text_sprite(
...
-    Creates a sprite containing text based off of :py:class:`~arcade.Text`.
-
-    Internally this creates a Text object and an empty texture. It then uses either the
-    provided texture atlas, or gets the default one, and draws the Text object into the
-    texture atlas.
+    Creates a texture containing text based off of :py:class:`~pyglet.text.Label`.
 
-    It then creates a sprite referencing the newly created texture, and positions it
-    accordingly, and that is final result that is returned from the function.
+    Internally this creates a pyglet.text.Label object and an empty texture. It then uses either the
```
eschan145: Use this instead:
```
:py:class:`~pyglet.text.label`
```

This will yield RST formatted text. Also, try to wrap this line to eighty characters to make it more human-readable.

gran4:
@einarf I think you can merge this. Is there any problem with it?

einarf:
I will look at it when back from hospital unless someone else beats me to it. It might linger here for a while,

einarf:
I'm wondering if we should try delay this feature after 3.0.0 when other related systems are in better shape. When I tested this last time there was some text getting cutt off as well in certain enviroments. Either afer 3.0.0 or closer to release.

gran4:
@einarf It doesn't really deal with any of the lower level stuff. All it is suppose to do is allow you to get the texture in certain use cases like issue #1410. If the text was cut off, it is either a problem with the existing code, or something I changed here. I do not think the other systems are relevant in this case.

pushfoo (changes requested):
If I'm not missing something, it looks like multiline doesn't work. Either something is wrong, or it's a limitation the doc needs to be clear about.

image
```python
import arcade


class UsageAttempt(arcade.Window):

    def __init__(self, width: int = 320, height: int = 240):
        super().__init__(width=width, height=height)

        self.sprites = arcade.SpriteList()
        text_sprite = arcade.create_text_sprite(
            "First line\nsecond line",
            multiline=True,
            width=200,
        )
        text_sprite.position = self.width // 2, self.height // 2
        self.sprites.append(text_sprite)

    def on_draw(self):
        self.clear()
        self.sprites.draw()


if __name__ == "__main__":
    w = UsageAttempt()
    arcade.run()
```

gran4:
> If I'm not missing something, it looks like multiline doesn't work. Either something is wrong, or it's a limitation the doc needs to be clear about.

Working to fix it. Don't review yet. I don't know what the problem is.

bunny-therapist:
To me, the `fbo.clear((0, 0, 0, 255))` is causing the text to have a black box behind it. Should that really be there?

pushfoo:
> To me, the `fbo.clear((0, 0, 0, 255))` is causing the text to have a black box behind it. Should that really be there?

Please try swapping the 255 for a 0 if you have time. I need to look at some other PRs right now, and your help would be appreciated.

If replacing it works, the following steps would probably work improve the clarity of the code in addition to fixing the issue:
1. add `from arcade.color import TRANSPARENT_BLACK` to the imports
2. replace the tuple literal with `TRANSPARENT_BLACK` for clarity

bunny-therapist:
`fbo.clear((0,0,0,0))` works. I mean, it does not seem (to me) to make a difference adding it since the texture is already empty, so clearing it makes no difference, but it does not add a black box around the text.

HI.py
```diff
@@ -0,0 +1,84 @@
...
+        window = arcade.get_window()
+        with window.ctx.pyglet_rendering():
+            self._label.draw()
+
+
+
+
+if __name__ == "__main__":
+    w = UsageAttempt()
+    arcade.run()
```
einarf: This file should definitely not be in the PR.

einarf (changes requested):

HI.py
einarf: We only really allow test scripts in places like `arcade/experimental` and that's mainly things not in the root.

gran4: I already know. It is a test file to test the dev version. I will remove the file right before it is merged. Before that happens, the multiline must be fixed. I do not know the problem still.

einarf:
Thanks again for the PR. Since it contains `Hi.py` and liely need some more tweaks I will close this for now and see what I can salvage from it.

Issue #1685, opened 2023-04-05
Add support for bitmapped fonts


gran4:
### What should be added/changed?
Add something for custom text.

### What would it help with?
It would make it easier to put custom fonts(not actually fonts, but pictures) into a game.

pushfoo:
> custom fonts(not actually fonts, but pictures)

Do you mean bitmap fonts from sprite sheets like this one?
![image](https://user-images.githubusercontent.com/36696816/230213378-8e577891-f032-40b4-825c-e1de7ae3394c.png)

I've brought it up in the past in #842, but closed it since [full bitmap support should probably be added to pyglet first](https://github.com/pythonarcade/arcade/issues/842#issuecomment-1454367176). That may be somewhat difficult and will require work on the arcade side as well. If you need something sooner, you may be able to put something together yourself with the UI classes.

gran4:
That's what I mean, just didn't know the name. The thing I created doesn't split the bit map for you. I can add that in a separate function

pushfoo:
As I outlined in [this PR discussion comment](https://github.com/pythonarcade/arcade/pull/1686#discussion_r1160374063):

1. General font rendering is very hard to solve
2. The reliable but difficult way to handle this is probably adding bitmap font support to pyglet, then arcade
3. This issue's intended use cases are unclear

For anyone looking for a fast development solution, here's a table of goals and ways to achieve them:

| Goal                                                     | Alternative to implementing custom text renderer                    |
|----------------------------------------------------------|--------------------------------------------------------------|
| Use a pixel font | Use [the built-in kenney fonts](https://api.arcade.academy/en/latest/examples/drawing_text_objects.html) or make a custom TTF font with a tool like [BitFontMaker2](https://www.pentacom.jp/pentacom/bitfontmaker2/) & load it                          |
| Embed custom images inline with text                     | Use pyglet's HTML labels & [their support for custom inline elements](https://pyglet.readthedocs.io/en/latest/programming_guide/text.html#custom-elements) |
| Arrange GUI icons in a line                                  | [Use UITextureButton with layouts](https://api.arcade.academy/en/latest/examples/gui_widgets.html#gui-widgets)                                        |

DigiDuncan:
This has come up so many times, that I really think we either need to
a) straight up add rudimentary bitmap text support to Arcade (maybe restrict it to monospace?)
b) get Pyglet to do it
c) have a page detailing why were *aren't* doing this, and maybe even "how to implement it yourself."

[...]

Issue #842, opened 2021-04-13
Add a DigitDisplay UI element


pushfoo:
### What should be added/changed?
A `DigitDisplay` UI class that allows the score to be set without using a text label. It would support custom bitmap fonts to display scores or time remaining fancily.

### What would it help with?
![image](https://user-images.githubusercontent.com/36696816/114509549-a3b03080-9c03-11eb-9aab-01de439569d3.png)

**tl;dr:** Things like the score, coin counter, and time remaining display in Super Mario Bros. It would be much faster than the current text labels.

Right now, text UI elements [allocate three new textures each time the text is changed](https://github.com/pythonarcade/arcade/blob/f493981dbe36f22c008c8c26e67937c18f0cd001/arcade/gui/elements/label.py#L54). This means that using a text label to display a score can thrash texture allocation when it is updated rapidly. At least two people (including myself) have been frustrated by this unexpected behavior from text labels when attempting to add time remaining or score displays to games.

Possible use cases include:
  - overriding a conversion method to convert raw seconds to a timer displaying `f"{new_value // 60}:{new_value % 60}"` in bitmap sprite form
  - displays in puzzle exploration games like [Myst](https://en.wikipedia.org/wiki/Myst) using unconventional digit systems (base 7, [cistercian numerals](https://www.zmescience.com/science/cirstercian-numbers-90432432/), etc)
   -  an emulator showing integers as hex or octal

### Why do this now instead of fixing UITextLabel?
My understanding of the discord discussion around text labels is that the text rendering update has been in progress for a while and has unresolved prerequisites. A small, usable working digit display class that doesn't confuse new users with performance issues can be implementable faster than the text rendering update gets merged.  Once we settle on an API for a digit display, the implementation can be replaced once the text rendering refresh is merged.

einarf:
Can't this be easily extended to a bitmapped font thing supporting latin1? Then it will probably have a longer life span.

If we end up making a custom renderer for UI this will be depreacted pretty fast. Something independent of the UI would be much easier.

pushfoo:
> Can't this be easily extended to a bitmapped font thing supporting latin1? Then it will probably have a longer life span.

[...]

PR #1686, opened 2023-04-06
Created a CustomTextSprite, Addresses https://github.com/pythonarcade/arcade/issues/1685

gran4:
Created a CustomTextSprite, Addresses https://github.com/pythonarcade/arcade/issues/1685

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -632,14 +632,15 @@ class CustomTextSprite(object):
         if not text:
             return
         words: list[str] = text.split(' ')
-        x: float = 0
-        y: float = 0
+        x: float = center_y
+        y: float = center_x
         for word in words:
-            if not word:
-                continue
             if x > width/2+center_x or word == "\n":
                 y -= text_margin
-                x = -width/2
+                x = -width/2+center_x
+            if not word:
+                x += text_margin*scale
+                continue
 
 
             for char in word:
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -569,6 +569,50 @@ class Text:
             self._label.position = *point, self._label.z
 
 
+class CustomTextSprite(object):
+    def __init__(self, string, Alphabet_Textures, scale=1, 
+                 center_x=0, center_y = 0, 
+                 text_scale=1, text_margin=16, width=100, height = 40,  Background_offset_x=0, Background_offset_y=0, Background_scale=1, Background_Texture=None) -> None:
+        super().__init__()
+        self.Sprite_List = arcade.SpriteList()
+        self.Background_Sprite = arcade.Sprite(Background_Texture, center_x=center_x+width/2+Background_offset_x, center_y=center_y-height*2+Background_offset_y, scale=Background_scale)
+        
+        self.text_scale = text_scale
+        self.width = width
+        self.height = height
+        self.update_text(string, Alphabet_Textures, scale=scale, text_scale=text_scale,
+                 center_x=center_x, center_y=center_y, 
+                 text_margin=text_margin, width=width, height = height)
+        
+    def update_text(self, text, Alphabet_Textures, scale=1, text_scale=1,
+                 center_x=0, center_y = 0, 
+                 text_margin=16, width=100, height = 40):
+        self.text = text
+        self.Sprite_List.clear()
+        if not text:
+            return
+        words = text.split(' ')
+        x = 0
+        y = 0
+
+        for word in words:
+            if x > width/2+center_x:
+                y -= text_margin
+                x = -width/2
+
+
+            for string in word:
+                sprite = arcade.Sprite(center_x=center_x+x, center_y=center_y+y, scale=scale*text_scale)
+                sprite.texture = Alphabet_Textures[string]
+                self.Sprite_List.append(sprite)
+                x += text_margin*scale
+            x += text_margin*scale
+    def draw(self):
+        self.Background_Sprite.draw()
+        self.Sprite_List.draw()
+
+
+
 def create_text_sprite(
     text: str,
     color: RGBA255 = arcade.color.WHITE,
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -9,7 +9,7 @@ import pyglet
 import arcade
 from arcade.types import Color, Point, RGBA255
 from arcade.resources import resolve
-from arcade.utils import PerformanceWarning, warning
+from arcade.utils import PerformanceWarning, TryNotToUseWarning, warning
 
 
 def load_font(path: Union[str, Path]) -> None:
@@ -581,7 +581,10 @@ def create_bitmap_font_from_spritesheet(spritesheet_file: str = "", order:str = 
     for i in range(len(order)):
         Alphabet_Textures[order[i]] = textures[i]
 
-
+@warning(
+    message="CustomTextSprite is case sensitive(Ex. It doesn't support emojis). Consider using making a custom text sprite instead.",
+    warning_type=TryNotToUseWarning,
+    )
 class CustomTextSprite(object):
     def __init__(
         self, 
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -638,7 +638,6 @@ class CustomTextSprite(object):
         self.Sprite_List.draw()
 
 
-
 def create_text_sprite(
     text: str,
     color: RGBA255 = arcade.color.WHITE,
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -638,11 +638,14 @@ class CustomTextSprite(object):
             if x > width/2+center_x or word == "\n":
                 y -= text_margin
                 x = -width/2
+            if not word: 
+                x += text_margin*scale
+                continue
 
 
             for string in word:
-                sprite = arcade.Sprite(center_x=center_x+x, center_y=center_y+y, scale=scale*text_scale)
-                sprite.texture = chracter_textures[string]
+                sprite: arcade.BasicSprite = arcade.BasicSprite(center_x=center_x+x, center_y=center_y+y, scale=scale*text_scale)
+                sprite.texture: arcade.Texture = chracter_textures[string]
                 self.Sprite_List.append(sprite)
                 x += text_margin*scale
             x += text_margin*scale
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -569,6 +569,19 @@ class Text:
             self._label.position = *point, self._label.z
 
 
+def create_bitmap_font_from_spritesheet(spritesheet_file: str = "", order:str = ""):
+    """
+         create bitmap font from spritesheet
+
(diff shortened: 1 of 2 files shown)


arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
             self._label.position = *point, self._label.z
 
 
+class CustomTextSprite(object):
+    def __init__(self, string, Alphabet_Textures, scale=1, 
+                 center_x=0, center_y = 0, 
+                 text_scale=1, text_margin=16, width=100, height = 40,  Background_offset_x=0, Background_offset_y=0, Background_scale=1, Background_Texture=None) -> None:
```
eschan145: Pythonic parameter/variable names never start with a capital letter. **Only constants are in full capitals, and only classes are in `CamelCase`**. Python mainly uses `snake_case`.

This line is also 172 characters, and that is too long. Try to keep this at eighty characters for readability.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+        super().__init__()
+        self.Sprite_List = arcade.SpriteList()
```
eschan145: You need to add a type hint to this, as the doc build marks this wrong. Type hints help users know what type an object is.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+        
+        self.text_scale = text_scale
+        self.width = width
+        self.height = height
+        self.update_text(string, Alphabet_Textures, scale=scale, text_scale=text_scale,
+                 center_x=center_x, center_y=center_y, 
+                 text_margin=text_margin, width=width, height = height)
+        
+    def update_text(self, text, Alphabet_Textures, scale=1, text_scale=1,
+                 center_x=0, center_y = 0,
```
eschan145: This line has inconsistent indent with the function definition line.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+        self.text = text
+        self.Sprite_List.clear()
+        if not text:
+            return
+        words = text.split(' ')
+        x = 0
+        y = 0
+
+        for word in words:
+            if x > width/2+center_x:
```
eschan145: Adding spaces in necessary areas adheres to the PIP guidelines and neatens up code.

eschan145 (changes requested):
Make sure to make parts of code consistent and lowercase, as well as removing any unneeded newlines and adding docs. 😉

arcade/text.py
pushfoo: Do you mean

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+                y -= text_margin
+                x = -width/2
+
+
```
pushfoo, line 648:
```suggestion

```
Double-newlines are reserved for use at the module level, not within classes or within class definitions.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+                sprite.texture = Alphabet_Textures[string]
+                self.Sprite_List.append(sprite)
+                x += text_margin*scale
+            x += text_margin*scale
+    def draw(self):
+        self.Background_Sprite.draw()
+        self.Sprite_List.draw()
+
+
+
```
pushfoo:
```suggestion


```
This should be a double newline, not a triple.

arcade/text.py
pushfoo:
```suggestion
        self.sprite_list = arcade.SpriteList()
```
This should also follow the member variable naming rules eschan145 describer earlier.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+        x = 0
+        y = 0
+
+        for word in words:
+            if x > width/2+center_x:
+                y -= text_margin
+                x = -width/2
+
+
+            for string in word:
```
pushfoo, line 649: `string` isn't clear here. `char` might be clearer, or `character`. 

There's also another hard problem left unaddressed here, and I don't blame you for missing it since text rendering is surprisingly complicated: your current code doesn't accurately cover multi-character [symbols such as emoji](https://unicode.org/emoji/charts/emoji-zwj-sequences.html). Pyglet already implements some support for detecting these if I remember correctly.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
             self._label.position = *point, self._label.z
 
 
+class CustomTextSprite(object):
+    def __init__(self, string, Alphabet_Textures, scale=1,
```
pushfoo: these need type annotations.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+
+        for word in words:
+            if x > width/2+center_x:
+                y -= text_margin
+                x = -width/2
+
+
+            for string in word:
+                sprite = arcade.Sprite(center_x=center_x+x, center_y=center_y+y, scale=scale*text_scale)
+                sprite.texture = Alphabet_Textures[string]
```
pushfoo: You could use a `BasicSprite` here instead and pass in the texture. It would have less overhead.

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+                 text_margin=text_margin, width=width, height = height)
+        
+    def update_text(self, text, Alphabet_Textures, scale=1, text_scale=1,
+                 center_x=0, center_y = 0, 
+                 text_margin=16, width=100, height = 40):
+        self.text = text
+        self.Sprite_List.clear()
+        if not text:
+            return
+        words = text.split(' ')
```
pushfoo: This seems to leave important cases uncovered:
1. What happens when the user intentionally includes multiple spaces?
2. What happens when there's a newline character?

arcade/text.py
```diff
@@ -569,6 +569,50 @@ def position(self, point: Point):
...
+        self.text_scale = text_scale
+        self.width = width
+        self.height = height
+        self.update_text(string, Alphabet_Textures, scale=scale, text_scale=text_scale,
+                 center_x=center_x, center_y=center_y, 
+                 text_margin=text_margin, width=width, height = height)
+        
+    def update_text(self, text, Alphabet_Textures, scale=1, text_scale=1,
+                 center_x=0, center_y = 0, 
+                 text_margin=16, width=100, height = 40):
```
pushfoo: 1. it's unclear what `text_margin` is
2. this seems like it should be a property which is also exposed via a private instance variable

arcade/text.py
pushfoo: What is the intent of this block? It seems like it will never be true unless the user specifies a `center_x` value negative enough to overcome `width`.

pushfoo (changes requested):
As I stated in my comments [on #1685](https://github.com/pythonarcade/arcade/issues/1685#issuecomment-1498184396) and [on #842](https://github.com/pythonarcade/arcade/issues/842#issuecomment-1454367176), this would be best if be implemented via adding full bitmap text support to pyglet rather than trying to build it in arcade.

I've left some comments to help you improve this for your own personal use, but text rendering is a surprisingly difficult problem. Although there are exceptions, it's often best to use mature text rendering libraries instead of trying to roll your own. Core arcade features are no exception, but I'll let others chime in on this as well.

arcade/text.py
gran4: It's not private, it is the space between the characters.

arcade/text.py
gran4: What should I do? Should I make a separate dictionary for emojis and have a thing for starting them? Like emoji_Face or @@@Face or something?

einarf:
There is a good reason this issue is left untuched. It's actually not a trivial problem to solve.

Like I mentioned a few days ago: It's always good do first discuss how an issue should be solved before you start on it.

gran4:
This class was good enough for my game. It can be useful in many cases. Maybe I can add a warning? I still think it is worth adding.

arcade/text.py
pushfoo: #### tl;dr Use this class in your projects as-is, but don't try to implement general text rendering

Why?
1. [Text rendering is a very hard problem](https://faultlore.com/blah/text-hates-you/) 
2. It's easier to use a TTF font maker like [BitFontMaker2](https://www.pentacom.jp/pentacom/bitfontmaker2/) as a stop-gap until pyglet's bitmap font support improves. 


#### I'm still not convinced!

If you're absolutely sure you want to use bitmap fonts straight from an image file, I'd use this class in your own projects or polish it into an add-on module outside of arcade. I'd be glad to help you with covering more edge cases via Discord.

The article I linked in the tl;dr doesn't even mention the difficulties you can have with other libraries. I paused [my own attempt at general text loading & rendering](https://github.com/pushfoo/fontknife) because of reliability & distribution problems with dependencies. Based on that experience, my advice on implementing your own general text rendering is the same as einarf's: don't try to unless you absolutely have to.

If you want to expand your class to cover as much of standard glyph rendering as possible for use in your own projects or as a hobby project, I'd start with the same route I took:

1. Add [an expanded regex module](https://pypi.org/project/regex/) to your project to allow chunking strings with multi-character glyph support
3. Use the library to map emoji annotations to image data from sprite sheet tiles
4. Use a recognizable filler glyph (often called the 'tofu') to represent missing glyphs in a font atlas

This will get you far enough to duplicate a large portion of pyglet's font loading plus some extra, but without its layout features. Variable width fonts will take extra work, and I didn't bother implementing those in my project yet.

Adding support for bitmap fonts and emoji to arcade is more complicated than just loading the fonts since we have to do it in pyglet. That project has a hard rule of no external python package dependencies, which means the following have to be written:

1. Well-polished bitmap font loading support that returns font atlas objects
2. New string parsing with full support for multi-character glyphs

Both are a lot harder than they sound, and this comment doesn't even cover issues like support for color emoji fonts. The last one is the show-stopping library issue blocking fontknife development.

eschan145:
@gran4 if you are just using this for _your_ game, maybe you should put it just in the code of your game instead of in arcade's repository. As @pushfoo stated, this is not something easy to do and it doesn't have much of a use case. And @einarf stated that we should discuss more about this before you immediately open a PR. BTW all of the checks have failed; you have to go in there and see what's wrong. I also don't think you should make an entirely new `TryNotToUseWarning` class; this the first time it will be used in the history of arcade, and probably the last too. Just write, "Warning" normally.

einarf:
In open source we would call this PR an unsolicited favor. It could potentially be a very good idea, but it has not been discussed anywhere. It's basically a dice roll if you get it right.

Adding new stuff to a library has a much higher bar than something you add to your personal project. 

Let's start talking more in the discord dev channels and things will be better 👍(It's a learning process). Closing this for now since I don't see this one going anywhere.

PR #1687, opened 2023-04-06
Touch-up top-level docstrings for Sprite & BasicSprite

pushfoo:
tl;dr

1. Replace the hard URL link from `Sprite`'s top-level docstring with a `:ref:` to this doc build's version
2. Rephrase `Sprite`'s docstring to be more PEP-compliant
3. Add a top-level docstring for `BasicSprite`

Doc built & tested locally.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -17,10 +17,9 @@ class BasicSprite:
     """
     The absolute minimum needed for a sprite.
 
-    .. warning:: Beginners should see :py:class:`~arcade.Sprite` instead!
-
-                 This class is best used for performance optimization or
-                 internal arcade components.
+    It does not support features like rotation or changing the hitbox
+    after creation. For more built-in features, please see
+    :py:class:`~arcade.Sprite`.
 
     :param texture: The texture data to use for this sprite.
     :param scale: The scaling factor for drawing the texture.
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -16,6 +16,16 @@ SpriteType = TypeVar("SpriteType", bound="BasicSprite")
 class BasicSprite:
     """
     The absolute minimum needed for a sprite.
+
+    .. warning:: Beginners should see :py:class:`~arcade.Sprite` instead!
+
+                 This class is best used for performance optimization or
+                 internal arcade components.
+
+    :param texture: The texture data to use for this sprite.
+    :param scale: The scaling factor for drawing the texture.
+    :param center_x: Location of the sprite along the X axis in pixels.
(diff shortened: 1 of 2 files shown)


arcade/sprite/base.py
```diff
@@ -16,6 +16,16 @@
 class BasicSprite:
     """
     The absolute minimum needed for a sprite.
+
+    .. warning:: Beginners should see :py:class:`~arcade.Sprite` instead!
+
+                 This class is best used for performance optimization or
+                 internal arcade components.
```
einarf: I don't think this comment is necessary. We can cover sprites in the programming manual. `BasicSprite` doesn't really work that differently from the normal sprite.

Maybe instead we should mention the limitations. BasicSprite don't support custom hit box and doesn't have rotation. That's basically the difference I think.

pvcraven: Maybe take out the warning. Note the major differences. I think BasicSprite for display-only, while Sprite includes collisions? (Not sure that's true, but I think something helping the choice of one vs. the other.)

pvcraven (changes requested):
Agree with  einarf. Everything else looks good to me.

arcade/sprite/base.py
einarf: BasicSprite supports basic AABB collision around the entire image. It's geat when you have things like simple tiles or anything that doesn't need any fancy point based collision.

pushfoo: I've removed the warning.

> BasicSprite don't support custom hit box 

My understanding is that it does, but with the following restrictions:

1. Only via a `Texture` instance created with a `hit_box_points specified` value since it copies the hitbox from the passed `Texture`
https://github.com/pythonarcade/arcade/blob/ee1bb2b4c49e2287ee4c1e4724a81d6c73390eb1/arcade/sprite/base.py#L55-L57

2. The hitbox points can't be changed afterward since the `texture` setter only swaps the `Texture` and not the actual local hitbox values
https://github.com/pythonarcade/arcade/blob/ee1bb2b4c49e2287ee4c1e4724a81d6c73390eb1/arcade/sprite/base.py#L385-L401

einarf: oh right. I think @Cleptomania very recently changed this with the new hit box types. There are lots of things related to hit boxes we still have to resolve.

`BasicSprite` must have a texture at least. We do some hacks in `Sprite` applying a default texture if none is supplied. It's kind of ugly, but a lot less ugly than the alternative.

einarf: `BasicSprite` doesn't have rotation at least. You can compare with `sprite.py`.

pvcraven (approved):

PR #1688, opened 2023-04-07
Remove duplicated / unreachable return from BasicSprite

pushfoo:
This PR removes a duplicated line found while following up on the comments for #1687


einarf:
We've had a few of these lately. @Cleptomania hitting some shortcut by accident 😆

PR #1690, opened 2023-04-07
Minor improvements to GUI system

eschan145:
The parameter docs look very ugly and compacted with the native RST formatting. Switching to Google-style would also leave me with more space to fit more characters per line.


arcade/gui/widgets/text.py
```diff
@@ -19,37 +19,51 @@
...
-    :param style: Not used.
+    :param str align: Horizontal alignment of text on a line. This only applies
+                      if a width is supplied. Valid options include ``"left"``,
+                      ``"center"`` or ``"right"``.
+    :param float dpi: Resolution of the fonts in the layout. Defaults to 96.
+    :param bool multiline: If enabled, a ``\\n`` will start a new line. A
+                           :py:class:`~arcade.gui.UITextWidget` with
+                           ``multiline`` of True is the same thing as
+                           a :py:class:`~arcade.gui.UITextArea`.
+
```
gran4: I think you should delete this space.

arcade/gui/widgets/text.py
```diff
@@ -185,6 +206,9 @@ def place_text(self,
 
     @property
     def text(self):
+        """Text of the widget. Modifying this repeatedly will cause significant
```
gran4: Move Text(the entire line) after """

arcade/gui/widgets/text.py
```diff
@@ -195,6 +219,10 @@ def text(self, value):
 
     @property
     def multiline(self):
+        """Multiline state of the widget. Newlines indicated with ``\n`` will
```
gran4: Move Multiline(the entire line) after """

arcade/gui/widgets/text.py
```diff
@@ -205,6 +233,9 @@ def multiline(self, value):
 
     @property
     def ui_label(self) -> UILabel:
+        """Internal :py:class:`~arcade.gui.UILabel` used for rendering the
```
gran4: Move Internal(the entire line) after """

pushfoo (commented):
Good catch on `Caret` being RGB-only in pyglet's source. That's something which needs to be addressed upstream. I'll give this PR a more thorough look later.

EDIT: I filed a ticket with pyglet checking for any blockers for making `Caret` take RGBA https://github.com/pyglet/pyglet/issues/805

arcade/gui/widgets/text.py
```diff
@@ -19,37 +19,50 @@
...
-    :param float font_size: size of font.
-    :param RGBA255 text_color: Color of font.
-    :param bool bold: Bold font style.
-    :param bool italic: Italic font style.
+    """A simple text label. This widget is meant to display user instructions or
+    information. This label supports multiline text.
+
+    If you want to make a scrollable viewing text box, use a
+    :py:class:`~arcade.gui.UITextArea`.
+
```
gran4, line 31: It looks like this section is creating a problem.

arcade/gui/widgets/text.py
```diff
@@ -371,23 +424,27 @@ def do_render(self, surface: Surface):
...
-    :param multiline: support for multiline
-    :param scroll_speed: speed of scrolling
-    :param size_hint: Tuple of floats (0.0-1.0), how much space of the parent should be requested
-    :param size_hint_min: min width and height in pixel
-    :param size_hint_max: max width and height in pixel
-    :param style: not used
+    A text area that allows users to view large documents of text by scrolling
+    the mouse.
+
+
```
gran4: I think this is the problem

gran4: There shouldn't be 2 spaces

gran4 (changes requested):

eruvanos:
@eschan145 For me it looks good in general, just fix the build error and we can merge it

pushfoo:
Something may be wrong with Sphinx here. The error still occurs locally after removing the top-level docstring from `UITextWidget`.

arcade/gui/widgets/text.py
```diff
@@ -195,6 +219,11 @@ def text(self, value):
 
     @property
     def multiline(self):
+        """
+        Multiline state of the widget. Newlines indicated with ``\n`` will only
+        be honored if this is set to true. If you want a scrollable text widget,
+        use :py:class:`~arcade.gui.TextArea`.
+        """
```
pushfoo:
```suggestion
        """
        Get or set the multiline mode.

        Newline characters (``"\\n"``) will only be honored when this is set to ``True``.
        If you want a scrollable text widget, please use :py:class:`~arcade.gui.UITextArea`
        instead.
        """
```
1. The unescaped `\n` is the root of the error sphinx is giving you, but it's not reporting it in the right place due to a bug in our current version
2. `TextArea` does not exist, but `UITextArea` does.
3. The summary line should be separate from the rest of the docstring as shown above.

pushfoo (changes requested):
1. It's actually the `\n` in `UITextWidget.multiline` causing the error; make it `\\n`
2. Our Sphinx version is broken, but I have a fix for that too (see #1745)
 
I have further suggestions, but the above should get this building & mergable. Having trouble fixing this isn't OP's fault.

doc/programming_guide/gui/concept.rst
```diff
@@ -127,99 +159,317 @@ Size hint support
...
++---------------------------+-----------------------------------------+
+|``UIMousePressEvent``      |``dx``, ``dy``, ``button``, ``modifiers``|
++---------------------------+-----------------------------------------+
+|``UIMouseDragEvent``       |``dx``, ``dy``                           |
++---------------------------+-----------------------------------------+
+|``UIMouseScrollEvent``     |``scroll_x``, ``scroll_y``               |
++---------------------------+-----------------------------------------+
+|``UIKeyEvent``             |``symbol``, ``modifiers``                |
++---------------------------+-----------------------------------------+
+|``UIKeyReleaseEvent``      |None                                     |
```
eruvanos: Should we mention, that UIKeyReleaseEvent inherits from UIKeyEvent and also has `symbol` and `modifiers`?

einarf:
Fixed the lingering issue with RGB just to check if things build. I'm guessing it should be `RGBOrA`, but I wasn't completely sure. @eruvanos knows?

eruvanos:
Finally merged, thanks for your work

PR #1681, opened 2023-04-05
Added __all__ to many files

gran4:
Attempts to fix https://github.com/pythonarcade/arcade/issues/1390.
Its unlikely to be perfect yet. I expect a bunch of tests to break

PR #1700, opened 2023-04-14
Add __all__ to scene

gran4:
Add __all__ to scene bc I messed it up in https://github.com/pythonarcade/arcade/pull/1681.


gran4:
Redundant to https://github.com/pythonarcade/arcade/pull/1681

Issue #1701, opened 2023-04-15
CI doesn't pick up multiple whitespace style issues


pushfoo:
####   Malformed blank lines in docstrings not detected
As in #1681, malformed blank docstrings lines with extra indentation aren't getting flagged.

```diff
"""
Docstring summary line
-
+    
More docstring text.
"""
```

#### Comment whitespace weirdness

This should demand the spacing be fixed?
```
    #NOTE: for speed becuase attribute look ups are slow.
```

### Expected behavior:
Linting should be able to detect that.

### Steps to reproduce/example code:
Make a PR with tighter restrictions on whitespace styling

pushfoo:
I made progress on this and will try to make a PR later tonight.

PR #1702, opened 2023-04-15
Clean _check_for_collision

gran4:
Fixed unpythonic code and made clearer variable names(some are up for debate).

### DIFF
--- a/arcade/sprite_list/collision.py
+++ b/arcade/sprite_list/collision.py
@@ -93,32 +93,28 @@ def _check_for_collision(sprite1: SpriteType, sprite2: SpriteType) -> bool:
     :returns: True if sprites overlap.
     :rtype: bool
     """
-    sprite1_position = sprite1._position
-    sprite1_width = sprite1._width
-    sprite1_height = sprite1._height
-    sprite2_position = sprite2._position
-    sprite2_width = sprite2._width
-    sprite2_height = sprite2._height
+
     radius_sum = (
-        (sprite1_width if sprite1_width > sprite1_height else sprite1_height)
-        + (sprite2_width if sprite2_width > sprite2_height else sprite2_height)
+        max(sprite1._width, sprite1._height)
+        + max(sprite2._width, sprite2._height)
     )
+    
     # Multiply by half of the theoretical max diagonal length for an estimation of distance
     radius_sum *= 0.71  # 1.42 / 2
-    radius_sum_x2 = radius_sum * radius_sum
+    radius_sum_sq = radius_sum * radius_sum
 
-    diff_x = sprite1_position[0] - sprite2_position[0]
-    diff_x2 = diff_x * diff_x
-    if diff_x2 > radius_sum_x2:
+    diff_x = sprite1_position[0] - sprite2._position[0]
+    diff_x_sq = diff_x * diff_x
+    if diff_x_sq > radius_sum_sq:
         return False
 
-    diff_y = sprite1_position[1] - sprite2_position[1]
-    diff_y2 = diff_y * diff_y
-    if diff_y2 > radius_sum_x2:
+    diff_y = sprite1_position[1] - sprite2._position[1]
+    diff_y_sq = diff_y * diff_y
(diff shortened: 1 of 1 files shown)


einarf:
This will affect the performance profile of the function. It was optimized to extract the the positions and size into new varaibles. @cspotcode should know more about this.

It's another example that we should have brifely discussed this first 😄

cspotcode:
Yeah, I benchmarked the code before and after, and this code made it about 20% faster.  Attribute lookups and subscripting in python are relatively slow, and this code runs a huge number of times per frame.  I should have added code comments to explain why we are intentionally using un-pythonic code.

I used the benchmarking script in /benchmarks/bench.sh to run the collision stress-test in /benchmarks/collisions/bench.py.  I shared benchmarking results in #1524.

However, your improved variable names, like `radius_sum_sq`, look good to me.

gran4:
So if I decrease the attributes look ups it should be fine.

arcade/sprite_list/collision.py
```diff
@@ -93,32 +93,36 @@ def _check_for_collision(sprite1: SpriteType, sprite2: SpriteType) -> bool:
...
     sprite1_position = sprite1._position
     sprite1_width = sprite1._width
     sprite1_height = sprite1._height
     sprite2_position = sprite2._position
     sprite2_width = sprite2._width
     sprite2_height = sprite2._height
+
     radius_sum = (
         (sprite1_width if sprite1_width > sprite1_height else sprite1_height)
         + (sprite2_width if sprite2_width > sprite2_height else sprite2_height)
```
pushfoo, line 107: @cspotcode What's the overhead of this vs `max` after 3.11's optimizations? It would be a little cleaner to use `max` if we can afford it.

cspotcode:
I forget, but I do remember doing my benchmarking on 3.11.  We could bench
again and find out.   `min()` is better than `math.min()` because the
latter has an extra dictionary lookup.

arcade/sprite_list/collision.py
cspotcode: I forget, but I do remember doing my benchmarking on 3.11.  We could bench again and find out.   `min()` is better than `math.min()` because the latter has an extra dictionary lookup.

arcade/sprite_list/collision.py
```diff
@@ -93,32 +93,36 @@ def _check_for_collision(sprite1: SpriteType, sprite2: SpriteType) -> bool:
     :returns: True if sprites overlap.
     :rtype: bool
     """
+    
+    #NOTE: for speed becuase attribute look ups are slow.
```
pvcraven, line 97: I surprised the linter doesn't flag this for needing a space after the #

arcade/sprite_list/collision.py
```diff
@@ -93,32 +93,36 @@ def _check_for_collision(sprite1: SpriteType, sprite2: SpriteType) -> bool:
...
-    diff_y2 = diff_y * diff_y
-    if diff_y2 > radius_sum_x2:
+    diff_y_sq = diff_y * diff_y
+    if diff_y_sq > radius_sum_sq:
         return False
 
-    distance = diff_x2 + diff_y2
-    if distance > radius_sum_x2:
+    distance = diff_x_sq + diff_y_sq
+    if distance > radius_sum_sq:
```
pvcraven, line 125: These are better names.

pvcraven (approved):

arcade/sprite_list/collision.py
pushfoo: We need to double-check our linting settings in general. I'll add this to #1701.

arcade/sprite_list/collision.py
Cleptomania: I’ll mark this as resolved for now. We can evaluate `min` and `max` performance in it’s own issue.

arcade/sprite_list/collision.py
Cleptomania: Marking this as resolved since we are evaluating our linting configuration separately.

PR #1708, opened 2023-04-21
Add / fix __all__ problems discovered by mypy and pyright

### LINKED ISSUE #1390: Missing __all__ in some modules prevents `mypy --strict` from raising useful diagnostics
## Bug Report

Summary: arcade is missing explicit `__all__` declarations in some files, which prevents `mypy --strict` from raising useful diagnostics.

Discussed on Discord here:
https://discord.com/channels/458662222697070613/696054807840030740/1037940854985343006

## System Info

```
Arcade 2.6.16
-------------
vendor: Intel
renderer: Intel(R) Iris(R) Xe Graphics
version: (4, 6)
python: 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)]
platform: win32
```

### Actual behavior:

Given the sample code below

`mypy --strict` raises a single diagnostic:
```
$ mypy --strict src
src\typechecking-test.py:5: error: "Sprite" has no attribute "drawww"; maybe "draw"?
Found 1 errors in 1 files (checked 1 source files)
```

### Expected behavior:

`mypy --strict` raises two diagnostics, one for sprite, one for sprite_list:
```
$ mypy --strict src
src\typechecking-test.py:5: error: "Sprite" has no attribute "drawww"; maybe "draw"?
src\typechecking-test.py:6: error: "SpriteList" has no attribute "drawww"; maybe "draw"?
Found 2 errors in 1 files (checked 1 source files)
```

Given this file `src/typechecking-test.py`:
```
import arcade

sprite = arcade.Sprite()
sprite_list = arcade.SpriteList()
sprite.drawww()  # Good; we get a typechecking error
sprite_list.drawww()  # Bad; we do not get a typechecking error
```

To fix the problem, open arcade's source code: ...site-packages/arcade/sprite_list/__init__.py
and add:

```
__all__ = ["SpriteList"]
```

### Explanation

`mypy --strict` enables `--no-implicit-reexport`, which means that it does not think SpriteList is exported, so it gets type `Any` and is not typechecked.

cspotcode:
Will fix #1390

Building off the great work in #1681, and using a combination of mypy and pyright, I discovered a few more missing or broken `__all__` declarations.

This PR's diff is misleading because it builds off of #1681.  The diff will shrink after #1681 merges.  These are the relevant changes:
https://github.com/cspotcode/arcade/compare/ae863d71159491b8726c5b1e1497fcf6085f3fa3..cspotcode_add__all__to_all

pyright's reportUnsupportedDunderAll tells us if `__all__` refers to something that doesn't exist, for example `"Texture"` instead of `"Texture2d"`

mypy's attr-defined tells us when something is imported yet the importee does not declare it in `__all__`


pushfoo (approved):
This looks good. I cloned it locally and merged development into it. Everything still passes.

The `__all__` in `arcade.gui.widgetes.layout` is a little long, but I think it's acceptable for now. We can split it into a multi-line version if more layouts get added.

cspotcode:
> The `__all__` in `arcade.gui.widgetes.layout` is a little long,...

To be honest, I'm so accustomed to a formatter modifying that stuff automatically, I didn't even think about.  I made it multiline

arcade/sprite/__init__.py
```diff
@@ -66,6 +66,5 @@ def load_animated_gif(resource_name) -> AnimatedTimeBasedSprite:
     "FACE_RIGHT",
     "FACE_UP",
     "FACE_DOWN",
-    "PymunkMixin",
-    "Pymunk",
```
pushfoo, line 70: There is a `PyMunk` imported at the top of this file, which this line may have been a misspelled reference to. Can you double check this?

pushfoo (changes requested):
I apologize for the double review, but I took a closer look at the sprite module after remembering a  pymunk experiment I did a while ago. It's worth double checking one of the entries in `__all__`.

arcade/sprite/__init__.py
cspotcode: Yeah, github is hiding it, but `"PyMunk"` actually appears earlier in this list. Come to look at it again, there's also `"SimpleSprite"` commented out, not sure why that is?

pushfoo: I thought I removed that line, but it seems you're going to be the one to do it. To my understanding, it refers to what became `BasicSprite`.

cspotcode: Nice, gimme those sweet sweet git blame points.

pushfoo (approved):
Looks good.

PR #1710, opened 2023-04-21
Fix sphinx incremental builds; add `make serve` to run sphinx-autobuild

cspotcode:
## Doc server w/automatic rebuild and livereload

`make serve` will:

- build the docs
- host a webserver for the docs
- watch filesystem for file changes
- when files change, docs will immediately rebuild and your browser will automatically reload the page

This is done with sphinx-autobuild

## Fixed incremental builds

Normal sphinx-build has also benefited, because this PR fixes incremental builds.  There were two problems:

### sitemap broke pickling

Sphinx accomplishes incremental builds by pickling its state, then reloading that pickled object when it starts the next build.  The sitemap plugin recently made a change that broke pickling.  This meant sphinx was forced to start from scratch every time.

This PR moves back to an older version of the sitemap plugin, until the fix is published to pypi: https://github.com/jdillard/sphinx-sitemap/pull/62

### util/ scripts always invalidated cache

When rewriting a file, even if you write the exact same contents to disk, you invalidate sphinx cache.  I suspect this is caused by changing a file's modification timestamp.

This PR tweaks the util/ scripts to use a virtual filesystem abstraction that avoids any unnecessary fs writes, if the regenerated file in memory matches what is already on disk.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ pytest tests/unit
 
 ### Building & Testing Documentation
 
-#### Livereload w/automatic rebuild
+#### Automatic Rebuild with Live Reload
 
 You can build & preview documentation locally using the following steps.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,31 @@ pytest tests/unit
 
 ### Building & Testing Documentation
 
+#### Livereload w/automatic rebuild
+
 You can build & preview documentation locally using the following steps.
+
+Change into the doc directory:
+```commandline
+cd doc
+```
+
+Run the doc build and host a webserver to preview:
+```commandline
+make serve
+```
+
+You can now open [http://localhost:8000](http://localhost:8000) in your browser to preview the doc.
+
+- The `build/html` directory will contain the generated website files.  
+- `sphinx-autobuild` will host a webserver to preview the docs.  
+- When source files are changed, it will automatically trigger a rebuild, and
+browser tabs will automatically refresh to show the updates.
+
+If you suspect the automatic rebuilds are failing to detect changes, you can
+run a simpler one-time build using the following instructions:
+
+#### One-time build
 
 Change into the doc directory:
 ```commandline
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,10 +101,8 @@ make serve
 
 You can now open [http://localhost:8000](http://localhost:8000) in your browser to preview the doc.
 
-- The `build/html` directory will contain the generated website files.  
-- `sphinx-autobuild` will host a webserver to preview the docs.  
-- When source files are changed, it will automatically trigger a rebuild, and
-browser tabs will automatically refresh to show the updates.
+The `build/html` directory will contain the generated website files.  When you change source files,
+it will automatically regenerate, and browser tabs will automatically refresh to show the updates.
 
 If you suspect the automatic rebuilds are failing to detect changes, you can
 run a simpler one-time build using the following instructions:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,18 +94,18 @@ Change into the doc directory:
 cd doc
 ```
 
-Run the doc build and host a webserver to preview:
+Run the doc build to build the web page files, and host a webserver to preview:
 ```commandline
 make serve
 ```
 
-You can now open [http://localhost:8000](http://localhost:8000) in your browser to preview the doc.
+You can now open [http://localhost:8000](http://localhost:8000) in your browser to preview the docs.
 
 The `build/html` directory will contain the generated website files.  When you change source files,
-it will automatically regenerate, and browser tabs will automatically refresh to show the updates.
+it will automatically regenerate, and browser tabs will automatically refresh to show your updates.
 
 If you suspect the automatic rebuilds are failing to detect changes, you can
-run a simpler one-time build using the following instructions:
+run a simpler one-time build using the following instructions.
 
 #### One-time build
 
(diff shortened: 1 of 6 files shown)


pvcraven:
* Might want to add a comment ahead of the library so someone (like me) doesn't re-up that library version.
* Might need to add docs for 'make serve' somewhere?
* Does this work for Mac/Linux both? Maybe WSL?

cspotcode:
I added a comment about the sitemap version and updated `CONTRIBUTING.md` to describe how to run `make serve`.  Second opinion welcome on the docs, I wasn't sure whether to fully replace what was there, so I put `make serve` first and left the old one as a fallback.

> Does this work for Mac/Linux both? Maybe WSL?

I actually tested this on Windows, not WSL.  I have make installed via `scoop install make` https://scoop.sh/#/apps?q=make&s=0&d=1&o=true

I can test in WSL2, but I don't have access to a Mac.

cspotcode:
I also fixed an unrelated bug where `create_resource_listing.py` was trying to import the globally installed `arcade` instead of the local one.  I suspect that devs usually do not notice this because they have already done `pip install -e .[dev]`, so any `import arcade` already redirects to the local copy.

However, I work with multiple `git worktree`s, so it's important for the scripts to always import their own local version of the sources, whatever that is.

CONTRIBUTING.md
```diff
@@ -85,13 +85,37 @@ pytest tests/unit
...
 
 Change into the doc directory:
 ```commandline
 cd doc
 ```
 
+Run the doc build and host a webserver to preview:
+```commandline
+make serve
+```
```
pushfoo: Before re-installing the package, this is broken on my system after running `make clean`:
```console
$ make serve
sphinx-autobuild --watch ../arcade -b html -d build/doctrees   . build/html
make: sphinx-autobuild: No such file or directory
make: *** [Makefile:62: serve] Error 127
```
It works after.

CONTRIBUTING.md
```diff
@@ -85,13 +85,37 @@ pytest tests/unit
...
+
+You can now open [http://localhost:8000](http://localhost:8000) in your browser to preview the doc.
+
+- The `build/html` directory will contain the generated website files.  
+- `sphinx-autobuild` will host a webserver to preview the docs.
```
pushfoo: Is this supposed to be an explanation of how this works, or instructions? If I'm confused be phrasing, less experienced contributors will be too.

If it's an explanation, I would  remove ambiguity by rephrasing it or removing it.

If it's supposed to be  an instruction step, it's broken both before and after re-installing. 

Before:
```console
$ sphinx-autobuild
bash: sphinx-autobuild: command not found
```

After:
```
$ sphinx-autobuild
usage: sphinx-autobuild [-h] [--port PORT] [--host HOST] [--re-ignore RE_IGNORE] [--ignore IGNORE]
                        [--no-initial] [--open-browser] [--delay DELAY] [--watch DIR]
                        [--pre-build COMMAND] [--version]
                        sourcedir outdir [filenames ...]
sphinx-autobuild: error: the following arguments are required: sourcedir, outdir, filenames
```

CONTRIBUTING.md
```diff
@@ -85,13 +85,37 @@ pytest tests/unit
 
 ### Building & Testing Documentation
 
+#### Livereload w/automatic rebuild
```
pushfoo:
```suggestion
### Automatic Rebuild with Live Reload
```

pushfoo (changes requested):
tl;dr `make serve` works after re-installing under Linux (Debian 11), but the instructions are currently ambiguous and hard to follow for beginners.

First, thank you for working on this. It will be a massive improvement to the current development & build process. I tested it on Debian Linux. 

However, I think we need to make the following changes:
1. Clarify some of the instructions
2. Add instructions on re-installing arcade, and make an announcement for contributors on the Discord server reminding them to re-install the package

einarf:
Can't we just make this into a pythons script like https://github.com/pyglet/pyglet/blob/master/make.py

CONTRIBUTING.md
cspotcode: I made this change but kept the fourth-level header `####` unless you wanted to make them third-level headers instead?  Would have to also change the other one.

CONTRIBUTING.md
cspotcode: Is this because `sphinx-autobuild` was not installed, or some other issue?

For a first-time contributor, they already had to run `pip install -e .[dev]`, otherwise `sphinx-build` would not be installed.  So is it sufficient to rely on the existing setup instructions?

CONTRIBUTING.md
cspotcode: Good point, I tried to clear it up, removing mention of `sphinx-autobuild` since I see the existing docs don't mention sphinx.
https://github.com/cspotcode/arcade/blob/doc-serve-and-incremental-builds/CONTRIBUTING.md#automatic-rebuild-with-live-reload

Please let me know if you have any suggestions.

cspotcode:
> Can't we just make this into a pythons script like https://github.com/pyglet/pyglet/blob/master/make.py

Should we move that to a Discord thread?  I have some opinions about improvements to arcade's build scripts.  However, for this PR, I'm trying to embrace what's already there, under the assumption that the team has chosen the current tooling for a good reason.

pushfoo (approved):
tl;dr works locally after reinstalling; anyone writing doc needs this yesterday

I think benefits of < 10 sec rebuild are worth helping a few people update their doc build setup

CONTRIBUTING.md
pushfoo: I think it's because `make clean` deletes the `build` folder, but I can no longer replicate this with the latest changes after re-install, so I consider it resolved.

PR #1714, opened 2023-04-23
Port documentation makefile to .py

cspotcode:
This will need to be rebased after #1710, since it branches off of #1710.  Or we can merge this, bring in both PRs at once.


pushfoo:
This works as-is on my system (Linux).

I think someone other than the contributors should give feedback on the following:
1. Are the changes to the variable names at the top ok (`s/PAPER/PAPER_SIZE/`)?
2. Should we still provide a stub makefile for backward compatibility with programmer habits / brevity?

Cleptomania:
I think providing a Makefile in addition to this is going to cause problems. It just adds something else to maintain and creates a divergence in how Arcade is developed. Having everyone using the same process is much simpler.

cspotcode:
As I imagine it, the stub `Makefile` would be no more than:
```
.PHONY html
html:
    python ./make.py html
# repeat...
```
It would be emitted by `make.py` at startup, but also committed to git, so would be kept in sync automatically.

PR #1720, opened 2023-04-24
Add missing type hints

gran4:
Ready to merge. @einarf the errors are either things like variable types undeclared or something I need help on. So pls review. Look at the windows test for the weird error I do not get.


arcade/application.py
```diff
@@ -31,7 +31,7 @@
 _window: 'Window'
 
 
-def get_screens():
+def get_screens() -> []:
```
cspotcode: Should this annotation be `list`?  I'm unfamiliar with using `[]` syntax for a hint, is that supported?

arcade/application.py
```diff
@@ -808,43 +808,43 @@ def flip(self):
...
         """ Maximize  the window. """
         super().maximize()
 
-    def set_vsync(self, vsync: bool):
+    def set_vsync(self, vsync: bool) -> None:
         """ Set if we sync our draws to the monitors vertical sync rate. """
         super().set_vsync(vsync)
 
-    def set_mouse_platform_visible(self, platform_visible=None):
+    def set_mouse_platform_visible(self, platform_visible: bool=None) -> None:
```
cspotcode: Should this be `Optional[bool]` since the default value is `None`?  I think otherwise, typecheckers may raise a diagnostic.

arcade/application.py
```diff
@@ -859,23 +859,23 @@ def set_mouse_platform_visible(self, platform_visible=None):
...
         """ Capture the mouse. """
         super().set_exclusive_mouse(exclusive)
 
-    def set_exclusive_keyboard(self, exclusive=True):
+    def set_exclusive_keyboard(self, exclusive: bool=True) -> None:
         """ Capture all keyboard input. """
         super().set_exclusive_keyboard(exclusive)
 
-    def get_system_mouse_cursor(self, name):
+    def get_system_mouse_cursor(self, name: str) -> Optional["DefualtMouseCursor", "XlibMouseCursor"]:
```
cspotcode: Do `DefaultMouseCursor` and `XlibMouseCursor` need to be imported from somewhere?

arcade/application.py
```diff
@@ -957,7 +957,7 @@ def has_sections(self) -> bool:
         else:
             return self.section_manager.has_sections
 
-    def add_section(self, section, at_index: Optional[int] = None, at_draw_order: Optional[int] = None) -> None:
+    def add_section(self, section: Optional[SectionManager], at_index: Optional[int] = None, at_draw_order: Optional[int] = None) -> None:
```
cspotcode:
```suggestion
    def add_section(self, section: Optional[SectionManager] = None, at_index: Optional[int] = None, at_draw_order: Optional[int] = None) -> None:
```

arcade/context.py
```diff
@@ -42,7 +42,7 @@ class ArcadeContext(Context):
 
     atlas_size = 512, 512
 
-    def __init__(self, window: pyglet.window.Window, gc_mode: str = "context_gc", gl_api: str = "gl"):
+    def __init__(self, window: pyglet.window.Window, gc_mode: str = "context_gc", gl_api: str = "gl") -> None:
```
cspotcode, line 46: I thought `__init__` methods never needed a return type annotation.

arcade/gui/property.py
```diff
@@ -11,7 +11,7 @@ class _Obs:
 
     __slots__ = "value", "listeners"
 
-    def __init__(self):
+    def __init__(self) -> None:
```
cspotcode: Ditto earlier question about whether or not `__init__` methods need to declare their return type.

arcade/sprite/sprite.py
```diff
@@ -201,7 +201,7 @@ def hit_box(self) -> HitBox:
         return self._hit_box
 
     @hit_box.setter
-    def hit_box(self, hit_box: Union[HitBox, RotatableHitBox]):
+    def hit_box(self, hit_box: Union[HitBox, RotatableHitBox]) -> None:
```
cspotcode, line 208: Do setters need a return type annotation?

arcade/application.py
gran4: Yeah, it should be.

arcade/application.py
gran4: Yeah, I guess, but it seems like a waste.

arcade/application.py
gran4: @cspotcode It might break some code. I don't know if we should do that.

arcade/context.py
gran4: Well code automatically does it so I just thought I would do it on __init__ as well. Not sure about the convention for that though.

gran4: It seems like it should be -> None.
Look at:
https://stackoverflow.com/questions/46779046/correct-type-annotation-for-init

arcade/sprite/sprite.py
gran4: I don't know look at:
https://stackoverflow.com/questions/66595278/what-should-the-type-annotation-for-a-python-property-setters-argument-be
Its a small Stack Over Flow Question so I don't know if it counts.

arcade/application.py
gran4: It works, but yeah it should be list.

arcade/application.py
cspotcode: Oh right, it was required before, should be required after.

arcade/application.py
cspotcode: I think python accepts any expression, but that doesn't mean that a static typechecker will understand.  So someone might do `result = get_screens()` and then try to use `result` and their editor will not know what `result` or `result[0]` is supposed to look like.

arcade/application.py
cspotcode: Fair point.  I think we still need to import it, but we can wrap the import in `if TYPE_CHECKING:` where we `from typing import TYPE_CHECKING:`  If we don't import it, then static typecheckers won't understand the shape of the possible returned values.

cspotcode:
```
def get_system_mouse_cursor(self, name: str) -> Optional[Union[DefaultMouseCursor, XlibMouseCursor]]:
```

pushfoo (changes requested):
This is already getting to be a somewhat large PR. I'd turn this change into a separate PR as there may be additional changes related to the ticket.

gran4:
I'll keep those in becuase I don't know how to roll them back, but I'll put anything more into a different PR.
You know what.. ok.
There are a bunch of problems bc using Self.

eruvanos:
#1751 was merged, should we review this one, so it gets merged, too?

cspotcode:
Definitely, I can help review tonight.

cspotcode:
Looks like this PR has conflicts with the development branch that will need to be resolved.

arcade/sprite_list/sprite_list.py
```diff
@@ -1058,19 +1059,19 @@ def _grow_sprite_buffers(self):
...
-            self._sprite_pos_buf.orphan(double=True)
-            self._sprite_size_buf.orphan(double=True)
-            self._sprite_angle_buf.orphan(double=True)
-            self._sprite_color_buf.orphan(double=True)
-            self._sprite_texture_buf.orphan(double=True)
+            self._sprite_pos_buf.orphan(double=True)# type: ignore
+            self._sprite_size_buf.orphan(double=True)# type: ignore
+            self._sprite_angle_buf.orphan(double=True)# type: ignore
+            self._sprite_color_buf.orphan(double=True)# type: ignore
+            self._sprite_texture_buf.orphan(double=True)# type: ignore
```
cspotcode: Why do these need to be ignored?  We might be able to figure out a clever way around this.

arcade/sprite_list/sprite_list.py
```diff
@@ -957,16 +957,17 @@ def initialize(self):
         """
         self._init_deferred()
 
-    def draw(self, *, filter=None, pixelated=None, blend_function=None):
+    def draw(self, *, filter: Optional[int]=None, pixelated: bool=False, blend_function: Optional[int]=None) -> None:
```
cspotcode: Should `blend_function` be `Union[Tuple[int, int], Tuple[int, int, int, int], None]` since a few lines later we do:

```python
if blend_function is not None:
            self.ctx.blend_func = blend_function
        else:
            self.ctx.blend_func = self.ctx.BLEND_DEFAULT
```

And `self.ctx.blend_func` is typed as `Union[Tuple[int, int], Tuple[int, int, int, int]]`

arcade/sprite_list/sprite_list.py
cspotcode: I see, it's because each might be `None`.

For what it's worth, pyright will not raise these errors because I've got those diagnostics turned off.

arcade/perf_info.py
```diff
@@ -87,8 +87,7 @@ def print_timings():
         on_draw           60       0.0020
     """
     global _timings
-    print()
-    print("Event          Count Average Time")
+    print("\nEvent          Count Average Time")
```
cspotcode: Nit: personally, I prefer the code before.
This is also unrelated to the purpose of the PR: adding type hints.

arcade/gui/property.py
```diff
@@ -158,72 +171,72 @@ class DictProperty(Property):
...
-    def set(self, instance, value: dict):
+    def set(self, instance: Any, value: Dict) -> None:
         value = _ObservableDict(self, instance, value)
         super().set(instance, value)
 
 
 class _ObservableList(list):
     # Internal class to observe changes inside a native python list.
-    def __init__(self, prop: Property, instance, *largs):
+    def __init__(self, prop: Property, instance: Any, *largs: List) -> None:
```
cspotcode: Can we make `_ObservableList` a `Generic[T]` so that all these `Any`s can instead have a type `T`?

arcade/gui/property.py
```diff
@@ -113,37 +126,37 @@ class MyObject:
 
 class _ObservableDict(dict):
     # Internal class to observe changes inside a native python dict.
-    def __init__(self, prop: Property, instance, *largs):
+    def __init__(self, prop: Property, instance, *largs) -> None:
```
cspotcode, line 130: Can we make `_ObservableDict` a `Generic[T]` so that the `Any`s below can instead be `T` or similar?

arcade/gui/property.py
```diff
@@ -36,7 +49,7 @@ def __init__(self, default: Optional[P] = None, default_factory: Optional[Callab
         self.default_factory = default_factory
         self.obs: WeakKeyDictionary[Any, _Obs] = WeakKeyDictionary()
 
-    def _get_obs(self, instance) -> _Obs:
+    def _get_obs(self, instance: Any) -> _Obs:
```
cspotcode: Ditto my other questions (further down) about using `Generic[T]` to clean up some of these `Any`s.

arcade/easing.py
```diff
@@ -197,17 +201,20 @@ def ease_angle_update(easing_data: EasingData, delta_time: float) -> Tuple:
...
+                         ease_function: Callable=linear) -> EasingData:
     """
     Get an easing value
     """
-    if rate is not None:
-        diff = abs(start_value - end_value)
-        time = diff / rate
 
-    if time is None:
+    if not time or not rate:
```
cspotcode, line 211: I think we should keep the `is None` logic?  Because `not time` will happen if `time` is set to `0`, which I think is a valid value.

arcade/application.py
```diff
@@ -43,7 +44,7 @@
 ]
 
 
-def get_screens():
+def get_screens() -> list:
```
cspotcode: Nit: can be `List[Something]`?  I'm not sure what `Something` should be.

arcade/sprite_list/sprite_list.py
gran4: Yeah

arcade/sprite_list/sprite_list.py
gran4: Yeah

arcade/perf_info.py
gran4: Ok

arcade/easing.py
eruvanos: agree

arcade/gui/property.py
pushfoo: It may be best to do that in a separate PR.

arcade/gui/property.py
pushfoo: Same as with the dict, this is probably a separate PR.

arcade/gl/vertex_array.py
```diff
@@ -240,7 +240,7 @@ def _build(
 
     def render(
         self, mode: GLenumLike, first: int = 0, vertices: int = 0, instances: int = 1
-    ):
```
pushfoo, line 243:
```suggestion
    ) -> None:
```
I think you may have meant this.

arcade/gui/property.py
pushfoo: I'd leave the the generics as a separate PR.

arcade/math.py
pushfoo, line 1: There there are two better ways of handling the same-type return from `_Vec2`:
1. Use `Self` rather than the string version. `Self` is imported `from typing_extensions import Self`
2. Add `from __future__ import annotations` to the top of the file and drop the quotes

arcade/application.py
pushfoo: It needs to be `List` rather than `list` since earlier python versions don't support the latter.

The following seems to work on my system after adding `from pyglet.canvas import Screen` to the imports:
```suggestion
def get_screens() -> List[Screen]:
```

arcade/gui/property.py
```diff
@@ -1,6 +1,19 @@
...
+    Optional,
+    TypeVar,
+    Generic,
+    Tuple,
+    Callable,
+    Any,
+    Iterable,
+    Dict,
+    List,
+    Set,
```
pushfoo, line 14:
```suggestion
    Set,
    SupportsIndex,
```
You're missing this import & it's causing an undefined symbol error lower down in the file.

pushfoo (changes requested):
There's a grammatical error introduced in `vertex_array.py` which needs to be addressed. 

Most of the other changes are good in concept, but some need refinement. I would avoid adding type hints to more files in this PR, and would focus on refining what you've already started.

gran4:
Can you explain to me why I should leave generics in a separate PR?

pushfoo:
> Can you explain to me why I should leave generics in a separate PR?

I'd like to get this PR merged sooner rather than later because:

1. Some of these annotations will be very nice to have.
2. It includes changes to some areas I'd like to work on without risking merge conflicts
3. Generics can be tricky to get right

This PR has already been in progress for a month, and I do not want to delay it further by adding more changes to an already-large PR.

arcade/gui/property.py
gran4: Should I remove that from this PR then?

eruvanos:
@gran4 @pushfoo @cspotcode Just mention me, if you need any help or if you thing we are ready for merging.

cspotcode:
@pushfoo @eruvanos from my perspective, as soon as @pushfoo's feedback is addressed, this PR is ready to merge.  I'm sure if anything is wrong, CI will tell us about it soon enough.

Agreed that we should not add anything new to this PR.

arcade/gui/property.py
```diff
@@ -158,72 +172,72 @@ class DictProperty(Property):
...
         list.__setitem__(self, key, value)
         self.dispatch()
 
-    def __delitem__(self, key):
+    def __delitem__(self, key) -> None:
         list.__delitem__(self, key)
         self.dispatch()
 
-    def __iadd__(self, *largs):  # type: ignore
+    def __iadd__(self, *largs: List) -> "_ObservableList":  # type: ignore
```
cspotcode: Should the `List` annotation on `*largs` be replaced with `Iterable`?  Since as far as I can tell, `list.__iadd__` accepts `Iterable`, same as `list.extend` which you've also annotated with `Iterable`.

image

cspotcode:
This should have the latest development branch merged into it.

Looking at the remaining errors:

```
arcade/math.py:199: error: Need type annotation for "vel"  [var-annotated]
arcade/math.py:237: error: Static methods cannot use Self type  [misc]
arcade/math.py:237: error: A function returning TypeVar should receive at least one argument containing the same TypeVar  [type-var]
arcade/math.py:237: note: Consider using the upper bound "_Vec2" instead
arcade/math.py:239: error: Incompatible return value type (got "_Vec2", expected "Self")  [return-value]
arcade/math.py:242: error: Incompatible return value type (got "_Vec2", expected "Self")  [return-value]
arcade/math.py:245: error: Incompatible return value type (got "_Vec2", expected "Self")  [return-value]
arcade/math.py:248: error: Incompatible return value type (got "_Vec2", expected "Self")  [return-value]
arcade/math.py:251: error: Incompatible return value type (got "_Vec2", expected "Self")  [return-value]
```

@pushfoo On `Vec2`, it turns out that using `Self` as return type is actually wrong.  If you subclass `Vec2` as `MyVec2` and then trigger `__add__` via `MyVec2(0, 0) + MyVec2(0, 0)` then what you get is a `Vec2`, *not* a `MyVec2`.

The correct return type annotation is thus `Vec2`, not `Self`.  If we think this is a bug in `Vec2`, then that will happen in a separate PR.  The goal for this PR should be for tests and linting to pass, which means changing the annotation to `Vec2` or removing the annotation.

---

```
arcade/application.py:47: error: Name "Screen" is not defined  [name-defined]
```

Can be solved by adding `Screen` to imports: `from pyglet.canvas.base import ScreenMode, Screen`

---

```
arcade/sprite_list/sprite_list.py:1065: error: Item "None" of "Optional[Buffer]" has no attribute "orphan"  [union-attr]
arcade/sprite_list/sprite_list.py:1066: error: Item "None" of "Optional[Buffer]" has no attribute "orphan"  [union-attr]
arcade/sprite_list/sprite_list.py:1067: error: Item "None" of "Optional[Buffer]" has no attribute "orphan"  [union-attr]
arcade/sprite_list/sprite_list.py:1068: error: Item "None" of "Optional[Buffer]" has no attribute "orphan"  [union-attr]
arcade/sprite_list/sprite_list.py:1069: error: Item "None" of "Optional[Buffer]" has no attribute "orphan"  [union-attr]
arcade/shape_list.py:121: error: Item "None" of "Optional[Geometry]" has no attribute "render"  [union-attr]
```

This class of error is suppressed in our pyright config, so I'm not sure how we want to handle this.  Turn off mypy?  Turn off `union-attr` diagnostics?  Add assertions?  Add `# type: ignore`?

pushfoo:
> The correct return type annotation is thus `Vec2`, not `Self`. 

Understood & agreed.

> If we think this is a bug in `Vec2`, then that will happen in a separate PR.

It's `_Vec2`, not `Vec2`, and that only makes your point stronger. The changes I could suggest to make this more flexible and inheritance safe are irrelevant since:

1. This is an internal class which only has a single current use: [the particle emitter](https://github.com/pythonarcade/arcade/blob/2583a89650d5bdd567f6babe7ac0625d18f622a0/arcade/particles/emitter.py#L135)
2. We might be able to replace the usage with a shader-based particle system or just use pyglet's `Vec2`.

Even if we don't replace this class and think it's useful, `Self` is mostly helpful for subclassing. However, having a stable internal class around could be useful to protect against API changes in pyglet's vectors.

> This class of error is suppressed in our pyright config, so I'm not sure how we want to handle this.

I'm not sure either. The `# type: ignore` is awful but it'll get things done quickly and run well. I know`assert` has been controversial before, so are any of the other alternatives performant?

gran4:
> I'm not sure either. The `# type: ignore` is awful but it'll get things done quickly and run well. I know`assert` has been controversial before, so are any of the other alternatives performant?

I added that for now.

einarf:
Still assuming this is a WIP

arcade/draw_commands.py
```diff
@@ -588,18 +579,18 @@ def draw_polygon_outline(point_list: PointList,
...
+        point2 = point_list[i + 1]
+        
+        points = get_points_for_thick_line(point1[0], point1[1], point2[0], point2[1], line_width)
+        triangle_point_list.extend([points[1], points[0], points[2], points[3]])
+    
+    # Handle the last segment (connecting back to the start)
+    point1 = point_list[-1]
+    point2 = point_list[0]
+    points = get_points_for_thick_line(point1[0], point1[1], point2[0], point2[1], line_width)
+    triangle_point_list.extend([points[1], points[0], points[2], points[3]])
```
pushfoo, line 593: 1. Why is this in a type hint PR?
2. Do we have unit tests covering these changes?

If this is a needed improvement. If it works, this should be in another, separate PR.

pushfoo (changes requested):

eruvanos:
Can not be merged right now, I changed it to be a draft.

einarf:
Thanks for the work on this, but this is sadly not possible to merge at this time.

I manually salvaged and merged this branch in: https://github.com/pythonarcade/arcade/pull/2143

PR #1721, opened 2023-04-24
Add `__len__` and `__delitem__` to Scene

tiffanyxiao:
For Issue[https://github.com/pythonarcade/arcade/issues/1651](https://github.com/pythonarcade/arcade/issues/1651)

## What was done

- Added `__len__` and `__delitem__` methods to `scene.py`. 
- Added tests for these new scene methods.

## How to test
- Run the pytests; both tests should pass (the individual tests are in `tests/unit/scene/test_scene_len_delitem.py`).

Special thanks to @Cleptomania for the hands on help at PyCon 2023!

### DIFF
--- a/arcade/scene.py
+++ b/arcade/scene.py
@@ -6,7 +6,7 @@ a name, as well as control the draw order. In addition it provides a
 helper function to create a Scene directly from a TileMap object.
 """
 
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 from arcade import Sprite, SpriteList
 from arcade.types import Color, RGBA255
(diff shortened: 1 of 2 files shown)


arcade/scene.py
```diff
@@ -29,6 +29,24 @@ def __init__(self) -> None:
...
+
+    def __delitem__(self, sprite_list: Union[str, SpriteList]) -> None:
+        """
+        Remove the SpriteList from `_sprite_lists` and `_name_mapping`.
+
+        :param Union[str, SpriteList] sprite_list: which SpriteList to delete - can
+        be either the name of the SpriteList or the SpriteList object.
+        """
+        if isinstance(sprite_list, str):
+            self.remove_sprite_list_by_name(sprite_list)
```
gran4, line 46: Maybe you could add an is instance check for an integer, because I'm sure some users will want that. Like: a animal simulator.

gran4: Maybe you could make a self.remove_sprite_list_by_index.

gran4 (commented):
Other than that it looks good to me.

arcade/scene.py
gran4: Too late I guess now.

Cleptomania: I would definitely be open to a follow up PR adding a case for integers to remove based on the list index.

gran4: @tiffanyxiao do you want to do this? If you don't I'll do it.

tiffanyxiao: Sure, I'll try it!

PR #1722, opened 2023-04-24
Add DocStrings to remove_sprite_list_by_object

gran4:


arcade/scene.py
```diff
@@ -240,6 +240,13 @@ def remove_sprite_list_by_name(
         del self._name_mapping[name]
 
     def remove_sprite_list_by_object(self, sprite_list: SpriteList) -> None:
+        """
+        Remove a SpriteList from the Scene.
+
+        This function serves to completely remove the SpriteList from the Scene.
+
+        :param str SpriteList: The SpriteList to remove.
```
Cleptomania: The typing/name for the `sprite_list` parameter here need corrected. The type should be `SpriteList` and the parameter name should be `sprite_list`

Cleptomania (changes requested):

arcade/scene.py
gran4: sry, lol

gran4: I changed the wrong thing.

Cleptomania (approved):

PR #1730, opened 2023-04-27
Simplify `use_spatial_hash` flag

codeguru42:
This changes the flag to a bool instead of Optional[bool]. The behavior should remain unchanged, even when the parameter is not passed in because the default is False.


einarf:
This breaks other code in arcade. Historically `None` was "make a choice for me", but that might not be needed any more. Check the failing tests and see if that can be cleaned up.

codeguru42:
@einarf There are some cases where a default of `None` makes sense. I am proposing that we directly set `use_spatial_hash` directly to `False` as the default.

I'll work on getting the tests to pass.

Cleptomania:
I think setting to False by default with no None option makes the most sense, we just need to update the typing for the other stuff in TileMap that passes down to that.

We should probably also put this in the breaking changes section of the release notes. While it’s not necessarily a big breaking thing, there could be users passing None to this(and I think functionally that would still work, but type checkers would complain).

codeguru42:
@Cleptomania Thanks for the input. I pushed the change last night and am looking at the failing tests now. I love that we have those to catch these kinds of things. I think I've updated all the places in tilemap now.

Also, I hadn't considered anyone passing an explicit `None` as the value. I think you have the right idea. ~~Do I need to note the breaking change in some file on this branch? Or how is that handled in our documentation?~~ Nevermind, I found the release notes.

einarf:
> I think setting to False by default with no None option makes the most sense,

Agree. I don't think we're doing the auto select thing any more. We usually fall back to gpu collision.

arcade/sprite_list/sprite_list.py
```diff
@@ -156,7 +156,7 @@ def __init__(
 
         self._spatial_hash_cell_size = spatial_hash_cell_size
         self.spatial_hash: Optional[SpatialHash] = None
-        if use_spatial_hash is True:
```
codeguru42, line 159: @einarf In re to your previous comment. The original version of the condition here evaluates to `False` for both `None` and `False`. My proposed change simplifies that, both making the code shorter as well as using a the boolean value directly. My assumption is that when `use_spatial_hash` defaults to `False`, this will fall back to checking collision of an object with every other object.

codeguru42:
@einarf @Cleptomania This is ready for review now that tests are passing.

arcade/sprite_list/sprite_list.py
einarf: This is reasonable 👍

einarf:
This change is fairly straight forward. It's simpy just removing a legacy option `None`, so I'll just merge.

Thanks for the contribution!

PR #1731, opened 2023-04-27
Improve readability & controls of sprite depth example

pushfoo:
Follow-up to reading https://github.com/pythonarcade/arcade/issues/1729#issuecomment-1524701804

## Changes

* Move sprite depth demo to a clearer file name
* Add a top-level docstring including run instructions per arcade's conventions
* Move hard-coded values into adjustable constants
* Add space bar toggle for depth testing
* Add on-screen instruction / status display using batched text matching other examples
* Add comments

## How to test

```console
python -m arcade.experimental.sprite_depth_cosine
```

PR #1732, opened 2023-04-27
Fix 3D sphere example always saying wireframe is False

pushfoo:
Follow-up to issues discovered during #1731

## Changes

* Make the wireframe label update from `window.wireframe` instead of `window.ctx.wireframe`, which is always left as `False` by `on_draw`'s text drawing
* Only update the relevant label during a keypress instead of  updating all of them
* Fix initial capitalization for cull face's label

## How to test

```console
python -m arcade.examples.gl.3d_sphere
```


einarf:
👍

PR #1738, opened 2023-04-29
Typechecking improvements

cspotcode:
This PR includes several things, some of which might be rejected.  But they're related closely enough, I thought it easier to open one PR for feedback, then modify to remove the stuff that doesn't pass review.

### adds pyright typechecking

`make pyright`

Configured by `pyrightconfig.json`.  The config can technically go into `pyproject.toml`, but the json config has nice editor tab-completion which makes it easier to tweak.

### Adds auto-generated `Makefile` and `justfile`

```
# All 3x do the same thing
python ./make.py lint
make lint
just lint
```

Probably we don't want both of these, but just for sake of demonstration.  When `make.py` starts, it automatically re-writes these files.  Then they're committed to git.  So they will *only* change if you yourself have made changes to `make.py`, and then you can commit the changes to git.

Nice thing about justfile: you can be in any directory and type `just lint`.  It will always run commands from the project root.

### Splits lint logs on CI

ruff, mypy, and pyright logs appear as 3x separate sections in Github Actions results.  I wired it up so that all 3 will always run even if an earlier one fails.

### `make.py --help` organized into sections

A nice feature of Typer: you can organize the subcommands into sections.  I split them into `Docs` and `Coding`; that name's not great.  Maybe `Dev`?

image

---

### Reviewing pyright diagnostics

If you're using VSCode and want to review the outstanding VSCode diagnostics to decide which we fix and which we disable, you can use this `.vscode/settings.json`

```
{
    "python.analysis.diagnosticMode": "workspace",
    "python.languageServer": "Pylance"
}
```

Then open the Problems pane and sort by the diagnostic you want to see.

image

PR #1739, opened 2023-04-30
Improve make script ergonomics & documentation

pushfoo:
### Changes

* Document how to enable running `./make.py` instead of `python make.py` for Mac & Linux users
* Document how to enable tab completions for `./make.py`
* Add a `./make.py whichshell` sub-command to assist users with setting up tab completion

### How to test

Both sets of instructions assume you are on Linux or a Mac

#### Testing short-form invocation & Shell Checking

1. Follow the instructions for enabling `./make.py`
2. Run `./make.py whichshell` to make sure it reports your default shell correctly

#### Testing Shell Completion

1. If you are running a supported shell, follow the included instructions
2. Type `./make.py h`
3. Press tab
4. Observe the result; it should auto-complete to `./make.py html`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,9 @@ Upgrade by running the following command:
 pip install --upgrade pip
 ```
 
-### Optional: Enable running the build script with `./make.py` on Mac & Linux
+### Optional: Improve Ergonomics on Mac and Linux
+
+#### Enable `./make.py`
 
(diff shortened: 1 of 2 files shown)


pushfoo:
1. I think the shebang line and chmod doc is definitely worth keeping
2. We can revert or split off the bash completion part to avoid conflict with #1738
3. Please provide feedback on which direction would be best

PR #1741, opened 2023-04-30
Correct coordinate calculation for rectangle

### LINKED ISSUE #1740: Rectangle from Tiled are load at the wrong coordinate
## Bug Report
When loading a .tmx file, the coordinate of rectangle are wrongly 
## System Info

Arcade 2.6.17
-------------
vendor: NVIDIA Corporation
renderer: NVIDIA GeForce GTX 1060/PCIe/SSE2
version: (4, 6)
python: 3.11.3 (tags/v3.11.3:f3909b8, Apr  4 2023, 23:49:59) [MSC v.1934 64 bit (AMD64)]
platform: win32

### Actual behavior:
rectangle are transformed into polygon with negative y

### Expected behavior:
the coordinates of the rectangles consistent with the coordinates of all the rest.

Create a rectangle object in tiled, load the .tmx file into a scene.

vanicat:
The new computation
- scale the rectangle as anyother coordinate
- use self.tiled_map.map_size.height * self.tiled_map.tile_size[1] to offset the y coordinate, otherwise they are negative.

Fix  #1740

### DIFF
--- a/arcade/tilemap/tilemap.py
+++ b/arcade/tilemap/tilemap.py
@@ -918,12 +918,14 @@ class TileMap:
 
                     shape = [x + offset[0], y + offset[1]]
                 else:
-                    x = cur_object.coordinates.x + offset[0]
-                    y = cur_object.coordinates.y + offset[1]
-                    sx = x
-                    sy = -y
-                    ex = x + cur_object.size.width
-                    ey = -(y + cur_object.size.height)
+                    sx = cur_object.coordinates.x * scaling + offset[0]
+                    sy = (
+                        self.tiled_map.map_size.height * self.tiled_map.tile_size[1]
+                        - cur_object.coordinates.y
+                    ) * scaling + offset[1]
+
+                    ex = sx + cur_object.size.width * scaling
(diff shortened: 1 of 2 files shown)


pvcraven:
Looks like it didn't pass tests? Please research if you could.

gran4 (changes requested):
Re-Add the 2 lines(921 and 922):
`x = cur_object.coordinates.x + offset[0]`
and
`y = cur_object.coordinates.y + offset[1]`
You deleted. Without them, x and y don't exist.

vanicat:
I will look at the test after the end of Ludumdare,

vanicat:
I've updated the patch, and it now pass all the test.
The two line proposed by gran4 was not correct, because the offset must be apply after the scaling (at least, it is what happen for others coordinate computation).

Thanks.

gran4:
> I've updated the patch, and it now pass all the test. The two line proposed by gran4 was not correct, because the offset must be apply after the scaling (at least, it is what happen for others coordinate computation).

I didn't mean to fix the bug, I meant to get it to run.

gran4 (approved):

PR #1742, opened 2023-05-01
Fix test that erroneously calls itself instead of waiting for pytest to call it

cspotcode:
`test_sprite_gif` was calling itself.  This means a failure in that test would prevent any other tests from running.

You can verify this with `pytest --collect-only tests`  This tells pytest to discover the list of tests but not run any of them.  So it should not fail even if there are test failures.


pushfoo (approved):

gran4 (approved):

einarf:
Whops. That was me!

PR #1746, opened 2023-05-03
Improvements to make.py and pyright configuration

cspotcode:
I would have submitted as 2x PRs, but unfortunately one depends on the other and stacked PRs don't really work if you're submitting from a fork.

These changes are extracted from #1738 plus some things we talked about on Discord.  Hopefully this changeset is purely non-controversial stuff.

- adds `./make.py pyright` but does _not_ add it to `./make.py lint`
- improve pyright config, telling it how to typecheck deps that don't bundle types
- add categorization to typer subcommands
- tweak contributing.md since `make.py` gets execute bit automatically, no need to `chmod` manually

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,6 +175,8 @@ Tab completion would allow you to press tab to auto-complete the command:
 ./make.py html
 ```
 
+Note that this may interfere if you work on other projects that also have a `make.py` file.
+
 To enable this feature, most users can follow these steps:
 
 1. Run `./make.py whichshell` to find out what your default shell is
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,15 +141,9 @@ Be sure to re-run build & refresh to update after making changes!
 
 ## Optional: Improve Ergonomics on Mac and Linux
 
-### Enable `./make.py`
+### `./make.py` shorthand
 
-On Mac & Linux, you can enable running `make.py` using `./make.py` instead
-of `python make.py` as follows:
-
-1. Make sure you are in the root directory of the repo
-2. Run `chmod u+x make.py`
-
-You can run the make script with `./make.py` instead of `python make.py`.
+On Mac & Linux, you can run the make script as `./make.py` instead of `python make.py`.
 
 For example, this command:
 ```commandline
@@ -163,8 +157,7 @@ can now be run this way:
 
 ### Enable Shell Completions
 
-After enabling the short-form syntax as outlined above, you can also enable tab
-completion for commands on the following supported shells:
+On Mac & Linux, you can enable tab completion for commands on the following supported shells:
 
 * `bash` (the most common default shell)
 * `zsh`
--- a/make.py
+++ b/make.py
@@ -151,7 +151,7 @@ def run_doc(args: Union[str, List[str]]) -> None:
     run(args, cd=FULL_DOC_DIR)
 
 
-@app.command()
+@app.command(rich_help_panel="Docs")
 def clean():
     """
     Delete built website files.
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -141,15 +141,9 @@ Be sure to re-run build & refresh to update after making changes!
...
+### `./make.py` shorthand
 
-On Mac & Linux, you can enable running `make.py` using `./make.py` instead
-of `python make.py` as follows:
-
-1. Make sure you are in the root directory of the repo
-2. Run `chmod u+x make.py`
-
-You can run the make script with `./make.py` instead of `python make.py`.
+On Mac & Linux, you can run the make script as `./make.py` instead of `python make.py`.
```
pushfoo, line 146: I think it's worth including a warning about `--install-completion` applying to all `make.py` instances rather than just the local one.

make.py
```diff
@@ -229,7 +230,7 @@ def htmlhelp():
           f".hhp project file in {FULL_BUILD_PREFIX}/htmlhelp.")
 
 
-@app.command()
+@app.command(rich_help_panel="Docs")
 def qthelp():
```
pushfoo, line 234: For the less standard help commands such as this one, is there group these under `Additional Help Formats` or something similar?

make.py
```diff
@@ -409,47 +410,56 @@ def coverage():
...
+    """
+    Run all linting tasks: ruff and mypy
+    """
+    ruff()
+    mypy()
     print("Linting Complete.")
 
 
-@app.command()
+@app.command(rich_help_panel="Coding")
```
pushfoo: `"Code Quality"` would be clearer for this category.

make.py
```diff
@@ -409,47 +410,56 @@ def coverage():
...
+
+
+@app.command(rich_help_panel="Coding")
 def test_full():
     run([PYTEST, TESTDIR])
 
 
-@app.command()
+@app.command(rich_help_panel="Coding")
 def test():
```
pushfoo, line 463:
```suggestion
def test():
    """Run unit tests (Run this before making a pull request)!"""
````

File: make.py [hunk 5]
```diff
@@ -409,47 +410,56 @@ def coverage():
...
 
-@app.command()
+@app.command(rich_help_panel="Coding")
 def lint():
-    run([RUFF, *RUFFOPTS])
-    print("Ruff Finished.")
-    run([MYPY, *MYPYOPTS])
-    print("Mypy Finished.")
+    """
+    Run all linting tasks: ruff and mypy
```
pushfoo: If you can squeeze it in, I would add a nudge for users to run this before making a PR.

pushfoo (changes requested):
tl;dr this looks good overall, especially splitting the sections further, but it raises some new questions

In addition to the specific comments I made on the files, I have some questions before we merge. Please keep in mind:

* Not all of them need to be addressed in this PR
* Since I made the suggestions, I can pick up the slack on implementing some of them if you're busy

Questions:

1. Can we group the shell completion commands into a new `"Shell Completion"` group?
2. Can we rearrange the order of the command groups?
3. Can we hide or further control the extra format grouping?
4. Can we move `linkcheck` into `Coding` / `Code Quality`?

It think it's worth cleaning up the output further to be focused on what people use most frequently.  Some of the extra formats are definitely worth keeping (epub, PDF, single html, man, .txt), but not immediately relevant to most users.

cspotcode:
Review feedback addressed.

> It think it's worth cleaning up the output further to be focused on what people use most frequently.

Funny enough, for long output like this, if we put them at the top then they're likely to scroll off the top of the screen, leaving only the less important ones visible.

I'd like to merge this PR as-is, and follow-ups can re-order the output.

pushfoo (approved):
I think this is good, and I agree with the point about the current arrangement of the long output.

Shortening and rearranging the output requires further thought, so we can probably merge this as-is for now.

PR #1747, opened 2023-05-03
Add SlotsView Wrapper to View

### LINKED ISSUE #1743: Add `__slots__` to `arcade.View`

As talked about in Discord, a `__slots__` attribute could be added to `arcade.View`.

### What would it help with?

This will make it easier to specify which attributes belong to a class and should also improve memory performance.

gran4:
Fix https://github.com/pythonarcade/arcade/issues/1743. Basically View now inherits from slots view.

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -933,14 +933,9 @@ class View:
     """
     Support different views/screens in a window.
     """
-    __slots__ = (
-        "window",
-        "key",
-        "_section_manager",
-        "__weakref__"
-    )
     def __init__(self,
-                 window: Optional[Window] = None):
+                 window: Optional[Window] = None
+                 ) -> None:
 
         self.window = arcade.get_window() if window is None else window
         self.key: Optional[int] = None
@@ -1137,3 +1132,16 @@ class View:
         :param int y: y position of mouse
         """
         pass
+
+class SlotsView(View):
+    """Wrapper to View that adds __slots__"""
+    __slots__ = (
+        "window",
+        "key",
+        "_section_manager",
+        "__weakref__"
+    )
+    def __init__(self,
+                window: Optional[Window] = None
+                ) -> None:
+        super().__init__(window)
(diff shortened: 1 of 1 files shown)


arcade/application.py
```diff
@@ -934,7 +953,12 @@ class View:
     """
     Support different views/screens in a window.
     """
-
+    __slots__ = (
+        "window",
+        "key",
+        "_section_manager",
+        "section_manager",
```
gran4: @einarf , it seems that some variables are not created in the class, so it is having issues? What should I do? I can't find anythin on it.

gran4: Like section_manager here `conflicts with class variable`

einarf: Slots is only for the data members. Not for properties or methods. The actual data memers is what we create instance data from.

I would skip doing slots for the Window itself. Just focus on the View. The main reason for useing slots of to save some memory per instance (no internal member dict).

einarf (changes requested):

gran4:
Thats what I figured might be it.

gran4:
What do I do about `AttributeError: 'View' object attribute 'on_hide_view' is read-only`?
It is a function

pushfoo:
As I commented on the original issue and discord, I'm concerned by the following:
1. The effects this might have on mix-ins
2.  Running into memory issues from a large number of views suggests potential underlying issues with a project's design. 

@einarf Would the following be an acceptable middle ground?

1. Define a `ViewProtocol` type mirroring the API of the current `View` class
3. Keep the current view class as-is
4. Optionally, add a performance-oriented `SlotsView` base class

gran4:
I think, you should try to use slots, but I think I understand why someone would want to add values to View.

EDIT: so not 2? Do 3?

EDIT pt.2: Where is this going @pushfoo ???

gran4:
The wrapper will not work, so we need a different solution.

P.S: @pvcraven I believe this is the PR you were looking for about slots in View. I do not know if this will help at all, lol.

einarf:
I think this was not a good idea in the first place. It's easier to see the issues once the change have been made. I think my comments on discord was ill-advised.

@pushfoo 's idea about making a protocol is not a bad one I think. Could be that we need different view types in the future? I think we'll just close this is for now. It's a bit too radical at this point in time and we really need to get this 3.0 release out soon.

Thanks for giving it an attempt at least! It was still useful in some way.

PR #1748, opened 2023-05-05
Improve scene.py

gran4:
Add dunder methods bool and contains. 
Goes off: https://github.com/pythonarcade/arcade/issues/1651


arcade/scene.py
```diff
@@ -396,3 +396,14 @@ def draw_hit_boxes(
...
+    
+    def __bool__(self) -> bool:
+        """Returns whether or not _sprite_lists contains anything"""
+        return bool(self._sprite_lists)
+
+    def __contains__(self, item: Union[str, SpriteList]) -> bool:
+        """
+        Returns whether or not the item is in _sprite_lists 
+        or is a value in _name_mapping
+        """
```
pushfoo: Minor nitpick: I think you can reduce this to a a one-line docstring if you don't want to add anything additional lines.
```suggestion
"""True when `item` is in `_sprite_lists` or is a value in `_name_mapping`"""
```

pushfoo (changes requested):
tl;dr looking good; I think we can probably merge this after adding some tests :+1: 

### Current Changes

 `__bool__` will enhance code clarity in custom room / map loading views. Thank you for adding it. It's exactly what I'll need for a prototype at a later point.:+1: 

We can also probably merge `__contains__`, although I have some lingering out-of-scope API structure concerns. I'll try to get around to looking at `Scene` more closely soon. If we don't think of something clearer or more expressive, this will probably do.

### What to Add

Tests for should go in one or two appropriately named files in `tests/unit/scene`. I'd do the following:
1. Use the other files in that folder as test style reference, aside from the spacing
2. If you want, make a separate PR to touch up the spacing style for the existing tests

tl;dr of what you need to know to write tests
1. Tests of a specific part of an arcade component should go in `tests/unit` or an appropriate subfolder inside it
2. Functions starting with `test_` in those folders get run as tests
3. Exceptions, including those caused by `assert` receiving False, cause a test to fail
4. Keep the style simple and beginner-friendly
5. Try not to go overboard on the pytest fixtures unless you need to; they can make reasoning about code much harder

I've started working on a more in-depth Contributing Guide, and will try to expand on the above in a dedicated section or page.

gran4:
I'll try to base the tests off of other ones

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+        scene["Walls"]
+
+    del scene[0]
+
+    with pytest.raises(KeyError):
+        scene["Coins"]
+
+def test_bool():
+    scene = arcade.Scene()
+    assert bool(scene) == False
```
pushfoo: tl;dr although this works, the following is cleaner:
```suggestion
    assert not scene
```
`assert` handles non-`bool` data types based on whether they're [truthy or falsy](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false).  For example, `assert 0` will fail because 0 is treated as falsy, while `assert 1` will succeed.

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+
+    scene.add_sprite_list("Walls")
+    assert bool(scene) == True
```
pushfoo: Same comment as above, except you can drop the `not`.

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_contains():
+    scene = arcade.Scene()
+    "Walls" not in scene
+    None not in scene
```
pushfoo: These lines seem to be missing `assert` statements?
```suggestion
    assert "Walls" not in scene
    assert None not in scene
```

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+    assert Walls_SpriteList in scene
+
+
+    Coins_SpriteList = arcade.SpriteList()
+    scene.add_sprite_list("Coins", sprite_list = Coins_SpriteList)
+    assert "Coins" in scene
+
+    test_sprite = arcade.Sprite()
+    Coins_SpriteList.append(test_sprite)
+    assert Coins_SpriteList in scene
```
pushfoo: Python's variable names are expected to be in `lower_case_with_spaces_between_words`. For example, `Walls_Spritelist` should be `walls_spritelist` if you want to use the same spacing, or `walls_sprite_list`.

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+    scene.add_sprite_list("Walls")
+    assert bool(scene) == True
+
+def test_contains():
+    scene = arcade.Scene()
+    "Walls" not in scene
+    None not in scene
+    
+    Walls_SpriteList = arcade.SpriteList()
+    scene.add_sprite_list("Walls", sprite_list = Walls_SpriteList)
```
pushfoo: 1. You can drop the space around the `=` when passing keyword arguments
2. I'd add `use_spatial_hash=True` here to make sure we have that condition covered
```suggestion
    scene.add_sprite_list("Walls", use_spatial_hash=True, sprite_list=walls_sprite_list)
```

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+    scene = arcade.Scene()
+    scene.add_sprite_list("Player")
+
+    assert len(scene) == 1
+
+    scene.add_sprite_list("Walls")
+    scene.add_sprite_list("Coins")
+
+    assert len(scene) == 3
+
```
pushfoo, line 15: For best style, add an extra line here. Aside from imports and constant groupings, first-level module members conventionally have two lines of space between them.

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+    del scene[scene._sprite_lists[0]]
+
+    with pytest.raises(KeyError):
+        scene["Walls"]
+
+    del scene[0]
+
+    with pytest.raises(KeyError):
+        scene["Coins"]
+
```
pushfoo, line 36: Same spacing issue here.

tests/unit/scene/test_scene_dunder_methods.py
```diff
@@ -0,0 +1,65 @@
...
+def test_bool():
+    scene = arcade.Scene()
+    assert bool(scene) == False
+
+    scene.add_sprite_list("Walls")
+    assert bool(scene) == True
+
```
pushfoo, line 45: Needs another newline.

pushfoo (changes requested):
Renaming the file was a good choice. :+1:

tl;dr of the changes requested:

1. Two of the lines seem to be missing `assert` statements
2. It would be a good idea to try `spatial_hash=True` when adding the wall SpriteList
2. Misc style cleanup issues

pushfoo (approved):
Looks good after them most recent changes. :+1:

Issue #1707, opened 2023-04-20
Optimize HitBox Drawing


Cleptomania:
Arcade has the ability to draw hitboxes for a Sprite, SpriteList, or Scene. However, this function essentially loops on every sprite, and makes a call to `arcade.draw_line_strip` with the adjusted points for that Sprite's hitbox. This is incredibly inefficient and will completely destroy performance even with only 100-200 sprites.

There may be a better way to batch draw these, perhaps using Pyglet shapes is an opportunity. This will definitely need more investigation into the best ways to handle this.

einarf:
We can do a quick fix for 3.0 just looping and bulding a buffer of lines to improve this greatly. It's not a perfect solutuon, it will be dramatically faster.

einarf:
Made it about 20 times faster here. It should still be improved a lot more but I suspect triangle data for hitboxes could make this easier in the future.
https://github.com/pythonarcade/arcade/pull/2627

einarf:
I'll close this for now. The issue is linked in the collision revamp task. New task should exist for the rewrite of this

PR #1749, opened 2023-05-05
Improve hitbox drawing(https://github.com/pythonarcade/arcade/issues/1707)

gran4:
Addresses https://github.com/pythonarcade/arcade/issues/1707. Not sure if this is the best way to do it.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -39,6 +39,7 @@ class BasicSprite:
         "sprite_lists",
         "_angle",
         "__weakref__",
+        "Hitbox_outline",
     )
 
     def __init__(
@@ -57,6 +58,7 @@ class BasicSprite:
         self._scale = scale, scale
         self._color: Color = Color(255, 255, 255, 255)
         self.sprite_lists: List["SpriteList"] = []
+        self.Hitbox_outline = None
 
         # Core properties we don't use, but spritelist expects it
         self._angle = 0.0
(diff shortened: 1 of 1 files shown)


einarf:
This may be a little faster, but it doesn't solve the origonal problem. We don't care about how fast you can debug draw hit box data for a single sprite. We care about drawing speed for an entire sprite list.

It could be better trying to batch some geometry in spritelist,

(Also we removed the ShapeElementList caching of points in the spirte in 3.0. You're just adding back something that was too expensive memory wise and caused GC problems.

Again: You can save a lot of time by discussing things on discord first 😄

gran4:
I thought it was simple. Too simple for discord, but I guess I didn't ge the problem.

einarf:
Like already discussed on discord it needs to be something in the spritelist possibly caching hit box geometry all the sprites in the spritelist, but how do we know the hit box data is dirty? Also: should we just have a bool property to enable debug drawing instead of a separate problem,

@gran4 The reason your suggested solution doesn't really improve the situation much is that this is a GPU based library. Lots of draw class will never be fast no matter how much you optimize it. You might be able to draw 500-1000 hit box points per frame if you are lucky, but when there's 100k sprites in large tilemaps it's going to be slow. We need to batch draw these hit box points. It can be done with ShapeElementList, but again.. there are defniely some challenges here.

einarf:
The code you wanted to add in this PR is basically what we removed in the sprite revamp (3..0).. We want the base sprites to be as light as possible.

PR #1751, opened 2023-05-05
Address type issues

cspotcode:
Mypy is unable to typecheck against pyglet and other deps that don't include a `py.typed` file.  So it's unable to catch certain issues within arcade's own codebase. Pyright has an option telling it to use the source code of `pyglet` as types.  I couldn't find an equivalent option for mypy.

Yeah, some of the diagnostics are mere nuisances, unlikely to cause real problems.  But some are legit.

This PR tweaks the pyright configuration and also makes minor changes to code to address some of the diagnostics.  I'm not sure the best way to solicit feedback, ideally the GUI changes would be reviewed by someone familiar with the GUI code, etc.

I haven't fixed all the issues, but I want to pause this work till I get the team's temperature re: making these changes & enabling pyright on CI.  I don't want to impose unwelcome work on the team.


cspotcode:
Note to self: merge with #1720, check for conflicts and type errors.

gran4:
How would that work?

cspotcode:
I see that you are making some of the same improvements to types that I am; changing `int` to `float`.  In my case, I'm doing it because `pyright` is catching the mistakes.  I've improved the pyright configuration allowing it to catch stuff that it could not before:

https://github.com/pythonarcade/arcade/pull/1751/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

So I'm thinking we can somehow test your type changes against my improved pyright configuration, to see if they're compatible.

Is that what you were asking?

arcade/camera.py
```diff
@@ -384,7 +384,7 @@ def scale(self) -> Tuple[float, float]:
         return self._scale
 
     @scale.setter
-    def scale(self, new_scale: Tuple[int, int]) -> None:
+    def scale(self, new_scale: Tuple[int | float, int | float]) -> None:
```
eruvanos: The | operator requires Python 3.10
Arcade still supports 3.8, so we should use Usion.

Still wonder if float would be enough.

arcade/gl/buffer.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         elif reserve > 0:
             self._size = reserve
             # populate the buffer with zero byte values
-            data = (gl.GLubyte * self._size)(0)
+            data = (gl.GLubyte * self._size)()
```
eruvanos: @einarf Is this a bug?

arcade/gl/compute_shader.py
```diff
@@ -17,7 +17,7 @@ class ComputeShader:
     def __init__(self, ctx: "Context", glsl_source: str) -> None:
         self._ctx = ctx
         self._source = glsl_source
-        self._uniforms: Dict[str, Uniform] = dict()
+        self._uniforms: Dict[str, UniformBlock | Uniform] = dict()
```
eruvanos: Use Union

arcade/gui/examples/scroll_area.py
```diff
@@ -16,12 +16,13 @@
...
 from arcade.gui import UIManager, UIWidget, Property, Surface, UIDummy, UIEvent, bind, \
     UIMouseDragEvent, UIMouseScrollEvent, UIMouseEvent, UIBoxLayout, UIFlatButton, UIInputText
+from arcade.types import Point
 
 
 class UIScrollArea(UIWidget):
-    scroll_x = Property(default=0)
-    scroll_y = Property(default=0)
-    canvas_size = Property(default=(300, 300))
+    scroll_x = Property[float](default=0)
```
eruvanos: Also change the default to a float

arcade/gui/property.py
eruvanos, line 1: This looks great, thank you!
(Hope it works like I expect it :D)

eruvanos (changes requested):
I read through all files but in most gl parts I am unsure.
The changes for the GUI look good, especially if they work ;)

The `|` operator should not be used due to the Python 3.10 requirement.

eruvanos:
@pvcraven maybe it is time to move to pyright.
I tested it also during my work time and it works better (for us) then mypy.
It looks also like it has a better detection and found actual bugs, which mypy did not detect.

pvcraven:
That's fine by me. I'd started a while back fixing some things, glad to see some other people working on it.

pvcraven:
Yes, we don't want to use the | operator yet. It is awesome, I agree, but we still need python 3.9 compatibility.

arcade/gl/buffer.py
pvcraven: I also wonder about this one.

arcade/gui/examples/scroll_area.py
cspotcode: I can do that, but I recently learned that python typechecking standards allow `int` everywhere `float` is expected.  They call this the "numeric tower."

To be honest I'm not totally sure what issues occur when using `int` where a `float` is expected.  I know using `float` where `int` is expected can cause issues, not sure about the other way around.

https://peps.python.org/pep-0484/#the-numeric-tower

arcade/gl/buffer.py
cspotcode: Passing X positional args initializes the first X items of the array.  The remaining items initialize to `0`.  So passing `0` wasn't hurting, but its presence was confusing.

https://docs.python.org/3/library/ctypes.html#arrays

arcade/camera.py
cspotcode: Hmm you might be right, due to the "numeric tower."  I'll check

cspotcode:
This PR is ready to merge.

I've moved all the `arcade.gl` changes to another PR #1767 for @einarf(?) to review, to avoid blocking this PR from merging.

I've also addressed review feedback.  Linting and tests are passing locally, hopefully will be passing on CI too.

pvcraven:
Cool, we'll see if @eruvanos has a sec to double check, but I think his concerns were addressed.

arcade/tilemap/tilemap.py
```diff
@@ -411,6 +412,7 @@ def _create_sprite_from_tile(
 
         # --- Step 1, Find a reference to an image this is going to be based off of
         map_source = self.tiled_map.map_file
+        assert map_source
```
eruvanos: I see the idea, but asserts should not be used in production code,
it is seen as an Anti-Pattern.

please use an if raise block.

arcade/tilemap/tilemap.py
```diff
@@ -442,6 +444,8 @@ def _create_sprite_from_tile(
 
             # Can image_file be None?
             image_x, image_y, width, height = _get_image_info_from_tileset(tile)
+            assert width
+            assert height
```
eruvanos: Assert used in production code

eruvanos (changes requested):
Thanks for the changes, from my point of view, all remaining changes (beside the asserts) look good.
Will merge as soon as they are replaced.

Again thank you for all your work!

arcade/tilemap/tilemap.py
cspotcode: I'm not sure that anti-pattern heuristic applies here.  The code was already assuming that this value is never `None`.  Do we want to be littering our code with if-raise blocks every time the typechecker doesn't understand that a value can never be `None`?

I'm guessing that anti-pattern heuristic comes from two goals:
a) raise a meaningful error message
b) avoid disabling with `python -O`

However, in this case, both of those seem like _good_ things: avoid a performance hit when pleasing the typechecker, and avoid forcing developers to write exception messages when pleasing the typechecker.

What about this, to make our intent clear and avoid asserts in production?

```python
if TYPE_CHECKING:
    assert map_source
```

eruvanos: I get the argumentation, but if map_file can never be `None`, why do we have to fix the misleading optional here and not where it is created?

`self.tiled_map.map_file` should not return an Optional in that case.

cspotcode: I'm not sure, maybe @Cleptomania can weigh in here.

The optional comes from here:
https://github.com/pythonarcade/pytiled_parser/blob/47297a72c700f9b8cd558ef26420bc742b84f7f2/pytiled_parser/tiled_map.py#L72

In our case, `_create_sprite_from_tile` is called by `_process_object_layer` and `_process_tile_layer`, which are called by `_process_layer`, which is only called by `__init__`.  `__init__` accepts either a path to a map, or a pre-parsed map.
https://github.com/pythonarcade/arcade/blob/development/arcade/tilemap/tilemap.py#L108-L118

cspotcode: I asked @Cleptomania about this on Discord: https://discord.com/channels/458662222697070613/1105548182768865330/1106672539075362867

Based on that, I went with the `if TYPE_CHECKING: assert` and added comments about the guarantees that pytiled_parser makes despite the type hints.

cspotcode:
I created https://github.com/pythonarcade/pytiled_parser/issues/69 to track changes to type annotations in pytiled_parser which eliminate the need for `assert` in arcade.  Per Discord conversation, we can merge this PR now, which unblocks others. And we can remove the `if TYPE_CHECKING: assert`s in the future.

eruvanos (approved):
Great!

PR #1754, opened 2023-05-06
Run pyright on CI

cspotcode:
More stuff extracted from #1738. 

ruff, mypy, and pyright logs appear as 3x separate sections in Github Actions results. I wired it up so that all 3 will always run even if an earlier one fails.

CI runs these tasks by calling `make.py`.  This means any changes to make.py will be kept in sync with CI.

We might remove mypy from both `lint` and CI in the future, but that is not the purpose of this PR.


eruvanos:
I think this is a great opportunity to move to pyright! Thanks for your work so far.

eruvanos:
@pvcraven Are you fine to merge this PR, you are hosting the windows machine, and I don't want to merge code without your explicit approval :)

cspotcode:
This doesn't change what runs on his machine, right? And it's Ubuntu, not windows, right?

pushfoo:
> This doesn't change what runs on his machine, right?

From what I see in the PR, yes. I think that's all specified in the self hosted runner yaml.

> And it's Ubuntu, not windows, right?

If I remember correctly, the self hosted system is a Windows system with a full GL context on bare metal.

cspotcode:
Paul saying it's Ubuntu: https://discord.com/channels/458662222697070613/458662458198982676/1108194849494139032

eruvanos:
> From what I see in the PR, yes. I think that's all specified in the self hosted runner yaml.

Ah I see

cspotcode:
Based on the comments above, this is ready to merge.

Are there any other people who need to review, any other topics to discuss?

PR #1755, opened 2023-05-06
Remove face_point name in sprite

gran4:
@pvcraven request


gran4:
Empty

pvcraven:
I've thought about this a while, and I'm not wild about including move methods into the sprite. I regret having done this earlier. I think movement should be done by whatever physics engine controls the sprite. This assumes no physics engine.

gran4:
Ok, deleted it.

PR #1759, opened 2023-05-07
Moved moving platforms code

cacheguy:
Move the moving platforms before moving the player. That way, the player's movement is up to date with the platform's position.

Without this change, the player seems to float on or sink into the platform by a few pixels.

### DIFF
--- a/arcade/physics_engines.py
+++ b/arcade/physics_engines.py
@@ -399,8 +399,6 @@ class PhysicsEnginePlatformer:
 
         # print(f"Spot B ({self.player_sprite.center_x}, {self.player_sprite.center_y})")
 
-        complete_hit_list = _move_sprite(self.player_sprite, self.walls + self.platforms, ramp_up=True)
-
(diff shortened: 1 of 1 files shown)


pvcraven:
Hi, could you look at the failing test? I think the code has some issues with number of blank lines for PEP-8 compliance.

cspotcode:
We should also add a `python ./make.py fmt` to automatically fix formatting issues.

arcade/physics_engines.py
```diff
@@ -432,6 +430,8 @@ def update(self):
                             platform.change_y *= -1
 
                     platform.center_y += platform.change_y
+
```
gran4: Delete white space from line 433. There must be an indent there. @cacheguy

einarf:
This PR still needs the whitespace fix before merge.

einarf:
Fixed lingering issue. Thanks for the PR:

PR #1760, opened 2023-05-07
Add Docstrings to paths.py

gran4:
I'll add more later


gran4:
@pvcraven 
What is faster?

`x2 < self.left or x2 > self.right or y2 < self.bottom or y2 > self.top`

or 
`b in self.barriers`

bc I see both

gran4:
get_vertex_neighbours seems inefficient.

PR #1762, opened 2023-05-08
Add docs to paths.py

gran4:


gran4:
get_vertex_neighbours seems inefficient.
Should I add slots?

gran4:
Should I add __slots__?

eruvanos (commented):
Looks good to me, I am unsure about the usage of google docs style, but I also filed a request in discord.
I would be fine with the change

arcade/paths.py
```diff
@@ -18,16 +28,30 @@
 def _spot_is_blocked(position: Point,
                      moving_sprite: Sprite,
                      blocking_sprites: SpriteList) -> bool:
+    """
+    Return if position is blocked
+
+    Returns whether or not the moving_sprite is blocked
+    at position is blocked of the sprite in blocking_sprites
```
pushfoo: This is redundant and/or a typo.

arcade/paths.py
```diff
@@ -18,16 +28,30 @@
...
+
+    Args:
+        position: position to put moving_sprite at
+        moving_sprite: Sprite to use
+        blocking_sprites: List of Sprites to check against
+
+    Returns:
+        If the Sprite would hit anything in blocking_sprites at the position
```
pushfoo: I understand that you may prefer Google style, but I think it's better to use the current style for now. As einarf has stated before, we can use a script to convert all docstrings in one PR rather than scattering the style changes across multiple commits.

arcade/paths.py
```diff
@@ -47,6 +71,16 @@ def _heuristic(start: Point, goal: Point):
 
 class _AStarGraph(object):
     # Define a class board like grid with two barriers
+    """
+    A gird with barriers that is used in _AStarSearch
```
pushfoo:
```suggestion
    A grid which tracks barriers to movement.
```
There's a typo (s/gird/grid) in this line. I also think "that is used in _AStarSearch" is redundant here because the name of the class communicates the idea.

arcade/paths.py
```diff
@@ -18,16 +28,30 @@
...
     hit_list = check_for_collision_with_list(moving_sprite, blocking_sprites)
     moving_sprite.position = original_pos
     return len(hit_list) > 0
 
 
-def _heuristic(start: Point, goal: Point):
+def _heuristic(start: Point, goal: Point) -> float:
     """
-    Returns the heuristic of the 2 poitns
+    Returns the heuristic of the 2 points
```
pushfoo:
```suggestion
    Returns a heuristic value for the passed points.
```
The original docstring may be subtly incorrect, as may the method body. I would handle the docstring as above for now.

arcade/paths.py
```diff
@@ -71,6 +105,15 @@ def __init__(self, barriers: Union[List, Tuple, Set],
             self.movement_directions = (1, 0), (-1, 0), (0, 1), (0, -1)  # type: ignore
 
     def get_vertex_neighbours(self, pos: Point) -> List[Tuple[float, float]]:
+        """
+        Returns vertexes around the point
```
pushfoo:
```suggestion
Return neighbors for this point according to ``self.movement_directions``

These are not guaranteed to be reachable or valid points.
```
The implementation does not seem to guarantee the points are reachable, only that they match the accepted movement directions for this grid.

arcade/paths.py
```diff
@@ -71,6 +105,15 @@ def __init__(self, barriers: Union[List, Tuple, Set],
...
+
+        Args:
+            pos: Gets the points around the position.
```
pushfoo: This phrasing reads like a summary line for this method, not a description of what `pos` is.

arcade/paths.py
```diff
@@ -81,9 +124,22 @@ def get_vertex_neighbours(self, pos: Point) -> List[Tuple[float, float]]:
             n.append((x2, y2))
         return n
 
-    def move_cost(self, a: Point, b: Point):
+    def move_cost(self, a: Point, b: Point) -> float:
+        """
+        Returns a float of the cost to move
+
+        moving to corners costs more than to the side.
```
pushfoo:
```suggestion
        Moving diagonally costs more than along the X & Y axes.
```
1. Capitalization
2. It's not just side to side, but vertically as well.

arcade/paths.py
```diff
@@ -81,9 +124,22 @@ def get_vertex_neighbours(self, pos: Point) -> List[Tuple[float, float]]:
...
+        A barrier's cost is float("inf) so that that
+        the Algorithm will never go on it
+
+        Args:
+            a: The 1st point to compare
+            b: The 2nd point to compare
+
+        Returns:
+            The move cost of moving bettween of the 2 points
```
pushfoo: s/bettween/between/

arcade/paths.py
```diff
@@ -169,13 +239,15 @@ class AStarBarrierList:
...
-    :param int top: Top of playing field
+    Attributes:
+        moving_sprite (BasicSprite): Sprite that will be moving
+        blocking_sprites (SpriteList): Sprites that can block movement
+        grid_size (int): Size of the grid, in pixels
+        left (int): Left border of playing field
+        right (int): Right border of playing field
+        bottom (int): Bottom of playing field
+        top (int): Top of playing field
+        barrier_list (Optional[Set]): SpriteList of barriers to use in _AStarSearch, None if not recalculated
```
pushfoo: As mentioned earlier, I am not in favor of switching to Google doc string style in this PR.

arcade/paths.py
```diff
@@ -266,19 +341,23 @@ def astar_calculate_path(start_point: Point,
...
-    :param Point point_2: End position position
-    :param SpriteList walls: List of all blocking sprites
-    :param int max_distance: Max distance point 1 can see
-    :param int check_resolution: Check every x pixels for a sprite. Trade-off
-                                 between accuracy and speed.
+    Determine if we have line of sight between two points.
+
+    NOTE: Try to make sure that spatial hashing is enabled
+    on the wall SpriteList or this will be very slow.
+    NOTE: Make sure max_distance is positive.
```
pushfoo: I would remove this  line and add an if statement bounds check at the top of the function:
```py
    if max_distance <= 0:
        raise ValueError("max_distance must be greater than zero")
```

arcade/paths.py
```diff
@@ -266,19 +341,23 @@ def astar_calculate_path(start_point: Point,
 def has_line_of_sight(point_1: Point,
                       point_2: Point,
```
pushfoo: We may as well rename these to something clearer. Maybe something like `observer` and `target`?

arcade/paths.py
```diff
@@ -266,19 +341,23 @@ def astar_calculate_path(start_point: Point,
...
-    :param Point point_1: Start position
-    :param Point point_2: End position position
-    :param SpriteList walls: List of all blocking sprites
-    :param int max_distance: Max distance point 1 can see
-    :param int check_resolution: Check every x pixels for a sprite. Trade-off
-                                 between accuracy and speed.
+    Determine if we have line of sight between two points.
+
+    NOTE: Try to make sure that spatial hashing is enabled
+    on the wall SpriteList or this will be very slow.
```
pushfoo: A warning directive may render better.
```suggestion
    .. warning:: Try to make sure spatial hashing is enabled on ``walls``!

                 If spatial hashing is not enabled, this function may run
                 very slowly!
 ```

pushfoo (changes requested):
### tl;dr

1. Some of these changes are good
2. However, there are typos in the PR
3. We should wait on using Google style docstrings since it can break rendering

### On Google docstrings

I've  experimented with enabling Google docstrings locally & asked about next steps. I agree with einarf's recommendation that we should do the following:
1. Stick to our current docstring format for now
2. Use a conversion script to update docstrings in a dedicated PR when we're ready

### get_vertex_neighbors

> get_vertex_neighbours seems inefficient.

I agree there's room for some small optimizations here. However:
1. That would be oustide the scope of a doc-focused PR
2. The best approach to pathfinding is offloading it to non-Python code such as Rust or C++

### Should I add `__slots__`?

Probably not in this PR, and I'm not sure about doing so in general. I think our pathfinding code is worth considering for refactoring in general, but it needs some careful consideration on how to do it right. It could be better to put a warning label on it and say that it's a suboptimal prototyping tool, but I'm not yet sure.

gran4:
Well, I was asking:
get_vertex_neighbours seems inefficient.
Should I add slots?

not for this PR put in general.
I've already submitted a separate PR to add slots.

I'll change it to use the other style.

arcade/paths.py
gran4: Yeah, I don't think we need that.
`
    if max_distance <= 0:
        raise ValueError("max_distance must be greater than zero")
`

If it is negative, it will return false no matter what. How about a waring instead?

pushfoo: > If it is negative, it will return false no matter what.

Yes, and that will take time to debug if they're accidentally passing a negative value. It's better to fail loudly and clearly.

> How about a war[n]ing instead?

A warning might be hard to see in other logging output. An exception will be much clearer and prompt the user to fix their code.

einarf:
There are still some unresolved conversation here. What is the status on that?

gran4:
> There are still some unresolved conversation here. What is the status on that?

@einarf  sorry for the delay. I have been busy with finals, but school is over so I have time now. The PR is ready for review.

PR #1767, opened 2023-05-10
Fix pyright errors in `arcade.gl`

cspotcode:
Followup to #1751.  Will need a rebase after #1751 gets squashed.  Requesting a review from @einarf , but no rush, I know you're slowly getting back into things.

### DIFF
--- a/arcade/gl/context.py
+++ b/arcade/gl/context.py
@@ -1,12 +1,14 @@
 import logging
+import typing
 import weakref
 from collections import deque
 from contextlib import contextmanager
 from ctypes import c_char_p, c_float, c_int, cast
-from typing import (Any, Deque, Dict, Iterable, List, Optional, Sequence, Set, Tuple,
+from typing import (TYPE_CHECKING, Any, Deque, Dict, Iterable, List, Optional, Sequence, Set, Tuple,
                     Union)
 
 import pyglet
+import pyglet.gl.lib
 from pyglet import gl
 from pyglet.window import Window
 
@@ -17,7 +19,7 @@ from .glsl import ShaderSource
 from .program import Program
 from .query import Query
 from .texture import Texture2D
-from .types import BufferDescription
+from .types import BufferDescription, GLenumLike, PyGLenum
 from .vertex_array import Geometry
 from ..types import BufferProtocol
 
@@ -872,9 +874,9 @@ class Context:
         components: int = 4,
         dtype: str = "f1",
         data: Optional[BufferProtocol] = None,
-        wrap_x: Optional[gl.GLenum] = None,
-        wrap_y: Optional[gl.GLenum] = None,
-        filter: Optional[Tuple[gl.GLenum, gl.GLenum]] = None,
+        wrap_x: Optional[PyGLenum] = None,
+        wrap_y: Optional[PyGLenum] = None,
+        filter: Optional[Tuple[GLenumLike, GLenumLike]] = None,
         samples: int = 0,
         immutable: bool = False,
     ) -> Texture2D:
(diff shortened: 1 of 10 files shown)


arcade/cache/texture.py
```diff
@@ -91,13 +91,13 @@ def put(
         name = texture.cache_name
         if strong:
             self._strong_entries.put(name, texture)
-            if self._strong_file_entries.get(file_path):
+            if self._strong_file_entries.get(str(file_path)):
```
gran4: Shouldn't the type be str and not Union[str, Path] if you are correct?

arcade/gl/context.py
```diff
@@ -1,12 +1,14 @@
 import logging
+import typing
```
gran4: delete import typing

arcade/gl/program.py
```diff
@@ -93,7 +93,7 @@ def __init__(
                 f"Valid modes are: {self._valid_capture_modes}."
             )
 
-        shaders = [(vertex_shader, gl.GL_VERTEX_SHADER)]
+        shaders: list[tuple[str, int]] = [(vertex_shader, gl.GL_VERTEX_SHADER)]
```
gran4: Shouldn't this be List[Tuple[str, int]] or does it matter?

arcade/gl/types.py
```diff
@@ -266,7 +271,7 @@ def __init__(
                 f"attributes ({len(self.attributes)})"
             )
 
-        def zip_attrs(formats, attributes):
+        def zip_attrs(formats: list[str], attributes: Sequence[str]):
```
gran4: List not list, again, does it matter?

arcade/gui/examples/scroll_area.py
```diff
@@ -22,12 +22,13 @@
...
+from arcade.types import Point
 
 
 class UIScrollArea(UIWidget):
-    scroll_x = Property(default=0)
-    scroll_y = Property(default=0)
-    canvas_size = Property(default=(300, 300))
+    scroll_x = Property[float](default=0.0)
+    scroll_y = Property[float](default=0.0)
+    canvas_size = Property[Point](default=(300.0, 300.0))
```
gran4: I think you can remove [float], and stuff. Unless there is something I am misunderstanding

arcade/gui/property.py
```diff
@@ -31,22 +29,21 @@ class Property:
...
         self.default_factory = default_factory
-        self.obs = WeakKeyDictionary()
+        self.obs: WeakKeyDictionary[Any, _Obs] = WeakKeyDictionary()
 
     def _get_obs(self, instance) -> _Obs:
         obs = self.obs.get(instance)
         if obs is None:
-            obs = _Obs()
-            obs.value = self.default_factory(self, instance)
+            obs = _Obs(self.default_factory(self, instance))
```
gran4: I remember doing this file, we might have some conflicts

arcade/gui/property.py
```diff
@@ -77,9 +74,9 @@ def bind(self, instance, callback):
     def __set_name__(self, owner, name):
         self.name = name
 
-    def __get__(self, instance, owner):
+    def __get__(self, instance, owner) -> P:
         if instance is None:
-            return self
+            return self # type: ignore
```
gran4: Make it a Union[P, Property]?

arcade/gui/widgets/layout.py
```diff
@@ -662,10 +662,10 @@ def do_layout(self):
             [None for _ in range(self.column_count)] for _ in range(self.row_count)
         ]
 
-        max_width_per_column = [
+        max_width_per_column: list[list[tuple[float, int]]] = [
             [(0, 1) for _ in range(self.row_count)] for _ in range(self.column_count)
         ]
-        max_height_per_row = [
+        max_height_per_row: list[list[tuple[float, int]]] = [
             [(0, 1) for _ in range(self.column_count)] for _ in range(self.row_count)
```
gran4: List[List[Tuple[int, int]]] instead of list[list[tuple[int, int]]]

arcade/gui/widgets/slider.py
```diff
@@ -226,7 +226,7 @@ def do_render(self, surface: Surface):
         )
 
     def _cursor_pos(self) -> Tuple[int, int]:
-        return self.value_x, self.y + self.height // 2
+        return self.value_x, int(self.y + self.height // 2)
```
gran4: value_x looks like a float, so instead of making self.y + self.height // 2 an int, change -> Tuple[int, int]: to -> Tuple[float, float]:

arcade/texture/transforms.py
```diff
@@ -217,7 +217,7 @@ def transform_hit_box_points(
 # but it's faster to just pre-calculate it.
 # Key is the vertex order
 # Value is the orientation (flip_left_right, flip_top_down, rotation)
-ORIENTATIONS = {
+ORIENTATIONS: dict[tuple[int, int, int, int], tuple[int, bool, bool]] = {
```
gran4: Dict[Tuple[]]

cspotcode:
@gran4 Thanks for taking a look, but unfortunately, this PR can't be reviewed yet.  It's stacked on top of #1751 so the diff is all wonky.  It'll be reviewable once #1751 merges.

gran4:
Oh, ok you should make it a draft then. (unless there is something I don't understand)

arcade/cache/texture.py
einarf: I will redo all the cache stuff anyway, but this doesn't hurt.

einarf:
The gl stuff looks perfectly reasonable 👍

einarf (approved):

cspotcode:
@einarf Thanks for taking a look!  I marked this PR as draft while we wait for #1751 to merge.  Once #1751 is merged, I'll rebase this PR, bring it out of draft, and ping you for a merge.

arcade/gl/context.py
cspotcode: It was actually used in this file, `typing.cast()`.  I did this because `cast` already refers to something else.  But I think I should remove that bit anyway.

cspotcode:
I rebased yesterday, but there are still one or two mypy errors I need to understand and fix before this can merge.  I'll solicit a review after I've fixed them.

cspotcode:
@einarf I need some help understanding some of the gl types.

https://github.com/pythonarcade/arcade/blob/889772df9b9c94d77bbb09a500e9dcb0be180e53/arcade/gl/vertex_array.py#L195-L196
I get a typecheck diagnostic `Argument 1 to "gl_name" has incompatible type "Optional[int]"; expected "int"`
This is because `prog_attr.gl_type` is `Optional[int]`

Is that correct, should `AttribFormat.gl_type` be `Optional`?

`AttribFormat` is instantiated here:
https://github.com/pythonarcade/arcade/blob/889772df9b9c94d77bbb09a500e9dcb0be180e53/arcade/gl/program.py#L394-L401
and here:
https://github.com/pythonarcade/arcade/blob/889772df9b9c94d77bbb09a500e9dcb0be180e53/arcade/gl/types.py#L311-L314

For the latter, `gl_type` is possibly `None` because it comes from `_formats`, which has some `None` values in x1, x2, x4, x8
https://github.com/pythonarcade/arcade/blob/889772df9b9c94d77bbb09a500e9dcb0be180e53/arcade/gl/types.py#L196-L220

arcade/gl/types.py
cspotcode: Good catch, python 3.8 needs it capitalized.

PR #1768, opened 2023-05-10
Seperated Slider, added UITexturedSlider

gran4:
Let me know what you think. I don't want to have to add all the extra code when I implement TexturedSlider


eruvanos:
I will have a deeper look into this, one thing that jumps into my mind: UISliders are interactive components and should make use of UIInteractiveWidget. Maybe it is possible to combine this.

arcade/gui/widgets/slider.py
```diff
@@ -91,13 +203,25 @@ def __init__(
         y: float = 0,
         width: float = 300,
         height: float = 20,
-        size_hint=None,
-        size_hint_min=None,
-        size_hint_max=None,
-        style: Union[Mapping[str, "UISlider.UIStyle"], None] = None,  # typing: ignore
+        size_hint: Optional[int]=None,
+        size_hint_min: Optional[int]=None,
+        size_hint_max: Optional[int]=None,
```
eruvanos: these are optional tuples, size_hint is float. the others are int.

gran4:
> I will have a deeper look into this, one thing that jumps into my mind: UISliders are interactive components and should make use of UIInteractiveWidget. Maybe it is possible to combine this.

@eruvanos , that does not seem helpful in this case.

eruvanos:
Hi, I had a deeper look into the changes now (took me way to long).
I validated the idea of extracting a `_SliderParent` but the current implementation does not make use of it. As it is marked as an internal class, it does also not provide benefits for other game devs at the moment.

But I will check, if I can change it a little, so UITextureSlider and UI(Styled)Slider can have a abstract base class.

Thank you for your input and patience. I will check what I take over to finish the UISliders for 3.0.

eruvanos:
The slider changes are merged. Thanks again for your input!

PR #1682, opened 2023-04-05
Improve NinePatchTexture doc

pushfoo:
tl;dr
1. Clearer phrasing for all `NinePatchTexture` docstrings
2. Better behavior diagram, now with arrows & parameter labels!
3. Stricter flake8 & docstring PEP compliance

While responding to #1675, I looked noticed `NinePatchTexture` had unclear docstrings. This PR addresses this and serves as a stepping stone to further improvements beyond those already in #1680.

Doc built and tested locally.

PR #1769, opened 2023-05-11
Fix Surface.draw_texture (#1705)

### LINKED ISSUE #1705: Bug: Fix Surface.draw_texture's NinePatchTexture handling
### Actual behavior:
`arcade.gui.Surface.draw_texture` has following incorrect behavior:

1. The check for `angle` uses `x` and `y` arguments
2. The check for `alpha` uses `x` and `y` arguments

https://github.com/pythonarcade/arcade/blob/e76f34554f04d7e6e8e0e88f8a279469f6d4e475/arcade/gui/surface.py#L100-L110

### Expected behavior:

1. The angle check uses the `angle` argument
2. The alpha check uses the `alpha` argument

pushfoo:
Closes #1705 

### Changes

* `arcade.gui.Surface.draw_texture` now raises `NotImplementedError` when alpha != 255 or angle != 0.
* Added test

### How to test

Run test suite as normal

### Retained for context 

~~I have the logic corrected, but there are two questions I need answered:~~

1. ~~Is support for position implemented? [One comment says it is](https://github.com/pythonarcade/arcade/pull/1682#discussion_r1158379744), while [another gives me the impression it isn't](https://github.com/pythonarcade/arcade/issues/1705#issuecomment-1514367795).~~  Updated per comments
2. ~~Is using the `window` fixture acceptable in `tests/unit`? If not, should I build out a new `mock_window` fixture to fully or partially mock the gl context + window?~~ Implemented. I'd prefer to decouple this better, but it won't get us closer to 3.0.

### DIFF
--- a/arcade/gui/surface.py
+++ b/arcade/gui/surface.py
@@ -98,9 +98,6 @@ class Surface:
         alpha: int = 255,
     ):
         if isinstance(tex, NinePatchTexture):
-            if x != 0 or y != 0:
-                raise NotImplementedError("Ninepatch does not support a position != (0,0) yet")
-
             if angle != 0.0:
                 raise NotImplementedError("Ninepatch does not support a angle != 0 yet")
 
--- a/arcade/gui/surface.py
+++ b/arcade/gui/surface.py
@@ -99,10 +99,10 @@ class Surface:
     ):
         if isinstance(tex, NinePatchTexture):
             if angle != 0.0:
-                raise NotImplementedError("Ninepatch does not support a angle != 0 yet")
+                raise NotImplementedError(f"Ninepatch does not support an angle != 0 yet, but got {angle}")
 
             if alpha != 255:
-                raise NotImplementedError("Ninepatch does not support a alpha != 255 yet")
+                raise NotImplementedError(f"Ninepatch does not support an alpha != 255 yet, but got {alpha}")
 
             tex.draw_sized(size=(width, height))
         else:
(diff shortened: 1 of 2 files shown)


eruvanos:
@pushfoo 
hi,

**to 1.** 

Please remove the exception, it is implemented and we should support it.

TL;DR

I see the confusion :D
NinePatchTextures do support positioning. So in a general sense, we would be able to use it to render the NinePatchTexture in some specific place.

The big drawback is, that in the normal flow, a widgets sets the viewport and scissor box to its rect.
In combination, any other value then 0,0 for position would mostly move it out of the scissor box.

So in general speak, the Exception is wrong, NinePatch does support a position, but we had the exception, because we wanted to prevent devs to use it in the wrong way.

So I change my opinion:
we should support this feature of NinePatchTexture and remove the overprotective restriction.

arcade/gui/surface.py
```diff
@@ -94,18 +94,18 @@ def draw_texture(
...
         height: float,
         tex: Union[Texture, NinePatchTexture],
-        angle=0,
+        angle: float = 0.0,
         alpha: int = 255,
     ):
         if isinstance(tex, NinePatchTexture):
             if x != 0 or y != 0:
-                raise ValueError("Ninepatch does not support a position != (0,0) yet")
+                raise NotImplementedError("Ninepatch does not support a position != (0,0) yet")
```
eruvanos: Can be removed totally.

eruvanos (changes requested):

eruvanos:
Regarding 2.
Yes, window fixture is used in many places in unit test. You are good to go

eruvanos (approved):
Thanks

PR #1772, opened 2023-05-15
WIP: Implement ColorFloat

pushfoo:
Core questions / tasks:

- [x] Implement core class
- [x] Resolve circular dependency questions for conversion methods
- [ ] Add tests

Follow-up work:

1. Resolve naming issues (`ColorFloat` vs `ColorNorm`, `RGBAFloat` vs `RGBANorm`)

### DIFF
--- a/arcade/types.py
+++ b/arcade/types.py
@@ -176,7 +176,7 @@ class Color(RGBA255):
         return cls(r, g, b, a=a)
 
     @property
-    def normalized(self) -> RGBANormalized:
+    def normalized(self) -> "ColorFloat":
         """
         Return this color as a tuple of 4 normalized floats.
 
@@ -192,7 +192,7 @@ class Color(RGBA255):
             (0.0, 0.0, 0.0, 0.0)
 
         """
-        return self[0] / 255, self[1] / 255, self[2] / 255, self[3] / 255
+        return ColorFloat(r=self[0] / 255, g=self[1] / 255, b=self[2] / 255, a=self[3] / 255)
 
     @classmethod
     def from_gray(cls, brightness: int, a: int = 255) -> Self:
--- a/arcade/types.py
+++ b/arcade/types.py
@@ -623,7 +623,7 @@ class ColorFloat(RGBANormalized):
         Examples::
 
             >>> ColorFloat.from_color255((255, 0, 0, 255))
-            Color(r=1.0, g=0, b=0, a=1.0)
+            Color(r=1.0, g=0.0, b=0.0, a=1.0)
 
             >>> normalized_half_opacity_green = (0, 255, 0, 127)
             >>> ColorFloat.from_color255(normalized_half_opacity_green)
--- a/arcade/types.py
+++ b/arcade/types.py
@@ -561,10 +561,10 @@ class ColorFloat(RGBANormalized):
         Example::
 
             >>> ColorFloat.from_uint24(16777215)
-            Color(r=1.0, g=1.0, b=1.0, a=1.0)
+            ColorFloat(r=1.0, g=1.0, b=1.0, a=1.0)
 
             >>> ColorFloat.from_uint24(0xFF0000)
-            Color(r=255, g=0, b=0, a=255)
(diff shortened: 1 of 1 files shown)


arcade/types.py
```diff
@@ -410,6 +412,324 @@ def random(
         return cls(r, g, b, a)
 
 
+class ColorFloat(RGBANormalized):
```
cspotcode, line 418: Is this a good candidate for `NamedTuple`?  I think it gives us the `@property`s, the deepcopy, and the default `a` value for free.

I used this snippet to test, I may have missed corner cases.

```python
class Color(NamedTuple):
    r: float
    g: float
    b: float
    a: float = 1.0

a = Color(1, 2, 3)
b = copy.deepcopy(a)
print(repr(a), repr(b))
print(a is b)
```

cspotcode:
Should there be a class like this for every form of color, with conversion methods between them all?  Or should this be the only color class, and all conversions between color formats are done into or out of this class?  E.g.

```python
my_color_rgba255 = ColorFloat.from_hex_string("FFEE00").as_rgba255
```
EDIT Actually, is that conversion safe?  I guess floats have enough precision to reliably convert `"EE"` -> `float` -> `238`.  It *feels* dirty but is probably fine.

arcade/types.py
pushfoo: tl;dr [`NamedTuple` subclassing made validation harder for `Color`](https://github.com/pythonarcade/arcade/pull/1639#discussion_r1140971447), and [is much slower according to einarf](https://github.com/pythonarcade/arcade/pull/1639#discussion_r1141326043). 

I don't know if the latter is still true on more recent Python versions, and it may be worth rebenchmarking periodically. I agree that your proposal could be cleaner if future minimum Python versions lose its current disadvantages.

pushfoo:
tl;dr to my knowledge, the current approach is compatible and performant at the price of verbosity

> Should there be a class like this for every form of color, with conversion methods between them all?

There are only two color formats I know of supported in arcade and pyglet:
1. Byte-value RGBA tuples on the user-facing side
2. Float-valued RGBA tuples on the internal OpenGL side, [as inside `SpriteList`](https://github.com/pythonarcade/arcade/blob/1c2975fcfeb0e82eac1611245dccfa20a0e7e207/arcade/sprite_list/sprite_list.py#L109)

The implementation in this PR will cover both of them with backward-compatible fancy versions. It will also help with a potential switch to storing colors as floats internally across the codebase. It's something which einarf has brought up at least once before, and I think it's worth pursuing.

> Or should this be the only color class, and all conversions between color formats are done into or out of this class? E.g.

I'm not sure which other color formats we'd want to add. HSV or HSL seem like the best candidates, but their primary use for games seems like it would be color shifting or generating palettes rather than real-time gameplay. Other color models don't seem as useful since this is a games library rather than image editor or color framework.

Building a fully generic, extensible `Color` type doesn't seem worth the disadvantages it seems to have:
1. Loss of easy compatibility with pyglet and older arcade-dependent code
2. Possible performance hits from more complicated conversion logic

If there's a way to do it without these problems, then it's worth considering.

> It feels dirty but is probably fine.

It's worth double checking. The current implementation is a conversion function from 2.6.X moved into an instance method and with nicer formatting.

arcade/types.py
cspotcode: Yeah, I see einarf saying NamedTuple is slow, but not saying that what you have written is faster.  That makes me think that both might be equally slow.  Did anyone benchmark at the time of that conversation?

There's good precedent for this question, since one of my first benchmarks found that shapely was slower even though the shapely code was meant to be faster.

pushfoo:
> Should there be a class like this for every form of color, with conversion methods between them all? Or should this be the only color class, and all conversions between color formats are done into or out of this class? E.g.

After giving the issue additional thought, I think the following approach is best:
1. Finish this PR
2. If needed, optimize or replace the backing implementation

We're getting bogged down in details. As much as I love the idea of HSV support, I don't think it's crucial for 3.0.

pushfoo:
Closing for the time being but keeping the branch.

PR #1777, opened 2023-05-16
Add basic type annotations & docstrings to arcade.hitbox.base

pushfoo:
### Changes

* Add missing annotations to `arcade.hitbox.base` members
* Add basic docstrings to `arcade.hitbox.base` members
* Add protected top-level workaround constant
* Add notes on why we don't cache axis extrema properties

### Why

* IDE tooltips & hints
* Help anyone who reads the source

### How to test

1. Run linting & unit tests as normal
2. Check IDE type annotation / docstring rendering for HitBox classes

### Questions for reviewers

* Is our rotation  direction consistent enough that I can add it here / to the programming guide?
* Should I update `update_quick_index.py` to generate this doc now?

### DIFF
--- a/arcade/hitbox/base.py
+++ b/arcade/hitbox/base.py
@@ -37,7 +37,8 @@ class HitBoxAlgorithm:
         """
         A string representation of the parameters used to create this algorithm.
 
-        This is used when caching :py:class:`~arcade.Texture` instances.
+        Subclasses should override this method to return a meaningful value since this
+        method is called when caching :py:class:`~arcade.Texture` instances.
         """
         return ""
 
--- a/arcade/hitbox/base.py
+++ b/arcade/hitbox/base.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from math import cos, radians, sin
-from typing import Any, Tuple
+from typing import Any, Tuple, Optional
 
 from PIL.Image import Image
 
@@ -9,6 +9,12 @@ from arcade.types import Point, PointList
 
 
 __all__ = ["HitBoxAlgorithm", "HitBox", "RotatableHitBox"]
+
+
+# Speed / typing workaround:
+# 1. Makes sure we don't reallocate new empty tuples
+# 2. Allows HitBox & subclass typing annotation to work cleanly
+_EMPTY_TUPLE = tuple()
 
 
 class HitBoxAlgorithm:
@@ -63,6 +69,19 @@ class HitBoxAlgorithm:
 
 
 class HitBox:
+    """
+    A basic hit box class supporting scaling.
+
+    It includes support for rescaling as well as shorthand properties
+    for boundary values along the X and Y axes. For rotation support,
+    use :py:meth:`.create_rotatable` to create an instance of
+    :py:class:`RotatableHitBox`.
+
+    :param points: The unmodified points bounding the hit box
+    :param position: The center around which the points will be offset
+    :param scale: The X and Y scaling factors to use when offsetting the
+        points
+    """
     def __init__(
         self,
         points: PointList,
@@ -78,15 +97,27 @@ class HitBox:
         self._bottom = None
         self._top = None
 
-        self._adjusted_points = None
+        # This empty tuple will be replaced the first time
+        # get_adjusted_points is called
+        self._adjusted_points: PointList = _EMPTY_TUPLE
         self._adjusted_cache_dirty = True
 
     @property
-    def points(self):
+    def points(self) -> PointList:
+        """
+        The raw, unadjusted points of this hit box.
+
+        These are the points as originally passed before offsetting, scaling,
+        and any operations subclasses may perform, such as rotation.
+        """
         return self._points
 
     @property
-    def position(self):
+    def position(self) -> Point:
+        """
+        The center point used to offset the final adjusted positions.
+        :return:
+        """
         return self._position
 
     @position.setter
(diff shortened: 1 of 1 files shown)


arcade/hitbox/base.py
```diff
@@ -11,34 +11,78 @@
...
+    Hit box algorithms are intended to calculate the points which make up
+    a hit box for a given :py:class:`~PIL.Image.Image`. However, advanced
+    users can also repurpose them for other tasks.
     """
 
     #: The name of the algorithm
     name = "base"
-    #: Should points for this algorithm be cached?
+
+    #: Whether points for this algorithm should be cached
```
cspotcode, line 32: In general, do we want these to be docstrings?  They would appear in tooltips and Sphinx docs.

arcade/hitbox/base.py
```diff
@@ -11,34 +11,78 @@
...
 
     @property
     def param_str(self) -> str:
         """
-        Return a string representation of the parameters used to create this algorithm.
+        A string representation of the parameters used to create this algorithm.
 
-        This is used in caching.
+        Subclasses should override this method to return a meaningful value since this
+        method is called when caching :py:class:`~arcade.Texture` instances.
```
cspotcode: This is a good description, but should we be more explicit that this string *is* used as part of a larger cache key?  "this method is called when caching" is a great description but stops just shy of saying that the value is actually a part of the cache key.

arcade/hitbox/base.py
```diff
@@ -49,20 +93,36 @@ def __init__(
...
         self._scale = scale
 
+        # Per Clepto's testing as of around May 2023, these are better
+        # left uncached because caching them is somehow slower than what
+        # we currently do. Any readers should feel free to retest /
+        # investigate further.
         self._left = None
         self._right = None
         self._bottom = None
         self._top = None
```
cspotcode: Are these attributes ever set to non-`None`?  Should we remove them entirely?

arcade/hitbox/base.py
```diff
@@ -107,11 +184,30 @@ def create_rotatable(
         self,
         angle: float = 0.0,
     ) -> RotatableHitBox:
+        """
+        Create a rotatable instance of this hit box.
+
+        The internal point list is transferred directly instead of
+        deepcopied, so care should be taken if using a mutable internal
+        representation.
```
cspotcode, line 190: I'm being super-nitpicky, but "internal point list" is slightly incorrect if it's a tuple, not a list.  Maybe something like "The internal point sequence is referenced directly instead of copied"?  I dunno.

arcade/hitbox/base.py
```diff
@@ -138,18 +240,28 @@ def __init__(
...
 
-    def get_adjusted_points(self):
+    def get_adjusted_points(self) -> PointList:
+        """
+        Return the offset, scaled, & rotated points of this hitbox.
+
+        As with :py:meth:`.HitBox.get_adjusted_points`, this method only
+        recalculates the adjusted values when necessary.
+        :return:
+        """
```
cspotcode, line 262: Do we want to avoid the duplicated docstring here, instead writing a description on `HitBox` that can apply generically to all subclasses?

arcade/hitbox/base.py
pushfoo: Possibly, but they won't render at the moment. To my understanding, we'd need to enable API doc generation for this file by altering `update_quick_index.py` and running it. I mentioned it in the PR, but I realize I wasn't very clear:

> Should I update update_quick_index.py to generate this doc now?

Sorry about that. :sweat_smile:

arcade/hitbox/base.py
pushfoo: I don't like it either, but I don't want to rename the `PointList` type in this PR. I'm trying to be very disciplined about PR scope to build good habits & build up a good set of examples for the beginners.

Although the change could go in a prequel PR this one merges or rebases onto, I think it should be in a follow-up one. I don't want to start changing things einarf and Clepto understand the full implications of far better than I do without consulting them first.

arcade/hitbox/base.py
pushfoo: I think it may be worth it to be explicit about the rotation for now. To my understanding, doing otherwise requires:

1. Getting the inheritance of docstrings and methods working first, which may be hard
2. Trusting the user to understand the idea of transformations on points

Please correct me if I'm wrong, because I'd love to be wrong about item 1.

The second concern is less important than I initially thought, as I found out these classes were intended by einarf and Clepto to be for advanced users only. This may be why they aren't included in the API doc at the moment. I'd be tempted to agree if we had a better sprite / texture generation API set up beyond `SpriteCircle` and `SpriteSolidColor`, but it's not a core 3.0 goal so I want to leave some notes in place for both ambitious beginners and my future 4am weekend self.

arcade/hitbox/base.py
pushfoo: To my understanding, they aren't. Which approach seems better?
1. Comment them out as well
2. Delete them and move the note to just above the `left` property

arcade/hitbox/base.py
cspotcode: Editor tooltips still seem useful for people working on arcade, even if this stuff doesn't render into Sphinx.  But no strong preference, just seems like easier to standardize on using docstrings everywhere.

arcade/hitbox/base.py
cspotcode: I think deleting them is good because it reduces memory usage and increases speed of hitbox creation.

arcade/hitbox/base.py
cspotcode: Oh I see, if it said `internal PointList is` then I would have understood what you were going for.  Seeing it as `point list` threw me off.

arcade/hitbox/base.py
cspotcode: I'm not sure.  "Tranformation" is definitely the correct term for what we're doing, where different `HitBox` implementations support different transformations.  I have no strong preference here; don't wanna block a merge.

arcade/hitbox/base.py
pushfoo: I may be wrong on the current comments not rendering. The `#:` syntax, it seems to be referred to [as a "doc comment" in some places](https://stackoverflow.com/questions/8649105/how-to-show-instance-attributes-in-sphinx-doc), and it seems to have multiple meanings according to the autodoc extension's documentation. We have autodoc enabled in our conf.py.

The tooltip issue is still important, as this format of documentation doesn't get picked up by IDEs.

arcade/hitbox/base.py
pushfoo: I'll leave this as is for now & mark it as resolved.

arcade/hitbox/base.py
pushfoo: Marking as resolved per discord discussion. These class attributes may be removed in future PRs, so there's no point in perfecting their doc.

Issue #1779, opened 2023-05-16
VideoPlayerView is not accessable


MrWardKKHS:
## Enhancement request:
Allow VideoPlayerView to be imported

### What should be added/changed?
arcade\experimental\__init__.py should have VideoPlayerView added to exports

### What would it help with?
Quality of life \ ease of use improvement for users trying to play videos \ cutscenes

einarf:
There's some more work needed to make it a bit more flexible, but I don't see any harm in this. The cv2 player could also potentially be imported wrapped in an ImportError.

Ideally we should complete it and move it into `arcade.video`. It's something we can wrap up later in 3.x.x.

MrWardKKHS:
This was fixed by #1780

PR #1780, opened 2023-05-16
#1779 Include VideoPlayerView in exports 

MrWardKKHS:
Will add the video player view for users. Especially helpful in quick prototyping and reducing boilerplate.

### DIFF
--- a/arcade/experimental/__init__.py
+++ b/arcade/experimental/__init__.py
@@ -2,6 +2,7 @@
 Experimental stuff. API may change.
 """
 from .video_player import VideoPlayer
+from .video_player import VideoPlayerView
 from .texture_render_target import RenderTargetTexture
 from .shadertoy import Shadertoy, ShadertoyBuffer, ShadertoyBase
 from .crt_filter import CRTFilter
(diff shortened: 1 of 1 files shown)


pushfoo (approved):
I think this looks good as-is. :+1:

I didn't even know we had this feature, and I appreciate you bringing it to my attention!

If you have time for a follow-up PR, I'd appreciate it if you could take a look at the following:

1. Add a small public domain MP4 we can include instead of the hard link to a local file on the original author's device
2. Type annotations
3. Better docstrings & formatting

arcade/experimental/__init__.py
```diff
@@ -2,6 +2,7 @@
 Experimental stuff. API may change.
 """
 from .video_player import VideoPlayer
+from .video_player import VideoPlayerView
```
gran4, line 5:
```suggestion
from .video_player import VideoPlayer, VideoPlayerView
```

gran4 (changes requested):

gran4:
@pvcraven it has a mistake.
`from .video_player import VideoPlayer`
`from .video_player import VideoPlayerView`

should be made
`from .video_player import VideoPlayer, VideoPlayerView`

PR #1786, opened 2023-05-18
Improvements to video players

MrWardKKHS:
Various improvements to the two Video Players Views in the experimental module. The ultimate goal here would be to have an easy to use class that can load a video and play it, esp as a cutscene. This PR gets closer even though it still doesn't feel solved. 

- The opencv video player and video player view is exposed to users now, 
- The ability to loop the cv2 player added
- The default behavior of the cv2 player to resize the window turned off
- Video added from pexels https://www.pexels.com/license/ 
- Typing improvements

Future improvements might be to merge VideoPlayerView and CV2PlayerView as this seems like duplication


arcade/experimental/__init__.py
```diff
@@ -1,8 +1,8 @@
 """
 Experimental stuff. API may change.
 """
-from .video_player import VideoPlayer
-from .video_player import VideoPlayerView
+from .video_player import VideoPlayer, VideoPlayerView
+from .video_cv2 import VideoPlayerCV2, CV2PlayerView
```
einarf: Wrap the cv2 imports in a try/except ImportError. The module should only be available if you have cv2 installed.

einarf (changes requested):

MrWardKKHS:
I'm not sure if what I've done is kosher. I was struggling with not putting out an error message for everyone every time they import arcade if they don't have cv2, but also having an error message when the user tries to use the class without the right dependencies. It does mean that what should be the CV2PlayerView class becomes a function callback, which feels like a hack.. but it works. 

... And the test have failed, so I guess it's too hacky. Advice?

arcade/experimental/__init__.py
MrWardKKHS: I'm not sure if what I've done is kosher. I was struggling with not putting out an error message for everyone every time they import arcade if they don't have cv2, but also having an error message when the user tries to use the class without the right dependencies. It does mean that what should be the CV2PlayerView class becomes a function callback, which feels like a hack.. but it works.

... And the tests have failed, so I guess it's too hacky. Advice? Just do `except: pass` and have it fail silently?

einarf: Just `pass` for now is fine. It's still something in experimental. It can be improved of we move it down into main namespace.

MrWardKKHS: Fixed in [8df8bd6](https://github.com/pythonarcade/arcade/pull/1786/commits/8df8bd65491a163fde791da3e4d4006c8e9c0d55)

einarf (changes requested):
If we're adding mp4 file to arcade we should also add the extension here so it's included in the wheel / source distribution.
https://github.com/pythonarcade/arcade/blob/development/MANIFEST.in

arcade/resources/assets/video/snake.mp4
pushfoo, line 1: I think it would be a good idea to add a citation for this. A `.txt` file with the following in the same directory would probably do:
1. The license it's used under
2. The URL this was taken from
3. Anything else you want to add

arcade/experimental/video_cv2.py
```diff
@@ -110,27 +114,37 @@ def update(self, delta_time):
...
+            elif self.loop:
+                self.time = 0.0
+                self.video.set(cv2.CAP_PROP_POS_FRAMES, 0)
+
 
 
 class CV2PlayerView(arcade.View):
-    def __init__(self, path: str):
+    """A simple view to hold a video player using cv2.
+    requires opencv-python
```
pushfoo: The formatting is a little off on this + you could be clearer about what opencv-python is.
```suggestion
    """
    A simple view to hold a video player using cv2.

    Requires the opencv-python module to be installed.
    """
```

arcade/experimental/__init__.py
```diff
@@ -1,17 +1,23 @@
...
+try:
+    from .video_cv2 import VideoPlayerCV2, CV2PlayerView
+except ImportError:
+    pass
 
 __all__ = [
     "VideoPlayer",
     "VideoPlayerView",
+    "VideoPlayerCV2",
+    "CV2PlayerView",
```
pushfoo: @cspotcode Should inclusion of these classes in `__all__` be conditional as well?

pushfoo (commented):
I think this is good overall, but there some unresolved questions as well as a few changes which I think would be a good idea.

arcade/experimental/__init__.py
MrWardKKHS: Updated this in [03fa9dc](https://github.com/pythonarcade/arcade/pull/1786/commits/03fa9dca617fe2bb1ab4ccf86b6ceb0679e79f9b). This does cause the linting test to fail as an unused import though... added an ignore here.

arcade/resources/assets/video/snake.mp4
MrWardKKHS: Resolved in [18819c2](https://github.com/pythonarcade/arcade/pull/1786/commits/18819c2f0459eb5fa2fc7d98e36e641f7bbd17ff) snake.mp4 has been removed and replaced with earth.mp4, a public domain video from NASA. Credit given in readme.txt

arcade/experimental/__init__.py
cspotcode: @pushfoo 

Why don't we put the conditional around `import cv2` within `video_vc2.py`, and remove the conditional from `__init__.py`?  Easier to reason about, and it conveniently sidesteps this `__all__` issue.

As for `__all__`, 2 factors to consider:

a) static tooling should understand the types of exported items
b) python should behave at runtime in these situations:
```python
import arcade.experimental
from arcade.experimental import * # <-- this requires any __all__ items we can't import to be declared in another way, perhaps `MissingItem = None`
```

cspotcode: @pushfoo since einarf and MrWardKKHS already talked about where to put the conditional, keeping it here as-is makes sense.  It prevents a runtime error if anyone does `from arcade.experimental import *`.  Any static tooling issues can be fixed in the future, and we can't detect them without #1791.

arcade/experimental/__init__.py
```diff
@@ -19,3 +17,15 @@
...
     "BloomFilter",
 ]
+
+# Keep cv2 an optional dependency
+try:
+    from .video_cv2 import VideoPlayerCV2, CV2PlayerView # noqa: F401
+
+    __all__.extend([
+        "VideoPlayerCV2",
+        "CV2PlayerView",
```
cspotcode, line 27: TIL that typecheckers apparently support `__all__.extend(`

https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md

pushfoo (approved):
I'm fine with this if einarf & CI are. The `MANIFEST.in` addition is very good attention to detail. :+1:

einarf (approved):

PR #1788, opened 2023-05-18
Controversial pyright fixes

cspotcode:
These are the pyright fixes that might involve a bit more discussion and refinement before merging.

### DIFF
--- a/arcade/gl/framebuffer.py
+++ b/arcade/gl/framebuffer.py
@@ -1,10 +1,12 @@
 from ctypes import c_int, string_at
 from contextlib import contextmanager
-from typing import Optional, Tuple, List, TYPE_CHECKING
+from typing import Optional, Tuple, List, TYPE_CHECKING, Union
 import weakref
 
 
 from pyglet import gl
+
+from arcade.types import RGBOrA255, RGBOrANormalized
 
 from .texture import Texture2D
 from .types import pixel_formats
@@ -344,7 +346,7 @@ class Framebuffer:
 
     def clear(
         self,
-        color=(0.0, 0.0, 0.0, 0.0),
+        color: Union[RGBOrA255, RGBOrANormalized] =(0.0, 0.0, 0.0, 0.0),
         *,
         depth: float = 1.0,
         normalized: bool = False,
--- a/arcade/gl/framebuffer.py
+++ b/arcade/gl/framebuffer.py
@@ -388,8 +388,10 @@ class Framebuffer:
                 if len(color) == 3:
                     gl.glClearColor(color[0] / 255, color[1] / 255, color[2] / 255, 1.0)
                 else:
+                    # mypy does not understand that color[3] is guaranteed to work in this codepath, pyright does.
+                    # We can remove this type: ignore if we switch to pyright.
                     gl.glClearColor(
-                        color[0] / 255, color[1] / 255, color[2] / 255, color[3] / 255
+                        color[0] / 255, color[1] / 255, color[2] / 255, color[3] / 255 # type: ignore
                     )
 
             if self.depth_attachment:
--- a/arcade/gl/glsl.py
+++ b/arcade/gl/glsl.py
@@ -1,8 +1,9 @@
-from typing import Dict, Iterable, List, Optional
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional
 import re
 
 from pyglet import gl
-from .context import Context as ArcadeGlContext
+if TYPE_CHECKING:
+    from .context import Context as ArcadeGlContext
 
 from .exceptions import ShaderException
 from .types import SHADER_TYPE_NAMES, PyGLenum
@@ -29,7 +30,7 @@ class ShaderSource:
     """
     def __init__(
         self,
-        ctx: ArcadeGlContext,
+        ctx: 'ArcadeGlContext',
         source: str,
         common: Optional[Iterable[str]],
         source_type: PyGLenum,
--- a/arcade/gl/glsl.py
+++ b/arcade/gl/glsl.py
@@ -2,6 +2,7 @@ from typing import Dict, Iterable, List, Optional
 import re
 
 from pyglet import gl
+from .context import Context as ArcadeGlContext
 
 from .exceptions import ShaderException
 from .types import SHADER_TYPE_NAMES, PyGLenum
@@ -28,7 +29,7 @@ class ShaderSource:
     """
     def __init__(
         self,
-        ctx: gl.Context,
+        ctx: ArcadeGlContext,
         source: str,
         common: Optional[Iterable[str]],
         source_type: PyGLenum,
--- a/arcade/gui/surface.py
+++ b/arcade/gui/surface.py
@@ -166,7 +166,6 @@ class Surface:
         The surface will be rendered at the configured ``position``
         and limited by the given ``area``. The area can be out of bounds.
 
-        :param Optional[Point] position: The position to draw the surface at.
         :param Optional[Rect] area: Limit the area in the surface we're drawing (x, y, w, h)
         """
         # Set blend function
--- a/arcade/gui/surface.py
+++ b/arcade/gui/surface.py
@@ -6,7 +6,7 @@ from arcade import Texture
 from arcade.color import TRANSPARENT_BLACK
 from arcade.gl import Framebuffer
 from arcade.gui.nine_patch import NinePatchTexture
-from arcade.types import FloatRect, Point, RGBOrA255
+from arcade.types import RGBA255, FloatRect, Point
 
 
 class Surface:
@@ -83,7 +83,7 @@ class Surface:
     def height(self) -> int:
         return self._size[1]
 
-    def clear(self, color: RGBOrA255 = TRANSPARENT_BLACK):
+    def clear(self, color: RGBA255 = TRANSPARENT_BLACK):
         """Clear the surface"""
         self.fbo.clear(color=color)
 
--- a/arcade/gui/surface.py
+++ b/arcade/gui/surface.py
@@ -6,7 +6,7 @@ from arcade import Texture
 from arcade.color import TRANSPARENT_BLACK
 from arcade.gl import Framebuffer
 from arcade.gui.nine_patch import NinePatchTexture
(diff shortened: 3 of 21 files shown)


arcade/gl/glsl.py
```diff
@@ -1,7 +1,9 @@
-from typing import Dict, Iterable, List, Optional
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional
 import re
 
 from pyglet import gl
+if TYPE_CHECKING:
+    from .context import Context as ArcadeGlContext
```
pvcraven, line 6: I would have expected this code to upset the linter, as imports aren't supposed to come after other code? I'm ok since it appears to work.

arcade/gui/widgets/layout.py
```diff
@@ -590,11 +586,6 @@ def min_size(child: UIWidget) -> Tuple[float, float]:
 
             max_height_per_row[row_num][col_num] = (shmn_h, row_span)
 
-            for row in child_sorted_row_wise[
-                row_num : row_num + row_span  # noqa: E203
-            ]:
-                row[col_num : col_num + col_span] = [child] * col_span  # noqa: E203
-
```
pvcraven, line 597: What was this doing, and did removing it cause any ill effects?

arcade/sound.py
```diff
@@ -108,13 +108,14 @@ def stop(self, player: media.Player) -> None:
         if player in media.Source._players:
             media.Source._players.remove(player)
 
-    def get_length(self) -> float:
+    def get_length(self) -> Optional[float]:
```
pvcraven: If this is optional, we should probably update the docs to note when we'd get None back.

arcade/sound.py
```diff
@@ -108,13 +108,14 @@ def stop(self, player: media.Player) -> None:
...
         """Get length of audio in seconds"""
         return self.source.duration
 
     def is_complete(self, player: media.Player) -> bool:
         """Return true if the sound is done playing."""
-        return player.time >= self.source.duration
+        # known limitiation: streaming sources do not set duration, will throw an error
```
pvcraven: limitation

arcade/text.py
```diff
@@ -386,7 +386,7 @@ def color(self, color: RGBA255):
         self._label.color = Color.from_iterable(color)
 
     @property
-    def width(self) -> int:
+    def width(self) -> Optional[int]:
```
pvcraven, line 389: If it is optional, when would we NOT return a width?

arcade/gl/glsl.py
cspotcode: Perhaps `if TYPE_CHECKING:` is special-cased, since the Python ecosystem recommends this pattern.

arcade/gui/widgets/layout.py
cspotcode: I'm guessing it's dead code copy-pasted from `def add`.  The same thing appears in `def add`, where it's actually used.

arcade/sound.py
cspotcode: Should we also update the docs for `is_complete` to say that it will throw an exception if you call it on a streaming source?

arcade/text.py
cspotcode: If you instantiate as `arcade.text.Text(width=None)` then width will be `None`.

https://github.com/pythonarcade/arcade/blob/d7c6986b0241f0304f85a00493e9cd3c51ce184c/arcade/text.py#L185
https://github.com/pythonarcade/arcade/blob/d7c6986b0241f0304f85a00493e9cd3c51ce184c/arcade/text.py#L212

arcade/sound.py
cspotcode: I based this on a comment in pyglet's code:

https://github.com/pyglet/pyglet/blob/515d7a52a63fc20d76109f06a02f059459ec03b8/pyglet/media/codecs/base.py#L207-L216

> Not all source durations can be determined; in this case the value is ``None``.

But I'm not sure when that happens, and I'm not sure how to test it.  If I load a sound and get a duration, it does not mean that *other* OSes or codecs won't return `None`.

cspotcode: After some Discord discussion, it sounds easier to assert that duration is always available, and throw an exception if not.

arcade/gui/surface.py
```diff
@@ -83,7 +83,7 @@ def width(self) -> int:
     def height(self) -> int:
         return self._size[1]
 
-    def clear(self, color: RGBA255 = TRANSPARENT_BLACK):
+    def clear(self, color: RGBOrA255 = TRANSPARENT_BLACK):
```
pushfoo: This was left as `RGBA255` intentionally as a nudge toward full RGBA. I don't have the benchmarks handy, but I believe it may be slightly faster to use RGBA colors in some cases. The `RGBAOrA255` alias is only in the defined for completeness reasons in case of edge cases that demand it.

arcade/gui/widgets/__init__.py
```diff
@@ -759,7 +760,7 @@ def __init__(
             size_hint_min=size_hint_min,
             size_hint_max=size_hint_max,
         )
-        self.color = (randint(0, 255), randint(0, 255), randint(0, 255))
+        self.color: RGBOrA255 = (randint(0, 255), randint(0, 255), randint(0, 255))
```
pushfoo: To my understanding, this should be upgraded to be use full alpha and `Color.from_iterable` or similar.

arcade/sound.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, file_name: Union[str, Path], streaming: bool = False):
             self.file_name, streaming=streaming
         )
 
+        if self.source.duration is None:
+            raise Exception("Audio duration must be known when loaded, but this audio source returned `None`")
```
pushfoo: This is probably better as a `ValueError` or a more specific type. I'm not sure if pyglet provides media-related exceptions, but we could probably add a `ValueError` subclass to `arcade.utils` or wherever the `ValueError` subclasses like `ByteRangeError` are right now.

pushfoo (changes requested):
tl;dr I don't have an opinion on the `#type: ignore` or other elements but do have some concerns about types used in a few places

1. The previous standard was to use `RGBA255` despite technically accepting RGB colors
2. We should probably use a more specific exception type when the media player gives nonsense values

arcade/gui/widgets/__init__.py
cspotcode: How about we extract that to an issue and tackle in a different PR?  That way, we can keep this PR focused solely on enabling pyright while preserving existing behavior.

arcade/gui/surface.py
cspotcode: How about we extract that to an issue and tackle in a different PR? That way, we can keep this PR focused solely on enabling pyright while preserving existing behavior.

arcade/gui/widgets/__init__.py
pushfoo: Sounds good. :+1:

arcade/sound.py
cspotcode: What should the exception be called?  The docs say `ValueError` means the wrong kind of value was passed, such as a number out of range.  Is that what triggers a source to have no duration? I don't really know.

https://docs.python.org/3/library/exceptions.html#ValueError

arcade/gui/surface.py
cspotcode: Nevermind me, the change to use `RGBA255` was tiny, I made the change.

arcade/gui/widgets/__init__.py
cspotcode: Nevermind me, the change to use `RGBA255` was tiny, I made the change.

arcade/sound.py
pushfoo: > Is that what triggers a source to have no duration? I don't really know.

At least two backends supported by pyglet can fail to know the duration of an audio / video source:

* https://superuser.com/questions/1179000/ffmpeg-get-duration-of-video-file-without-meta-data
* https://github.com/beetbox/audioread/issues/31

There are also might be unlikely edge cases like trying to read from a live streaming audio / video source for which the length can't be known beforehand.

Although `ValueError` seems the most appropriate base exception type since we're receiving a value which doesn't match the expected requirements, there are also [some exception types in pyglet which may be relevant](https://github.com/pyglet/pyglet/blob/ff89b4f68db321a2e1699a8e70ca7fcf06d52ae0/pyglet/media/exceptions.py):

* `MediaException(Exception)`
* `MediaFormatException(MediaException)`
* `CannotSeekException(MediaException)`

`MediaException` seems like the most appropriate of them, but I'm not sure either.

arcade/gl/framebuffer.py
```diff
@@ -344,7 +346,7 @@ def _use(self, *, force: bool = False):
 
     def clear(
         self,
-        color=(0.0, 0.0, 0.0, 0.0),
+        color: Union[RGBOrA255, RGBOrANormalized] =(0.0, 0.0, 0.0, 0.0),
```
pushfoo, line 349: It may be worth using the RGBA-only types here as well.

arcade/sound.py
cspotcode: `MediaException` or `MediaFormatException`, yeah.  Note that these are *not* subclasses of `ValueError`, so anyone expecting to catch `ValueError` will not catch these exceptions.

https://github.com/pyglet/pyglet/blob/c6cb278ed2b2053c044a60e26a2a929e105f2fda/pyglet/media/drivers/pulse/interface.py#L286
https://github.com/pyglet/pyglet/blob/c6cb278ed2b2053c044a60e26a2a929e105f2fda/pyglet/media/codecs/ffmpeg.py#L168-L175

arcade/gl/framebuffer.py
cspotcode: Maybe, but that sounds like a bigger conversation, and I'm worried that changing it here will draw other reviewers out of the woodwork who will argue that we should keep the convenience of omitting the alpha value.

pushfoo (approved):
Nothing stands out as a problem to me aside from `clear`, which needed a rework anyway from what I remember from discussing it with einarf.

PR #1793, opened 2023-05-22
Refactor hitbox algo cache names into an instance property

pushfoo:
~~Stacks onto #1792~~ Rebased onto `development`, no longer depends on that PR.

### Changes

1. Remove `name` class attribute from hit box algorithm classes
2. Add `cache_name` per-instance property
3. Update tests

### Why

Per [discord discussion with einarf](https://discord.com/channels/458662222697070613/1108023300589879386/1108497287979737169), the current cache names are from the old caching system:

> I think we can drop the name and just use self.__class__.__name__, right?
> No need to make it complicated

### How to test

`./make.py test` or `pytest tests/unit` as usual

### Follow-up work

Refactor `cache` class attribute into a `cache_points` instance property.

### DIFF
--- a/arcade/hitbox/base.py
+++ b/arcade/hitbox/base.py
@@ -24,20 +24,10 @@ class HitBoxAlgorithm:
     cache = True
 
     def __init__(self):
-        self._texture_cache_base = self.__class__.__name__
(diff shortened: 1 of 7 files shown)


einarf:
Much better 👍

PR #1809, opened 2023-06-04
Fix pyright issues in pathing

pushfoo:
Follow-up to #1762 

### Changes

* Cast a return type to silence a pyright error
* Use `#type: ignore` on a short-circuit evaluation pyright fails to understand

### Follow-up work

The UI stack needs similar fixes, but I do not yet understand the intended design well enough to be confident fixing it.

### DIFF
--- a/arcade/paths.py
+++ b/arcade/paths.py
@@ -2,6 +2,7 @@
 Classic A-star algorithm for path finding.
 """
 from typing import (
+    cast,
     List,
     Optional,
     Set,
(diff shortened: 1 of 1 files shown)


arcade/paths.py
```diff
@@ -168,7 +169,7 @@ def _AStarSearch(start: Point, end: Point, graph: _AStarGraph) -> Optional[List[
         current = None
         current_fscore = None
         for pos in sorted(open_vertices):
-            if current is None or F[pos] < current_fscore:
+            if current is None or F[pos] < current_fscore:  # type: ignore
```
cspotcode, line 172: Another option: remove `current_fscore = None` so that the type is `float`

eruvanos: Will merge and replace None with `math.inf`

Issue #1796, opened 2023-05-24
Capturing typehinting guidance


cspotcode:
This issue serves as a living document describing how we use typehints.

*Note: at the time of writing, some of the following is forward-thinking.  It describes things that will be true in the near future.*

# Summary

We typecheck arcade's own codebase with pyright, not mypy.  However, we support consumers of arcade who might use mypy on their projects.

# Guidelines

## Consult pyright's guidance

There are a ton of goodies in this doc.

https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md

## Annotate with type hints, do *not* put type information into docstrings

The Sphinx documentation generator can be configured to parse typehints so that the type information does not need to be repeated within docstrings.

## Annotate `__init__` with `-> None`

https://stackoverflow.com/questions/46779046/correct-type-annotation-for-init

## Errors where something is typed `Optional` but can't be `None` in a given code path

Our pyright configuration suppresses a bunch of `Optional` errors, but not all of them.  If the typechecker complains that something might be `None`:

A) it really might be `None`

Do a runtime check and raise a descriptive exception.

```python
if thing_that_cannot_be_none is None:
    raise Exception("Description here")
```

B) it's guaranteed never to be `None`, but the typechecker doesn't understand

```python
if TYPE_CHECKING:
    assert thing_that_cannot_be_none
```

gran4:
Should there be a line after a class decleration(no docstrings) or not? I see both in the arcade code.

pushfoo:
> Should there be a line after a class decleration(no docstrings) or not? I see both in the arcade code.

For arcade, I've steered toward putting the summary line on a new line, like this:
```py
def func_name(value: int):
    """
    An example summary line on its own line.

    Additional function details.

    :param value: some value the function will use
    """
    print(value)
```

einarf:
Linked in `CONTRIBUTING.md`

PR #1810, opened 2023-06-04
Add  -> None to many __init__s

gran4:
Fixes some of https://github.com/pythonarcade/arcade/issues/1796.
Error bc some weird thing, not my PR's problem.


einarf:
Fixed some conflicts. I assume this one needs pyright to actually pass? mypy definitely doesn't like this unless I missed something important.

gran4:
??????????????? I only added __init__ to the files as the title says. The errors are not created by my PR

einarf:
Some more conflicts to solve here before merge.

einarf:
After a discussion: We don't really want these type annotations in examples and tutorials. That's 90% of the changes here. Adding None return types elsewhere is fine, but it's faster to redo this work than changing ths PR.

PR #1818, opened 2023-06-10
Camera Refactor

DragonMoffon:

The refactor includes:
Two ProjectionData PoDs that hold the data for making either an orthographic or projection matrix.

See the 
#Cameras
topics in the arcade-dev forum on the Python Arcade discord server for more info.

There are also likely issues that are relevant such as #1651.


DragonMoffon:
Adding to my TODO list.

Splitting the code into multiple files
Unit tests for every core piece of the code.
Examples for every camera type and a few examples of controllers.

arcade/cinematic/default.py
```diff
@@ -0,0 +1,52 @@
...
+
+from pyglet.math import Mat4
+
+from arcade.cinematic.types import Projector
+from arcade.window_commands import get_window
+if TYPE_CHECKING:
+    from arcade.application import Window
+
+
+class DefaultProjector:
```
gran4: Make it _DefaultProjecter? if you do not want it to be used by a developer.

DragonMoffon: Good idea, thank you!

DragonMoffon:
Moving branch to better reflect the goals of this PR

PR #1827, opened 2023-06-30
Small updates for pygame comparison

Starbuck5:
Hello! I was happy to see pygame-ce had been added to this comparison page, it spurred me to do a few small updates. Hopefully these are agreeable.

- added pygame-ce examples link like pygame
- pygame and pygame-ce are type hinted and have been for a while. Perhaps this is an artifact from before this was the case?
- pygame-ce supports many different kinds of transparency. If you load an image that has transparency it will have transparency. Pygame and pygame-ce support colorkeying too, but we also support per pixel alpha and global alpha for a surface.
- I rewrote note 1 a bit, spurred by "write the image to a surface." Any loaded image is already a Surface, there's no write operation needed. And you don't need to make a Sprite out of a surface to use it, and even if you did that's not an expensive operation in pygame so it's strange to mention.
- Also the links at the bottom were unused as far as I could tell so I removed them

### DIFF
--- a/doc/programming_guide/pygame_comparison.rst
+++ b/doc/programming_guide/pygame_comparison.rst
@@ -35,7 +35,7 @@ Library Information
    * - Example code
      - :ref:`Arcade Examples<example-code>`
      - `Pygame Examples <https://github.com/pygame/pygame/tree/main/examples>`_
-     -
+     - `Pygame-ce Examples <https://github.com/pygame-community/pygame-ce/tree/main/examples>`_
    * - License
      - `MIT License`_
      - LGPL_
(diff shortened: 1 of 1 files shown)


doc/programming_guide/pygame_comparison.rst
```diff
@@ -83,10 +83,10 @@ Here are some comparisons between Arcade 3.0 and Pygame 2.2.0 ce:
      - No
    * - Type Hints
      - Yes
-     - No
+     - Yes
    * - Transparency support
      - Yes
-     - Must specify transparent colorkey
+     - Yes
```
pvcraven, line 89: Does Pygame no longer need to set transparent color key?

Starbuck5: My apologies, I must've missed the notification for your comment on my end.

Pygame Surfaces can and frequently do have other types of transparency, such as per pixel alpha (use pygame.SRCALPHA or convert_alpha()), and global opacity (set_alpha())

einarf:
Any updates on this one?

Starbuck5:
I'd like to get it merged :)

doc/programming_guide/pygame_comparison.rst
pvcraven: Confirmed, loading an image requires no extra steps for transparency support now. Yay!

PR #1829, opened 2023-07-06
Update outdated resolve_resource_path statement in video_player.py

bandit-masked:
Line 26 of video_player.py contains a deprecated `resolve_resource_path` statement. Replace this statement with the updated `resolve` function.


einarf:
Thanks!

Issue #1805, opened 2023-05-28
Rotated objects in Tiled appear in the wrong location


bunny-therapist:
## System Info
Arcade 3.0.0.dev20
------------------
vendor: Intel
renderer: Intel(R) UHD Graphics
version: (3, 3)
python: 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
platform: win32
pyglet version: 2.0.6
PIL version: 9.4.0

Tiled 1.9.2

### Actual behavior:

An object which in Tiled is rotated by 90 degrees appears at a position (-1,+1) relative to where it is placed in Tiled. An object roated by -90 degrees appears at (+1,-1).

### Expected behavior:

A rotated tile appears in an arcade game in the same position as it does in Tiled.

### Steps to reproduce/example code:

Create a map in Tiled. Place an object at a position. Rotate it by 90 degrees by editing its properties in Tiled. Load it using `arcade.load_tilemap` and view the results. Compare the position of the object with where it appears in Tiled.

(I am aware this might possibly be a bug in pytiled-parser rather than arcade, but it is hard for me to determine, so I will create an issue here. If we are sure this is in pytiled-parser, I can create an issue there instead if needed.)

pushfoo:
> An object which in Tiled is rotated by 90 degrees appears at a position (-1,+1) relative to where it is placed in Tiled. An object roated by -90 degrees appears at (+1,-1).

This may be due to rotation changes in 3.0.0. If you have time, could you please add vanicat's fork as a remote and clone #1741 to see if it fixes the problem?

I haven't worked with tiled much and I'm not sure if I have it installed at the moment, so I'm not the best person to investigate this.

Possibly related / dupes:

* #1740
* #1763

Cleptomania:
I believe this is fixed in dev22. See https://github.com/pythonarcade/arcade/issues/1695

Let me know if it’s not the same as that issue, but it sounds like the same thing to me. I’m unfortunately relatively unavailable to work on much of anything for at least the next probably 3 months, so not sure when I would be able to get to fixing this.

bunny-therapist:
I will try dev22 if it is released.

Btw, this is very similar to something I previously reported: https://github.com/pythonarcade/arcade/issues/928
I think arcade would benefit from some kind of tests to guard against regressions. For example, have a small Tiled map with rotated objects that gets loaded in a test and then the position of the objects gets asserted.

[...]

PR #1831, opened 2023-07-10
Solution for problem with rotation in tiled ?

vanicat:
Several test for rotation, and probable fix for #1805

A sign had to be changed in the tiled loader.

I've add to the :fixtures:tilemaps/rotation.json an object layer, with object that are tile, and are rotated. And arcade do not place some of the rotated tile correctly.

I believe that this is related to #1805, but I didn't not check for the correct direction of the rotation yet.


bunny-therapist:
This does in fact solve #1805 
I created https://github.com/bunny-therapist/arcade_issue_1805 in order to demonstrate that issue, and installing from your fork solves both the problem of bad rotation and the problem of placement.

vanicat:
I've submitted this pull request some time ago now. 

Is-there something I should do to make it go forward ?

Thanks.

Cleptomania:
Sorry for the long wait on this one, I've been out of commission on Arcade for a while, looks good though.

PR #1833, opened 2023-07-12
Two fix to import of tiled image layer

vanicat:
I tried to fix the import of tiled image layer
- the Call to Sprite was made with an incorrectly named keyword argument
- corrected the placement of the new sprite

I also added test for image layer import.


einarf:
Thanks!

PR #1850, opened 2023-08-04
WIP: Switch to callback function for custom classes in TileMap.

### LINKED ISSUE #1840: Allow more flexibility with regard to sprite classes when loading Tiled maps

Currently, you can pass `custom_class` kwarg to `arcade.load_tilemap`, but there are several limitations:
1. your class must subclass `arcade.Sprite` - it is not enough to subclass `arcade.BaseSprite` (even though it probably should be, otherwise the usefulness of `arcade.BaseSprite` is very limited).
2. if there are animated tiles in a Tiled layer which is loaded with `custom_class`, then the custom class must subclass `arcade.AnimatedTimeBasedSprite`, which (more or less) means that *all* tiles in that layer must be animated. (This is basically a consequence of there only being a single custom class for the entire layer.)

I suggest that instead of passing in a single custom class, it should be possible to pass in some kind of callback function that creates a sprite from a tile. Given enough information about the tile (e.g. if it is animated or not), that function could then decide to return an animated sprite or something else. It should be enough for that function to return `arcade.BaseSprite` - it should not be necessary to subclass `arcade.Sprite`.

### What would it help with?

- Users can create their own subclasses of `arcade.BaseSprite` without losing out on using `arcade.load_tilemap`.
- Users can create different sprites for different tiles in their map without restrictions based on if they are animated or not. For example, one could make a callback create-sprite function which uses the "class" property of tiles in Tiled to instantiate different classes for enemies, items, etc, and still use `arcade.AnimatedTimeBasedSprite` for some simple animated tiles.

Cleptomania:
This is a PR to address #1840 by changing the `custom_class` option of the `layer_options` dictionary in TileMap to a new `custom_class_callback` function which allows users to define a callable function that will return a class type to use for all of the sprites in that layer.

This does completely remove the `custom_class` option, it doesn't feel necessary to support both, as this is a fairly advanced feature anyways, I don't think anyone who is making use of it will have a problem switching to the new system, as it also affords them a lot more flexibility in how they handle this.

Currently this does not address the concerns in the issue around `arcade.BasicSprite`, and is only limited to the callback function idea.

### DIFF
--- a/arcade/tilemap/tilemap.py
+++ b/arcade/tilemap/tilemap.py
@@ -18,13 +18,7 @@ import pytiled_parser
 import pytiled_parser.tiled_object
 from pytiled_parser import Color
 
-from arcade import (
-    AnimatedTimeBasedSprite,
-    AnimationKeyframe,
-    Sprite,
-    SpriteList,
-    get_window,
-)
+from arcade import AnimatedTimeBasedSprite, AnimationKeyframe, Sprite, SpriteList, get_window
 from arcade.hitbox import HitBoxAlgorithm, RotatableHitBox
 from arcade.texture.loading import _load_tilemap_texture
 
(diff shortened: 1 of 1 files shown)


Cleptomania:
I'm going to go ahead and close this. This feature needs more thought and maybe we can revisit it in 3.1.

PR #1854, opened 2023-08-06
Fixed PyRight issues.

DragonMoffon:
The latest version of PyRight is a little over zealous.. There where two flagged non-issues. One where it didn't realise that the code would escape early, and another which is all pyglets fault.

### DIFF
--- a/arcade/gui/widgets/text.py
+++ b/arcade/gui/widgets/text.py
@@ -552,7 +552,7 @@ class UITextArea(UIWidget):
     def on_event(self, event: UIEvent) -> Optional[bool]:
         if isinstance(event, UIMouseScrollEvent):
             if self.rect.collide_with_point(event.x, event.y):
-                self.layout.view_y += event.scroll_y * self.scroll_speed   # type: ignore
+                self.layout.view_y += event.scroll_y * self.scroll_speed   # type: ignore  # pending https://github.com/pyglet/pyglet/issues/916
                 self.trigger_full_render()
 
         if super().on_event(event):
--- a/arcade/gui/widgets/text.py
+++ b/arcade/gui/widgets/text.py
@@ -23,7 +23,6 @@ from arcade.gui.widgets.layout import UIAnchorLayout
 from arcade.types import RGBA255, Color, RGBOrA255, RGB
 
 
-
 class UILabel(UIWidget):
     """A simple text label. This widget is meant to display user instructions or
     information. This label supports multiline text.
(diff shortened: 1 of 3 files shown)


cspotcode (commented):
When `# type: ignore` is due to pyglet issues, we have a convention of adding a comment linking to the pyglet issue.  Should we keep that convention?  You can search for `# pending` to see it.

For example: https://github.com/pythonarcade/arcade/blob/development/arcade/sound.py#L143

arcade/physics_engines.py
```diff
@@ -53,7 +54,7 @@ def _circular_check(player: Sprite, walls: List[SpriteList]):
         vary *= 2
 
 
-def _move_sprite(moving_sprite: Sprite, walls: List[SpriteList], ramp_up: bool) -> List[Sprite]:
+def _move_sprite(moving_sprite: Sprite, walls: List[SpriteList], ramp_up: bool) -> List[Union[Sprite, BasicSprite]]:
```
cspotcode:
```suggestion
def _move_sprite(moving_sprite: Sprite, walls: List[SpriteList[SpriteType]], ramp_up: bool) -> List[SpriteType]:
```

I think this should be generic.  The returned list contains wall sprites, so it'll return whatever type of BasicSprite (sub)class is contained in the `walls` SpriteLists.

arcade/gui/widgets/text.py
```diff
@@ -488,8 +487,10 @@ def __init__(
 
         # Set how fast the mouse scroll wheel will scroll text in the pane.
         # Measured in pixels per 'click'
-        self.scroll_speed = scroll_speed if scroll_speed is not None \
-                            else font_size
+        self.scroll_speed = (
+            scroll_speed if scroll_speed is not None
+            else font_size
+        )
```
cspotcode, line 493: Is this related to pyright?  It's fine if not, the code looks better when parens.  Just wondering.

DragonMoffon: The line was too long and had originally used a slash which I find hard to read. If you have a recommendation on how to format it I'm happy to update it

cspotcode: No recommendation, I wanted to confirm it wasn't pyright choking on backslashes or something.  I agree with your formatting.

arcade/physics_engines.py
pushfoo: @DragonMoffon Bumping since this PR fixes issues which flag #1876 as broken.

Issue #1839, opened 2023-07-19
Text alignment issue in arcade 3


bunny-therapist:
## System Info

```
Arcade 3.0.0.dev24
------------------
vendor: Intel
renderer: Intel(R) UHD Graphics
version: (3, 3)
python: 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
platform: win32
pyglet version: 2.0.8
PIL version: 9.4.0
```

### Actual behavior:

arcade.Text objects with align="center" are not center aligned (the multiline kwarg seems to also influence this; at least sometimes the text becomes aligned by setting multiline=True). It looks fine in arcade 2.6.17 but not in 3.0.0.dev24.

### Expected behavior:

Text that was center aligned in 2.6.17 should still be in the same position when upgrading to arcade 3.0.0.

### Steps to reproduce/example code:

Run this code with arcade==2.6.17 - the text is in the middle of the window.
Run this code with arcade==3.0.0.dev24 - the text is off to the right.
```
import arcade


class IssueDemonstrationWindow(arcade.Window):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.text = arcade.Text(
            text="MIDDLE",
            start_x=400,
            start_y=300,
            font_size=72,
            width=800,
            anchor_x="center",
            align="center",
        )
        self.set_viewport(
            left=0,
            right=800,
            bottom=0,
            top=600,
        )

    def on_draw(self):
        self.text.draw_debug()


window = IssueDemonstrationWindow()
arcade.run()

```

bunny-therapist:
It seems like the problem disappears if you remove the "width" argument.

gran4:
The text fits into the box if you comment out `anchor_x="center",`
But, the box is not in the middle if you do that.

gran4:
I think I found the issue. If you increase the width, it moves the text even more:
This is because line 1411 in layout.py in pyglet `return self._x - width // 2`

`layout.TextLayout` inside `DocumentLabel` inside `pyglet.text.Label`

P.S: I will make an issue in pyglet.

gran4:
According to caffinepills, it is not a bug, it is intended.

bunny-therapist:
Why did it change in arcade between 2.6.17 and 3? If this is due to a change in pyglet version, does that mean that pyglet intended for this to change?

bunny-therapist:
[...]

PR #1864, opened 2023-08-07
Fix https://github.com/pythonarcade/arcade/issues/1839[WIP]

gran4:
It seems like there is a problem unrelated to my code that is breaking the tests.
Do not merge, there is an on going conversation on whether this was removed because it broke something.

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -198,7 +198,7 @@ class Text:
         if align != "center" and align != "left" and align != "right":
             raise ValueError("The 'align' parameter must be equal to 'left', 'right', or 'center'.")
 
-        if align is "center":
+        if align == "center":
             multiline = True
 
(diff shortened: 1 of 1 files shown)


einarf:
Related to this? https://github.com/pythonarcade/arcade/issues/1839

eruvanos:
Hi,

there is an issue with the UI.

UILabel often gets its real width during the layout phase, which means it has no initial width given. Still setting align=center will effect the text position after the layout phase and center the text.

Also from my point of view it makes no sense, to enforce multiline based on the alignment. Also it overwrites the given value, which feels unintuitive from my point of view.

eruvanos:
Update: I will remove the multiline overwrites when align set !=left in all arcade.text areas.
While align might not have an effect without multiline, the given multiline value should not be overwritten by some magic.

PR #1868, opened 2023-08-08
typing(sprites): enhance typings on Sprite and SpriteList

Snipy7374:
i'm not entirely sure if i'm missing some of the functions [here](https://github.com/pythonarcade/arcade/compare/development...Snipy7374:arcade:development?expand=1#diff-67172d8b71c0ebbf18528a297ec5f527af418cac4c5a6ca4fb832a4caa311a50R58-R59) and [here](https://github.com/pythonarcade/arcade/compare/development...Snipy7374:arcade:development?expand=1#diff-67172d8b71c0ebbf18528a297ec5f527af418cac4c5a6ca4fb832a4caa311a50R65-R68), if that's so let me know!


arcade/sprite/sprite.py
```diff
@@ -313,7 +313,13 @@ def stop(self) -> None:
 
     # ---- Draw Methods ----
 
-    def draw(self, *, filter=None, pixelated=None, blend_function=None) -> None:
+    def draw(
+            self,
+            *,
+            filter: Optional[OpenGlFilters] | None = None,
```
pushfoo:
```suggestion
            filter: Optional[OpenGlFilters] = None,
```
Although the `|` syntax is shorter, it's not appropriate here:
1. Arcade has to support 3.8+, but `|` is a 3.10+ feature.
2. You've already used `Optional` to mark the possible `None` value.

arcade/types.py
```diff
@@ -49,6 +52,22 @@
...
+# equivalent to the following blend_functions (under arcade.gl.enums):
+# ZERO, ONE, SRC_COLOR, ONE_MINUS_SRC_COLOR,
+# SRC_ALPHA, ONE_MINUS_SRC_ALPHA, DST_ALPHA,
+# ONE_MINUS_DST_ALPHA, DST_COLOR, ONE_MINUS_DST_COLOR
+BlendFunctions: TypeAlias = Union[Literal[0], Literal[1], Literal[768],
+                                  Literal[769], Literal[770], Literal[771],
+                                  Literal[772], Literal[773], Literal[774],
+                                  Literal[775]]
+
+
```
pushfoo: tl;dr 
1.`Literal` is not the right choice here
2. `IntEnum` may be better.

@DragonMoffon @einarf Might have better feedback on this since they're the most involved OpenGL users on the server.

### Literal Defeats the Point of `enums.py`

We want to use the constants rather than require users to enter literal values each time. However, `Literal` is an indication they should use the latter.

In the example below, `Literal` causes checkers to mark passing `CONST_FOO` as incorrect.

```py
from typing import Union, Literal
# TypeAlias was added in 3.10 and we support 3.8+, but we
# can still use if if we import it from the backport package.
from typing_extensions import TypeAlias

CONST_FOO = 0xF00
CONST_BAZ = 0x847

Constant: TypeAlias = Union[Literal[2119], Literal[3840]]

def print_raw_value(c: Constant) -> None:
    print(f"value is {c}")

if __name__ == "__main__":
    # Type checkers will mark using CONST_FOO as wrong
    print_raw_value(CONST_FOO)
```

### `IntEnum` as a sidegrade

[`IntEnum`](https://docs.python.org/3/library/enum.html?highlight=intenum#enum.IntEnum) is largely compatible with `int`s during execution. Subclasses also raise `ValueError`s if a user attempts to create an instance from `int`s not included in the type declaration.

However, its performance impacts are unclear. Using its subclasses as annotations is also a double-edged sword: arguments will be marked as incorrect by type checkers if someone passes an int:

```python
from enum import IntEnum

class ConstantType(IntEnum):
    CONST_FOO = 0xF00
    CONST_BAZ = 0x847

def print_raw_value(c: ConstantType) -> None:
    # The upside is that we get free, 1-line value checking. For example,
   # the line below will raise a ValueError if c isn't in in the enum.
    enum_value = ConstantType(c)
    print(f"value is {enum_value}")

if __name__ == "__main__":
    # The downside is we have to either access the members directly, or declare
    # the enum "twice" by declaring module-level variables for each of member.
    print_raw_value(ConstantType.CONST_FOO)
    # type checkers will mark this as wrong
    print_raw_value(0)
```

arcade/scene.py
```diff
@@ -432,9 +432,9 @@ def update_animation(
     def draw(
         self,
         names: Optional[Iterable[str]] = None,
-        filter: Optional[int] = None,
+        filter: Optional[OpenGlFilters] = None,
         pixelated: bool = False,
-        blend_function: Optional[Union[Tuple[int, int], Tuple[int, int, int, int]]] = None,
+        blend_function: Optional[BlendFunctions] = None,
```
pushfoo: It looks like `ArcadeContext` still has references to what look like tuple blend functions as class-level constants, so this may be a premature change.

arcade/types.py
```diff
@@ -17,7 +17,10 @@
     Sequence,
     Tuple,
     Union,
-    TYPE_CHECKING, TypeVar
+    TYPE_CHECKING,
+    TypeVar,
+    TypeAlias,
+    Literal,
 )
 from typing_extensions import Self
```
pushfoo:
```suggestion
    Literal,
)
from typing_extensions import Self, TypeAlias
```
We have to import `TypeAlias` from the backports package since it's a 3.10 feature.

arcade/types.py
```diff
@@ -49,6 +52,22 @@
 
 RGBA255OrNormalized = Union[RGBA255, RGBANormalized]
 
+# equivalent to the following filters (under arcade.gl.enums):
+# NEAREST, LINEAR, NEAREST_MIPMAP_NEAREST, LINEAR_MIPMAP_NEAREST,
+# NEAREST_MIPMAP_LINEAR, LINEAR_MIPMAP_LINEAR
+OpenGlFilters: TypeAlias = Union[Literal[9728], Literal[9729], Literal[9984],
```
pushfoo:
```suggestion
OpenGlFilter: TypeAlias = Union[Literal[9728], Literal[9729], Literal[9984],
```
Since it's a single object, should this be singular? `ArcadeContext` seems to confirm it as such given its `blend_func` instance variable is singular.

arcade/types.py
```diff
@@ -49,6 +52,22 @@
...
+                                 Literal[9985], Literal[9986], Literal[9987]]
+
+# equivalent to the following blend_functions (under arcade.gl.enums):
+# ZERO, ONE, SRC_COLOR, ONE_MINUS_SRC_COLOR,
+# SRC_ALPHA, ONE_MINUS_SRC_ALPHA, DST_ALPHA,
+# ONE_MINUS_DST_ALPHA, DST_COLOR, ONE_MINUS_DST_COLOR
+BlendFunctions: TypeAlias = Union[Literal[0], Literal[1], Literal[768],
```
pushfoo:
```suggestion
BlendFunction: TypeAlias = Union[Literal[0], Literal[1], Literal[768],
```
Same comment as above: since it represents a single piece of GL configuration, should it be singular?

pushfoo (changes requested):
Thank you for taking a look at this, and I apologize for the delay in reviewing it.

tl;dr I think the physics engine changes are fine, but I have concerns about the use of `Literal`, the type aliases created with it, and the performance impacts of the alternatives.

I'd spend more time looking at `ArcadeContext` and `gl.Context` to better understand blend functions and our current implementations of them.

If you've got Discord, feel free to join the server and bring these changes up in `#arcade-dev-1`.

arcade/sprite/sprite.py
Snipy7374: yeah sorry, I forgot to remove the `| None` part

arcade/types.py
Snipy7374: I already asked to make some of the members in `enums.py` actual enum classes but the response was negative so idk

pushfoo: > but the response was negative

That's disappointing.  Do you remember where that discussion took place, anything about the context, or the reasons given for avoiding it? I was hoping to explore moving to `IntEnum` in the long term once more pressing matters are settled.

Snipy7374: tho i feel like that it's not a definitive answer https://discord.com/channels/458662222697070613/458662458198982676/1137813484344721489

cspotcode: Let's not invest too much energy here.  If we use `PyGlEnum`, it's less contentious and we can get this merged.

https://github.com/pythonarcade/arcade/blob/09832834846cae44068e752ff0d96f80c28f9034/arcade/gl/types.py#L12-L13

Using these Unions of literals is *technically* more strict, but I don't think it's practical.  It will cause more friction than benefit.  It also add repetition.

PR #1876, opened 2023-08-13
Fix exception type when Window.show_view is passed a non-View

pushfoo:
The errors reported for this PR are fixed by #1854

### Changes

1. Raise a `TypeError` instead of a `ValueError` when `Window.show_view` is passed something other than a view
2. Include the passed object's string representation in the `TypeError`'s string
3. Test the exception type in Window's unit tests

### How to test

Run unit tests as usual.

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -721,8 +721,9 @@ class Window(pyglet.window.Window):
         :param View new_view: View to show
         """
         if not isinstance(new_view, View):
-            raise TypeError("Must pass an arcade.View object to "
-                             f"Window.show_view() {type(new_view)}")
+            raise TypeError(
+                f"Window.show_view() takes an arcade.View,"
+                f"but it got a {type(new_view)}: {new_view}")
 
         # Store the Window that is showing the "new_view" View.
         if new_view.window is None:
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -721,7 +721,7 @@ class Window(pyglet.window.Window):
         :param View new_view: View to show
         """
         if not isinstance(new_view, View):
-            raise ValueError("Must pass an arcade.View object to "
+            raise TypeError("Must pass an arcade.View object to "
                              f"Window.show_view() {type(new_view)}")
 
         # Store the Window that is showing the "new_view" View.
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -723,7 +723,7 @@ class Window(pyglet.window.Window):
         if not isinstance(new_view, View):
             raise TypeError(
                 f"Window.show_view() takes an arcade.View,"
-                f"but it got a {type(new_view)}: {new_view}")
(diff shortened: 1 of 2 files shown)


arcade/application.py
```diff
@@ -721,8 +721,9 @@ def show_view(self, new_view: 'View'):
         :param View new_view: View to show
         """
         if not isinstance(new_view, View):
-            raise ValueError("Must pass an arcade.View object to "
-                             f"Window.show_view() {type(new_view)}")
+            raise TypeError(
+                f"Window.show_view() takes an arcade.View,"
+                f"but it got a {type(new_view)}: {new_view}")
```
cspotcode: Are there any cases where it's dangerous or too verbose to convert any wrong value to a string?  For example, if someone reads an XML file and then accidentally passes that into `show_view()`, then they'll get pages of XML blowing away the rest of the error message?  Maybe I'm overthinking

pushfoo: tl;dr It's a decent idea, but it might belong in a separate PR.

If we truncate values here, we should also support doing so elsewhere. That makes it a more complicated design issue than just changing this file.

How about these as part of a follow-up?
1. Move `utils.*Error` into an `arcade.exceptions` module
2. Add an output templating function with a `truncate_value: Optional[int] = None` keyword

I'm not sure how we should handle global configuration of value truncation. I don't see an `arcade.options` object like pyglet's. Keeping a default at the module level is definitely possible, even if it is a little ugly.

cspotcode: Why can't we avoid logging the value at all, only log the type?

pushfoo: 1. It's inconsistent with the rest of the framework
2. It hides the reason for the failure
3. These are both especially hard on beginners

Pages of XML in the output is useful feedback during development since the contents inform you where the value came from in your code. This specific case is unlikely to be encountered by a beginner.

A more likely case is a user passing something like `{'name':'Player1', 'score':1}`. Printing the value helps them understand where the value came from. Beginning users will not understand using a debugger.

cspotcode: > It's inconsistent with the rest of the framework

I don't think that's true. I searched the codebase for `TypeError`s.  They're not consistent, but *most* include type without value.

I don't wanna block on this, that's just my 2cents.

pushfoo: tl;dr I've reverted the change for now since this is touching on larger exception and style issues.

> I don't think that's true.

> They're not consistent, but most include type without value.

I looked at recent changes by einarf and Clepto to  `hit_box.py`, sprite, and texture caching as reference. Those areas match what I understand to be good practice. 

Much of the codebase doesn't match it. There are also intentional backward-compatible inconsistencies, such as `draw_lrtb_rectangle_filled` in `draw_commands.py`, which still raises `AttributeError` since it's a deprecated and won't be fixed.

PR #1881, opened 2023-08-15
Camera Overhaul

### LINKED ISSUE #1853: Complete Camera Overhaul
The camera provided in Arcade currently has many issues that need to be fixed.
• Only provides a 2D orthographic camera
• Has janky and incorrect zoom functionality
• Has an update/use method that breaks arcade conventions
• Does not provide a way to rotate the camera
• SimpleCamera is not simple as it should be
• Normal camera has features that would be better served as an external class(es). 

Okay, point three about the update methods might need explaining. The camera provides an update method, but it does not respect delta_time and is automatically called in the use method without a way to disable it. This means that camera speeds are variable based on the update rate and how often a camera is used in a single draw call. On top of this, it also breaks arcade conventions by having non-draw logic within the draw call. 

I propose that the cameras in Arcade be totally overhauled. 

Proposed Changes:
• Provide simple data classes to store information about cameras.
• Provide an interface called Projector for objects, which sets the underlying arcade matrices.
• Create a Default Projector, Orthographic Projector, and Perspective Projector, which satisfy most people's needs.
• Use the Default Projector to allow users to use cameras in context managers safely
• Create a new Camera2D that follows conventions and makes it easy for users to plug and play.
• Provide controller classes that users can use to move their Cameras easily in complex ways. [EDIT: these have been called 'grips' after the colloquial term for camera, and other operators on movie sets]
• An all-new tutorial explaining how cameras in games work and how to use the arcade system.
• Move all examples over to using Camera2D and Grips. 

Why do it this way?:
Cameras are just represented by PoDs (Packets of Data) rather than fully fledged classes to promote composition over inheritance. While Camera2D can be treated as though it stores all the data internally, the user should look to the PoDs for more complex use cases. An example benefit is when a user wants multiple projections to share a single position. If you have multiple cameras with the same projection but are placed in different locations (like a CCTV system, for example), having them all share one projection PoD would save memory and decrease the chance of mistakes. 

It also means that camera controllers don't need to know anything about the camera they are moving, and they can know about the position, direction, etc. 

These breaking changes should occur in 3.0, with later improvements and tutorials coming in 3.1+. 

TODO LIST:
- [x] Dataclasses for view and projection matrices
     - [x] Camera Data (viewport, view data, zoom)
     - [x] Orthographic Projection
     - [x] Perspective Projection
- [x] Projector and Camera Protocols for typing
- [x] Default Projectors
     - [x] A simple projector that Arcade uses internally 
     - [x] An orthographic projector capable of both 2D and 3D
     - [x] A perspective projector capable of both 2D and 3D
- [x] A competent replacement for `SimpleCamera`
- [x] A re-implementation of `SimpleCamera` with a depreciation warning
- [x] Provide camera controllers for ease of use
     - [x] A track-to-point method as seen in `SimpleCamera`
     - [x] A lerp to point method (2D and 3D)
     - [x] An lerp to point which accepts an easing function (2D and 3D)
     - [x] Methods to rotate the camera around the view axis (forward, up, right)
     - [x] 2D screen shake
- [x] VERY IMPORTANT. Remove internal assumptions about projection being Orthographic
- [x] Make `Window` and `ArcadeContext` consistent about the name of `projection` and `viewport`
- [ ] Complete typing and Doc Strings
- [x] Fix all examples which used the old camera
- [ ] Adding new examples 
     - [ ] One that shows off the easing methods
     - [x] A replacement for the `SimpleCamera` example
     - [ ] An isometric example
- [ ] A detailed tutorial page on how the cameras work
- [ ] A camera section for the API quick index

DragonMoffon:

The refactor includes:
Two ProjectionData PoDs hold the data for making either an orthographic or projection matrix.

See Issue #1853 for more details


DragonMoffon:
@cspotcode Sorry you're going to have to add me as an assignee again.

pushfoo:
Are you ok with me making PRs into this branch at some point?

DragonMoffon:
@pushfoo go right ahead

pushfoo:
I'll do so as time allows. Don't worry about overwriting my changes. I'll try to keep them to technical information only and will do more final grammar / formatting at a later point.

I'm also wondering if we should try to break this thing into chunks. I'm not sure how to do that yet, but it would make merging and reviewing it easier if we could.

pushfoo:
@DragonMoffon would it be a good idea to mark this as WIP for now?

DragonMoffon:
@pushfoo Id rather we took the unstable stuff out and put that into WIP. I really want the core work pulled so people can test it in actual use cases.

DragonMoffon:
I'm splitting the PR so the stable work can be pulled. Will make into another PR.

PR #1884, opened 2023-08-19
Update build procedure

pvcraven:
* Github actions will now run only the linting/typing on Python 3.11
* Self-hosted runners will run unit tests on Python 3.8-3.11.
* There are four self-hosted runners now going on my laptop to speed things along
* Updated the names of things, re-organized a bit
* Fixed a few things that were Python 3.9 only


pyproject.toml
```diff
@@ -45,16 +45,16 @@ dev = [
...
     "furo",
     "pyright==1.1.322",
     "pyyaml==6.0",
-    "sphinx==7.1.2",
+    "sphinx==7.2.2",
     "sphinx-autobuild==2021.3.14",
     "sphinx-copybutton==0.5.2",
     # Intentionally kept at 2.3 until this bugfix is published: https://github.com/jdillard/sphinx-sitemap/pull/62
-    "sphinx-sitemap==2.3.0",
+    "sphinx-sitemap==2.5.1",
```
cspotcode, line 57: I know they said the fix is published 2 days ago, but have we verified ourselves that incremental rebuilds are actually working on the new version?

If this does fix the bug, then we should remove the comment on the line above.

doc/tutorials/menu/menu_05.py
```diff
@@ -155,7 +155,7 @@ def on_draw(self):
 class SubMenu(arcade.gui.UIMouseFilterMixin, arcade.gui.UIAnchorLayout):
     """Acts like a fake view/window."""
 
-    def __init__(self, title: str, input_text: str, toggle_label: str, dropdown_options: list[str], slider_label: str):
+    def __init__(self, title: str, input_text: str, toggle_label: str, dropdown_options: list, slider_label: str):
```
cspotcode: Should this be `List[str]` with `from typing import List`?

pvcraven: That's not legal in Python 3.8

pyproject.toml
pvcraven: I have not verified the fix works for us.

doc/tutorials/menu/menu_05.py
pvcraven: Updated

pyproject.toml
cspotcode: If running `python ./make.py html` twice in a row, the second time should be way faster, that's how we know it's working.  The bug affects its ability to un-pickle the cache from the previous invocation.

PR #1905, opened 2023-09-20
Camera Overhaul Core

DragonMoffon:
This PR includes the core functionality in a stable (without unit tests or docs) state. This includes a stripping out of all references to the old system. 

Removed:
- the set_viewport method. Has been replaced by a variety of options. Either directly setting matrices, using the default camera, or using a Camera2D
- The old Camera class. This will not be replaced in a backward-compatible manner as it is deeply flawed and has been replaced by Camera2D

Removed but coming in the future:
- Screen shake.

The refactor includes:
Two ProjectionData PoDs hold the data for making either an orthographic or projection matrix.

See Issue https://github.com/pythonarcade/arcade/issues/1853 for more details


alejcas:
Quick question: Does this new camera work with arcade collision methods when applying zoom or rotation?

DragonMoffon:
@janscas Generally they should as the collisions are done in world space which the cameras don't touch. The cameras only affect where it looks like sprites are. Should still be double-checked, but I would be surprised.

doc/tutorials/lights/01_light_demo.py
```diff
@@ -25,9 +25,8 @@ def __init__(self, width, height, title):
         # Physics engine
         self.physics_engine = None
 
-        # Used for scrolling
-        self.view_left = 0
-        self.view_bottom = 0
+        # camera for scrolling
+        self.cam = None
```
pvcraven: I'd prefer to spell out camera.

tests/unit/camera/test_default_camera.py
pvcraven, line 1: Are these placeholder files? Still looking for tests? If so, probably need some comments stating that.

arcade/camera/simple_camera.py
```diff
@@ -0,0 +1,381 @@
...
+                 position: Optional[Tuple[float, float]] = None,
+                 up: Optional[Tuple[float, float]] = None,
+                 zoom: Optional[float] = None,
+                 near: Optional[float] = None,
+                 far: Optional[float] = None,
+                 camera_data: Optional[CameraData] = None,
+                 projection_data: Optional[OrthographicProjectionData] = None
+                 ):
+        """
+        Initialise a Simple Camera Instance with either Camera PoDs or individual arguments
```
pvcraven: "Initialize" for u.s. spelling.

arcade/examples/platform_tutorial/06_camera.py
```diff
@@ -121,19 +121,17 @@ def on_key_release(self, key, modifiers):
...
-        if screen_center_y < 0:
-            screen_center_y = 0
+        if screen_center_x - self.width/2 < 0:
+            screen_center_x = self.width/2
+        if screen_center_y - self.height/2 < 0:
+            screen_center_y = self.height/2
         player_centered = screen_center_x, screen_center_y
 
-        self.camera.move_to(player_centered)
+        arcade.camera.controllers.simple_follow_2D(0.2, player_centered, self.camera.view_data)
```
pvcraven: Probably should use a named parameter or variable so that 0.2 isn't a random number appearing out of no where. I think elsewhere you used a camera_panning variable of some sort.

pvcraven (changes requested):

tests/unit/camera/test_default_camera.py
DragonMoffon: Yup, tests are still not done.

arcade/camera/camera_2d.py
```diff
@@ -0,0 +1,713 @@
...
+                    Defaults to the currently active window.
+            viewport: A 4-int tuple which defines the pixel bounds which the camera with project to.
+            position: The 2D position of the camera in the XY plane.
+            up: The 2D unit vector which defines the +Y-axis of the camera space.
+            zoom: A scalar value which is inversely proportional to the size of the camera projection.
+                  i.e. a zoom of 2.0 halves the size of the projection, doubling the perceived size of objects.
+            projection: A 4-float tuple which defines the world space
+                        bounds which the camera projects to the viewport.
+            near: The near clipping plane of the camera.
+            far: The far clipping place of the camera.
```
MiCurry: Should `place` be `plane` here?

arcade/camera/camera_2d.py
```diff
@@ -0,0 +1,713 @@
...
+
+__all__ = [
+    'Camera2D'
+]
+
+
+class Camera2D:
+    """
+    A simple orthographic camera. Similar to SimpleCamera, but takes better advantage of the new data structures.
+    As the Simple Camera is depreciated any new project should use this camera instead.
```
MiCurry: There should be a comma after 'deprecated':

> As the Simple Camera is depreciated, any new project should use this camera instead.

arcade/camera/camera_2d.py
DragonMoffon: plane

arcade/camera/camera_2d.py
```diff
@@ -0,0 +1,713 @@
...
+            window=self._window,
+            view=self._data,
+            projection=self._projection
+        )
+
+    @property
+    def view_data(self) -> CameraData:
+        """
+        Return the view data for the camera. This includes the
+        viewport, position, forward vector, up direction, and zoom.
```
MiCurry: I think `CameraData` just contains `position`, `up`, `forward` and `zoom` and **not** the `viewport`, correct?

MiCurry: A print of the view_data property shows:
```
CameraData(position=(0.0, 0.0, 0.0), up=(0.0, 1.0, 0.0), forward=(0.0, 0.0, -1.0), zoom=1.0)
```

DragonMoffon: yup this doc string was written before a change in the data structures good call

arcade/camera/camera_2d.py
```diff
@@ -0,0 +1,713 @@
...
+        self._projection.viewport = (self._projection.viewport[0], self._projection.viewport[1],
+                                     _width, self._projection.viewport[3])
+
+    @property
+    def viewport_height(self) -> int:
+        """
+        The height of the viewport.
+        Defines the number of pixels drawn too vertically.
+        """
+        return self._projection.viewport[3]
```
MiCurry, line 510: Same thing as above with the viewport height. This will fail if the viewport starts at anything other than the origin. I think it should be:

`return self._projection.viewport[3] - self._projection.viewport[1]`

arcade/camera/camera_2d.py
```diff
@@ -0,0 +1,713 @@
...
+    def viewport(self, _viewport: Tuple[int, int, int, int]) -> None:
+        self._projection.viewport = _viewport
+
+    @property
+    def viewport_width(self) -> int:
+        """
+        The width of the viewport.
+        Defines the number of pixels drawn too horizontally.
+        """
+        return self._projection.viewport[2]
```
MiCurry, line 497: I think this is the viewport width only when the viewport starts at (0,0) (e.g. (0, 0, x, y), but this will produce an incorrect width when the viewport starts at anything other than the origin. Should it be something like:

`return self._projection.viewport[2] - self._projection.viewport[0]`

arcade/camera/camera_2d.py
DragonMoffon: The viewport is stored as left, bottom, width, and height. So this is not necessary.

arcade/camera/camera_2d.py
DragonMoffon: See the previous comment with the viewport being stored as left, bottom, width, and height.

DragonMoffon:
Closed as a new PR was made to replace it. Third times a charm :)

PR #1922, opened 2023-10-07
Add Sound page to the Programming Guide

pushfoo:
Closes #1274
Closes #1900

### Changes

1. Add link targets to the platformer tutorial
2. Add new Sound page to the programming guide covering usage and compatibility
3. Link the Sound page from the top-level index
 
### How to Test

1. Clone the branch locally
2. Run `pip install -I -e .[dev]` to make sure you have the dev dependencies fully installed & upgraded
3. `./make.py html` (optional, but helps prevent an occasional rebuild loop)
4. `./make.py serve`
5. Open https://localhost:8000 in your browser
6. Verify the page is in the sidebar and Programming Guide homepage section
7. Open http://localhost:8000/programming_guide/sound.html
8. Proof-read the Sound Basics sections brutally:
    * http://localhost:8000/programming_guide/sound.html#sound-basics-loading
    * http://localhost:8000/programming_guide/sound.html#sound-basics-playing
    * http://localhost:8000/programming_guide/sound.html#sound-basics-stopping
9. Verify that [#1900](https://github.com/pythonarcade/arcade/issues/1900) is covered by http://localhost:8000/programming_guide/sound.html#backends-determine-playback-features 
10. If time / interest allow, proof read and verify other parts

### Follow-up work for future PRs

1. Move some of the preloading concepts to a new performance page / heading
2. Add an easier to digest reference counting section somewhere and link it instead of the Python contributor doc
3. Move some compatibility information to its own page / heading

PR #1926, opened 2023-10-16
Fix exception types in arcade.sound

pushfoo:
### Changes

1. Use `TypeError` instead of `ValueError` when the wrong types are passed to sound helpers
2. Add unit tests to verify this + helper text phrasing
3. Turn a raise into a raise from
4. Rephrase some exceptions and restructure their generation

The items in 4 are generally well-separated out and easy to revert, and I can understand if reviewers think we should do that.

This is a follow-up to some reading done for #1922.

### DIFF
--- a/arcade/sound.py
+++ b/arcade/sound.py
@@ -210,8 +210,8 @@ def play_sound(
     elif not isinstance(sound, Sound):
         raise TypeError(
             f"Error, got {sound!r} instead of an arcade.Sound."
-            f"" if not isinstance(sound, (str, Path, bytes)) else\
-            " Make sure to use load_sound first, and use the result with play_sound.")
+            if not isinstance(sound, (str, Path, bytes)) else\
+            " Make sure to use load_sound first, then play the result with play_sound.")
 
     try:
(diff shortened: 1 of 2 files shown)


arcade/sound.py
```diff
@@ -206,12 +206,13 @@ def play_sound(
...
-        msg = (
-            "Error, passed in a string as a sound. "
-            "Make sure to use load_sound first, and use that result in play_sound."
-        )
-        raise Exception(msg)
+
+    elif not isinstance(sound, Sound):
+        raise TypeError(
+            f"Error, got {sound!r} instead of an arcade.Sound."
+            f"" if not isinstance(sound, (str, Path, bytes)) else\
```
eruvanos: Should not be an f string

eruvanos (changes requested):
Will merge after the small fix

arcade/sound.py
pushfoo: I'm assuming you mean the extra bit of f-string on 213. That's fixed.

eruvanos (approved):

Issue #1910, opened 2023-09-23
CONTRIBUTING.rst has broken links when viewed on GitHub


pushfoo:
## Bug Report

tl;dr `:ref:`s and other directives don't  seem to render in a GitHub view of an `.rst` file, even for targets within the same file
 
## System Info

```
Version 117.0.5938.62 (Official Build) built on Debian 11.7, running on Debian 11.6 (64-bit)
```

### Actual behavior:

The `Testing` link in [After making changes](https://github.com/pythonarcade/arcade/blob/development/CONTRIBUTING.rst#after-making-changes) does not link anywhere and does not render correctly:
![image](https://github.com/pythonarcade/arcade/assets/36696816/640611b3-5782-4cfe-bd25-954467d7c41e)

### Expected behavior:

The `Testing` link in [After making changes](https://github.com/pythonarcade/arcade/blob/development/CONTRIBUTING.rst#after-making-changes):
1. Works when clicked
2. (Ideally) renders cleanly

### Steps to reproduce/example code:

1. Go to  [After making changes](https://github.com/pythonarcade/arcade/blob/development/CONTRIBUTING.rst#after-making-changes)
2. Observe the link
3. Attempt to click it

Correcting the capitalization does not change this either. It appears GitHub cannot render `:ref:` targets.

### Root causes

There seems to be limited post-processing of directives in GitHub views

### Implications

This may block #1302 and affect #1886, both of which I was looking at when I noticed this.

### Possible ways to fix it

1. Use a script which renders CONTRIBUTING.rst as markdown
2. Go back to old approach: the abbreviated model + links to the web site

I'm not in favor of clever scripts because those will probably break at some point.

cspotcode:
> Go back to old approach: the abbreviated model + links to the web site

It can be more than abbreviated, it can be a single link to the website with zero additional information.

pushfoo:
I'm ok with that option, but might make it more or less a toctree dump for the relevant pages. @pvcraven any objections? I don't see the value of keeping this page seperate then reaching back through the hierarchy to include it.

einarf:
Can't we just change README and CONTRIBUTING into markdown files? They don't really need the complexity rst has to offer in the first place.

Cleptomania:
Actually the CONTRIBUTING file is included directly in the doc build, though it would be trivial to add in MyST Parser extension to sphinx to allow us to just include the Markdown(and maybe use more markdown in the future?)

einarf:
[...]

PR #1939, opened 2023-12-12
Convert README and CONTRIBUTING to Markdown

Cleptomania:
This converts the README and CONTRIBUTING files in the Git repo to use Markdown as discussed in #1910. These are only displayed on the GitHub repo(and the PyPI page) and are not part of the Sphinx docs build, so there's no real reason for them to stay RST, we don't have a need for the extra complexity it adds to them.


Cleptomania:
Upon further inspection this actually does get used in the docs directly, I didn't realize we were including this file into the doc build. Need to think about that more.

einarf:
Should be easy enough to fix. We can just link to it from the docs. Easy cleanup (TM)

Cleptomania:
Yeah we should be able to just link to the GitHub page for it in the docs, anyone contributing is gonna have to use GitHub anyways so it's not that wild of a re-direct. I'll push a change to docs page to do this soon.

Cleptomania:
This is probably ready to go now, but I'll wait for at least @einarf or @pvcraven to give an okay on merging that we're happy with the changes here. Basically removes most any documentation about contributing from the actual docs, and instead directs to the GitHub CONTRIBUTING file, which is now written in Markdown.

Gives us a single source of truth for the contributing guidelines without having to be dependent on the somewhat fragile include of the CONTRIBUTING file into Sphinx(and the incompatibilities between GitHub and Sphinx's rendering of said file).

PR #1958, opened 2024-01-23
Atlas track all textures

einarf:
Fixed various issues related to texture tracking in atlases.

### DIFF
--- a/arcade/experimental/atlas_load_save.py
+++ b/arcade/experimental/atlas_load_save.py
@@ -1,112 +1,112 @@
-"""
-Quick and dirty atlas load/save testing.
-Loading and saving atlases are not officially supported.
-This is simply an experiment.
+# """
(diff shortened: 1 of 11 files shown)


einarf:
Remaining:
* Don't raise exception when removing a texture that is no longer in the atlas. This is too error prone.
* Ensure atlas ref is injected into the texture when added every time

einarf:
* Might need an internal `unique_texture` counter as well so we don't remove one that has a duplicate
* In addition we need to wrap the `add()` method so we don't increment the ref coutner multiple times when `add()` causes a resize or rebuild
* The above point can also be solved by simplifying the `clear()` method. Splitting it into smaller private methods might be a good idea. The problem is that `resize()` calls clear with all param flags as false and it's unclear what that even do. We end up wiping texture list, but keep the image ref count.

Issue #1959, opened 2024-01-28
Add ability to change the ladder list in platform


FriendlyGecko:
### What should be added/changed?
A method to add a ladder list post initialization of the physics engine.

### What would it help with?
Situations where a ladder does not yet exist but will.

PR #1960, opened 2024-01-28
Add ability to update the platform physics engine's ladder list

FriendlyGecko:
Added an update ladder list method to the platformer engine. This allows the player to add a ladder list outside of the init method.

See #1959


arcade/physics_engines.py
```diff
@@ -313,6 +313,10 @@ def __init__(self,
         self.allowed_jumps: int = 1
         self.allow_multi_jump: bool = False
 
+    def update_ladders(self, ladders: Union[SpriteList, Iterable[SpriteList]]):
+        """ Update the current ladder list. """
+        self.ladders = [ladders] if isinstance(ladders, SpriteList) else list(ladders)
+
```
pvcraven, line 331: Hm, should this instead be done as a property?

FriendlyGecko: Let me research how to do that and why. I am personally only familiar with methods and a much lesser extent properties.

FriendlyGecko: Ah ok, so it would make it more intuitive for users who go to set the ladder list. 


    

    def _get_ladders(self):
            return self._ladders

    def _set_ladders(self, ladders: Union[SpriteList, Iterable[SpriteList]]):
        if ladders:
            self._ladders = [ladders] if isinstance(ladders, SpriteList) else list(ladders)
        else:
            self._ladders = None
    
    def _del_ladders(self):
        self._ladders = None

    ladders = property(
        fget=_get_ladders,
        fset=_set_ladders,
        fdel=_del_ladders,
        doc="The ladder list of the physics engine."
    )

What are your thoughts on something more like this?

FriendlyGecko: Or something more concise with decorators like below:

```
@property
    def ladders(self):
        return self._ladders

@ladders.setter
def ladders(self, ladders: Union[SpriteList, Iterable[SpriteList]]):
    if ladders:
        self._ladders = [ladders] if isinstance(ladders, SpriteList) else list(ladders)
    else:
        self._ladders = None

@ladders.deleter
def ladders(self):
    self._ladders = None
```

Issue #1848, opened 2023-08-04
Fix/Improve PhysicsEnginePlatformer Jumping System


Cleptomania:
The PhysicsEnginePlatformer class has the beginnings of a jumping system. In the platformer tutorial we have previously largely handled the jumping logic within the game, with only the use of the `can_jump` function to tell us if the player is currently on a ground or a ladder.

There is some logic in place for multi-jumps, but it is currently not working(the jump checks don't take it into account at all), and it also needs some improvement, namely an interval/timer between multi-jumps.

This stuff all appears in intellisense when users are working on games, so it would be good to get it actually usable, because newer users may be spending a lot of time trying to figure out why something doesn't work.

a-nthonyy:
Hi! I'd love to work on this issue as my first ever open-source contribution. Let me know if there’s anything I should know before I get started. Thanks!

Cleptomania:
Feel free to pick it up, generally just a look through the contributing docs should get you started. If you have any specific questions on getting started or as you go I'd recommend joining the discord

PR #1961, opened 2024-01-28
Added multi-jumps to jump method/can_jump interval test

FriendlyGecko:
This is an attempt to start resolving #1848 .

As of current this only has added multi-jump capabilities to the jump method. I added in the ability to set various types of multi-jump: set, additive, and limited (additive to a point). 

I also added a jump interval timer delay based on ticks of the physics engine being updated.


FriendlyGecko:
I believe that there is more work to be done on this. Specifically making it so players can jump off ladders. In the meantime, I would like some critiques on how it is currently coded so that we can ensure that this is inline with the rest of the library.

pushfoo:
@FriendlyGecko are you still interested in working on this?

FriendlyGecko:
Yes, I am.

However, I am unsure if it is ready/what do we need in order to determine if it is ready.

Items I am considering
- [x] Singular jump/multijump capable from one command
- [x] Multijump with configurable delay
- [x] Multiple styles of multijump
- [x] Capability to jump from ladder (without jumps instantly resetting)

Is there anything else you can think of that would be necessary?

FriendlyGecko:
This request also adds an is_on_ground test which we can use to differentiate cases where we want altering behavior based on if the player is on a ladder vs the ground.

NOTE: In order to test this properly in regards to jumping while on a ladder, it is imperative that you alter the default jump_delay time to a number greater than 0 (10 is a good number). Otherwise player will launch like a rocket up the ladder with infinite jumps. The delay functions to limit the player from instantaneously jumping repeatedly and prevents them from just jumping up the entire ladder unchecked. 

To further support this, I believe that if we want to merge this, then we need to update the test AND change the default delay_time to 10.

arcade/physics_engines.py
```diff
@@ -378,17 +381,17 @@ def is_on_ladder(self):
         # Check for touching a ladder
         if self.ladders:
             hit_list = check_for_collision_with_lists(self.player_sprite, self.ladders)
-            if len(hit_list) > 0:
+            if len(hit_list) > 0 and self.jump_ticks >= self.jump_delay:
```
pvcraven: A bit more 'pythonic' way of doing this, is just `if hitlist and`

arcade/physics_engines.py
```diff
@@ -432,17 +450,51 @@ def disable_multi_jump(self):
...
+        to their air jump speed. """
+        if self.can_jump():
+            # Air Jump logic
+            if self.jumps_since_ground > 0:
+                # This checks if air_jump_velocity is set. If not it will default to the velocity for all air jumps.
+                if air_jump_velocity:
+                    air_jump = air_jump_velocity
+                else:
+                    air_jump = velocity
+                if air_jump_style == "additive":
```
pvcraven, line 469: Would it be better to use a constant/enum for this?>

pvcraven (changes requested):
Any chance to add some unit tests for this? Documentation? Examples?

FriendlyGecko:
> Any chance to add some unit tests for this? Documentation? Examples?

Perhaps not in near-near future, but early July my schedule should be a little more open.

By documentation do you mean within the code or within the rest of the project's documentation? And by examples do you mean like example code showing/explaining how one could use it? If so, then absolutely.

arcade/physics_engines.py
FriendlyGecko: Looking into it.

FriendlyGecko: Does the current version of the code better fulfill your intent?

einarf:
After some discussion: This feature can easily be implemented on the user side. I don't think we want more complexity in these simple physicsengines. We will probably brush them up a bit in 3.x. At that point we could look at something like this.

PR #1964, opened 2024-02-03
Support zoom in Camera

eruvanos:
This is a fallback to fix the current camera implementation, can be dropped if #1965 is merged.


einarf:
Superseded by the new camera

PR #1965, opened 2024-02-04
Camera overhaul simple

DragonMoffon:
This PR includes an updated Orthographic Camera and Camera2D to replace the old SimpleCamera provided with <3.0 Arcade.

There is a more detailed Draft PR with extra components (#1905).


eruvanos:
Wow, ok, that are many modified files :/
There are also changes in non camera code files.

DragonMoffon:
Yeah, there were a bunch of files that used the old camera code that had to be changed. If any changes aren´t related to the camera, I will reverse them
.

pushfoo:
My first read through only found a few spots with non-Camera changes. Most seemed to be opportunistic whitespace changes. The majority of the changes seemed to be in doc which uses cameras.

einarf:
Went through the hell fire and resolved all conflicts. 

Some notes:
* Platformer tutorial must have the camera applied again (it was completely changed. Even renamed)
* The TextureAtlas did NOT like the camera in `render_into`, so I had to yank that out setting viewport and projection manually

Still need to go through the code in more detail and wrap my head around things.

einarf:
This PR also touching things not related to the camera itself. It would be better to move the spritelist changes into separate PR. I will yank that out for now since it conflicts with other plans. Taking in string arguments that translates into some internal magic behavior is something we want to avoid as much as possible. Instead we want the "renderpass" object thingy.

Issue #1838, opened 2023-07-18
Make Color constants usable without affecting alpha


bunny-therapist:
### What should be added/changed?

Add an "rgb" method to arcade.types.Color which return a 3-tuple (type "RGB") based on the color, i.e., it leaves out the alpha.

### What would it help with?

In arcade 2, Sprite.alpha was separate from Sprite.color, so one could use the constants in arcade.color to set the color without changing the alpha of the sprite. For example, `self.color = arcade.color.WHITE` would not affect the alpha of the Sprite. Now it does, which means makes the color constants much less usable. With an "rgb" property, one could do `self.color = arcade.color.WHITE.rgb` which would not affect alpha since the color setter checks the length of the input.

One can currently do the same, of course, by either inputting the color numbers directly as a tuple (but then what are the color constants even good for?) or by grabbing the constant and accessing its r, g, and b properties (but that is very messy).

```
Arcade 3.0.0.dev23
------------------
vendor: Intel
renderer: Intel(R) UHD Graphics
version: (3, 3)
python: 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
platform: win32
pyglet version: 2.0.7
PIL version: 9.4.0
```

bunny-therapist:
Alternatively, the color constants could simply all be of type RGB, instead of type Color.

pushfoo:
> With an "rgb" property, one could do `self.color = arcade.color.WHITE.rgb` which would not affect alpha since the color setter checks the length of the input.

tl;dr: We clearly need to document this better, but `self.color = arcade.color.WHITE[:3]` already does this.

Example:
```python
import arcade
from arcade.color import RED, GREEN, BLUE


class ColorSlicingExample(arcade.Window):

   def __init__(self, width: int = 200, height: int = 200):
       super().__init__(width=width, height=height)
       self.sprites = arcade.SpriteList()

        # Use transparent gray as the initial color
       initial_color = (127, ) * 4

       # Create a line of sprites which use the RGB values of from color constants.
       for index, color_instance in enumerate((RED, BLUE, GREEN)):
           sprite = arcade.SpriteCircle(10, initial_color)
           sprite.position = 50 + (index * 50), height / 2

           # Use only the RGB components from each color constant.
           # Color is a subclass of tuple and supports the same operations, including slicing.
           sprite.color = color_instance[:3]

           self.sprites.append(sprite)


   def on_draw(self):
       self.clear()
[...]

Issue #1953, opened 2024-01-16
Make arcade 2.x python 3.12 compatible


einarf:
Bumping some depedencies with minor code changes is the only thing needed to make this happen.

See #1950

professorwade:
What is the timeframe for this change?

einarf:
> What is the timeframe for this change?

I'm looking into if this is even possible at the moment

einarf:
There seems to by pyglet issues related to text with pygket 2.0dev23 meaning making a 2.x verison with python 12 support is a bit more challenging than just bumping Pillow and Pymunk.

```
Traceback (most recent call last):
  File "C:\Users\efors\projects\moderngl\arcade\arcade\examples\sprite_bullets.py", line 177, in <module>
    main()
  File "C:\Users\efors\projects\moderngl\arcade\arcade\examples\sprite_bullets.py", line 173, in main
    arcade.run()
  File "\arcade\arcade\window_commands.py", line 323, in run
    pyglet.app.run()
  File "site-packages\pyglet\app\__init__.py", line 107, in run
    event_loop.run(interval)
  File "site-packages\pyglet\app\base.py", line 184, in run
    timeout = self.idle()
              ^^^^^^^^^^^
  File "site-packages\pyglet\app\base.py", line 245, in idle
    self.clock.call_scheduled_functions(dt)
  File "site-packages\pyglet\clock.py", line 277, in call_scheduled_functions
    item.func(now - item.last_ts, *item.args, **item.kwargs)
  File "site-packages\pyglet\app\base.py", line 154, in _redraw_windows
    window.dispatch_event('on_draw')
  File "site-packages\pyglet\window\__init__.py", line 1361, in dispatch_event
    super().dispatch_event(*args)
  File "site-packages\pyglet\event.py", line 422, in dispatch_event
    if getattr(self, event_type)(*args):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "arcade\examples\sprite_bullets.py", line 110, in on_draw
    arcade.draw_text(f"Score: {self.score}", 10, 20, arcade.color.WHITE, 14)
  File "arcade\text_pyglet.py", line 724, in draw_text
    label = pyglet.text.Label(
            ^^^^^^^^^^^^^^^^^^
  File "site-packages\pyglet\text\__init__.py", line 474, in __init__
    super().__init__(doc, x, y, width, height, anchor_x, anchor_y, multiline, dpi, batch, group, rotation)
[...]

PR #1970, opened 2024-02-11
Updated the color setter so that it does not change the current alpha

FriendlyGecko:
Will allow the closure of #1838.

I just swapped the last value self.color[3] with self._alpha which will the alpha getter to maintain the current.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -346,7 +346,16 @@ class BasicSprite:
 
     @color.setter
     def color(self, color: RGBA255):
-        if len(color) == 4:
+        if len(color) == 3 or isinstance(color, Color):
+            if (
+                    self._color[0] == color[0]
+                    and self._color[1] == color[1]
+                    and self._color[2] == color[2]
+            ):
+                return
+            self._color = Color(color[0], color[1], color[2], self.alpha)
+
+        elif len(color) == 4:
             if (
                 self._color[0] == color[0]
                 and self._color[1] == color[1]
@@ -356,14 +365,6 @@ class BasicSprite:
                 return
             self._color = Color.from_iterable(color)
 
-        elif len(color) == 3:
-            if (
-                self._color[0] == color[0]
-                and self._color[1] == color[1]
-                and self._color[2] == color[2]
-            ):
-                return
-            self._color = Color(color[0], color[1], color[2], self._alpha)
         else:
             raise ValueError("Color must be three or four ints from 0-255")
 
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Iterable, List, TypeVar, Any
+from typing import TYPE_CHECKING, Iterable, List, TypeVar
 
 import arcade
-from arcade.types import Point, Color, RGBA255, PointList
+from arcade.types import Point, Color, RGBA255, PointList, RGB
 from arcade.color import BLACK
 from arcade.hitbox import HitBox
 from arcade.texture import Texture
@@ -49,7 +49,7 @@ class BasicSprite:
         scale: float = 1.0,
         center_x: float = 0,
         center_y: float = 0,
-        **kwargs: Any,
+        **kwargs,
     ) -> None:
         self._position = (center_x, center_y)
         self._depth = 0.0
@@ -58,6 +58,7 @@ class BasicSprite:
         self._height = texture.height * scale
         self._scale = scale, scale
         self._color: Color = Color(255, 255, 255, 255)
+        self._rgb: RGB = RGB(255, 255, 255)
         self.sprite_lists: List["SpriteList"] = []
 
         # Core properties we don't use, but spritelist expects it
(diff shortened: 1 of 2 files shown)


FriendlyGecko:
Recognized that that wasn't enough to solve the open issue so I modified the code to check for use of the arcade.Color type. Now it should be ready for testing/merge.

TO CLARIFY: I only modified the len = 3 check. Not len = 4, so users can still set the alpha through this if they wish.

einarf:
We probably should have separate `.rgb` and `.alpha` properties and let the `color` one just take whatever length is passed in. This is more in line with arcade 2.6 and solves two issues.

FriendlyGecko:
That makes sense, it is a little late for me now, but I will try to draft something like this when I get up.

FriendlyGecko:
I created a workable draft, larger part will be cleaning up the RGB class which for the moment is just a copy of the Color class. As of right now, it is functional just not clean.

arcade/sprite/base.py
```diff
@@ -342,28 +362,31 @@ def color(self) -> Color:
...
-                and self._color[2] == color[2]
+                self._rgb[0] == color[0]
+                and self._rgb[1] == color[1]
+                and self._rgb[2] == color[2]
+                and self._rgb[3] == color[3]
             ):
                 return
-            self._color = Color(color[0], color[1], color[2], self._color[3])
+            self._rgb = color
+            self._color = Color.from_iterable(color)
```
pushfoo, line 388: This is cleaner, but it's worth doing benchmarks to see if it's slower. The reason pyglet code and the `from_iterable` method use the strange-looking unpacking method is that it's faster.

arcade/types.py
```diff
@@ -450,11 +534,11 @@ class TiledObject(NamedTuple):
     type: Optional[str] = None
 
 
-if sys.version_info >= (3, 12):
```
pushfoo, line 453: Why remove this? My concerns:
* We don't yet have 3.12 compatibility merged (see #1953)
* We aim to support the lowest non-EOL version unless forced to drop it

pushfoo (changes requested):
I know this is marked as a draft, but I'm not sure I understand all the changes intended or mentioned. I've commented on two specific places in the source.

> larger part will be cleaning up the RGB class

einarf's comment above seems relevant here:
> We probably should have separate `.rgb` and `.alpha` properties and let the color one just take whatever length is passed in.

I'm not yet sure having an `RGB` class is worth it at the moment. The main benefits I see are:
1. Cleaner RGB-only behavior
2. In theory, adding a `arcade.color.rgb` submodule which only has RGB colors

However, that has risks and costs:
1. Inheritance can slow things down, so copy and pasting code is a useful speed optimization
2. Copy and pasting creates riskjs of desyncing things which should have the same API behavior
3. Preventing that risk requires more complicated unit tests

These downsides are best dempnstrated by pyglet's shape, label, and sprite abstractions. The label and shape issues affect arcade since we depend on those classes. In C or C++, we could rely on the compiler to perform [inlining](https://en.cppreference.com/w/cpp/language/inline), but Python doesn't have that option. For pyglet, we have to (eventually) rework tests to be better at enforcing API consistency in abstractions if we don't want to sacrifice execution speed.

arcade/types.py
FriendlyGecko: This wasn't intentional, I still have to clean it up before it is ready.

arcade/sprite/base.py
FriendlyGecko: "I'm not sure I understand all the changes intended or mentioned."

The intent is to make it so RGB and alpha are used together to create color and by default not changing the alpha value of a sprite unless specifically called on.

The idea being that RGB can more easily be manipulated separately.

pushfoo:
@FriendlyGecko Are you still interested in working on this?

### TL;DR

1. An `.rgb` property with getter+setter is definitely good
2. We should defer an `RGB` class till a later PR.

#### How?

This is my opinion, but I'd take the following approach:

1. Make a backup branch with these changes and push it to your fork
3. Switch back to this PR's branch
4. Revert the following:
    * Changes to `BasicSprite.color`
    * The `RGB` class (We'll wait till later, see bottom of comment)

4. Add a `Color.rgb` property which returns `self[:3]`
5. Modify `BasicSprite.rgb`'s setter to use an inlined style like #2029
6. Strongly consider reverting the `_rgb` slot

### Why revert `BasicSprite.color` changes?

TL;DR: Computes rarely used data + has a merge conflict

We noticed visibility behavior is broken at the moment. There's already a PR to fix it over at #2029. The implementation there seems like good reference for a `BasicSprite.rgb` property since it seems to make efficiency improvements to the color setter. However, they also conflict with the current implementation of this branch.

Calculating an rgb value each time we set also doesn't seem worthwhile if:
* Color property writes are common
* Color property reads are rare
* The GPU is sent RGBA each time anyway

### Why revert `arcade.types.RGB` for now?

TL;DR: Not a hard no forever, but we need to ship 3.0 + there are ongoing perf questions about tuple subclassing

#### New Features Discouraged until 3.1

To be clear, this isn't a hard no forever. However, the current priority is shipping Arcade 3.0. New features are somewhat discouraged for the moment unless:

* It's clear they won't break things somehow
* They need to be done now to prevent huge future pain

In this case, expanding the already-overfilled `arcade.types` module:
*  risks breaking things
* doesn't seem necessary to prevent future issues

Issues with these are why #1772 is on hold.

### Why revert the `_rgb` slot addition?

TL;DR: Duplicates rarely read data

If we read `.rgb` from sprites rarely, it's not worth:
* Duplicating data already found in `BasicSprite.color`
* Recomputing duplicated value with an expensive function call
* The complexity of a lazily updated property as an alternative

### Perf Questions about Tuples / Etc

Another reason to wait is the perf profiling the pyglet project has been doing. They've been benchmarking variations on implementing vector types somewhat similar to `Color`.

Although their findings need further investigation, they suggest Arcade's colors might need a closer look. This applies to both the current `Color` and any future color types.

### When do we add `RGB` as a type?

TL;DR: After we finish things critical to 3.0

Right now, the `RGB` class conflicts with the `RGB` type alias we have. There might be a way to change these aliases and their names to be nicer, but we first need to finish addressing a lot of other things.

1. #1965
2. A big doc build system refactor to help with the camera PR
4. Various GL issues with blend state and texture managers
5. Various doc issues
6. Time / update issues
7. Input problems we should fix now
8. pyglet math type updates

I'm putting some thought into our nomenclature for colors again. There might be a way to use your concept. However, it might require cleaning up types. You can see that started in #2030, but nothing uses that alias yet, so it doesn't break anything. Avoiding major unintended breakage + the big todo list above both take priority.

FriendlyGecko:
Marking as closed due to features being implemented in #2060

PR #1978, opened 2024-02-20
GUI: add gui.experimental and move experimental components here

eruvanos:
Add support for package level doc string in quickindex

### DIFF
--- a/arcade/gui/examples/__init__.py
+++ b/arcade/gui/examples/__init__.py
@@ -1,0 +1,7 @@
+"""
+This package contains GUI specific examples.
+
+These show explicit GUI examples, and are not part of the main examples.
+
+Examples which begin with "exp_" use experimental components from the `arcade.gui.experimental` module.
+"""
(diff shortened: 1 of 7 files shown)


util/update_quick_index.py
```diff
@@ -241,7 +245,15 @@ def process_directory(directory: Path, quick_index_file):
...
-            api_file.write(f"{underline}\n\n")
+            api_file.write(f"{underline}\n")
+
+            # Use automodule to add package level doc string
+            if api_file_name in include_module_docs:
+                api_file.write(
+                    f".. automodule:: {include_module_docs[api_file_name]}\n"
+                )
+                api_file.write("    :no-index:\n")
+            api_file.write(f"\n")
```
eruvanos: @pvcraven this would add module doc_string between title and the first members, this is for now only activated for `arcade.gui.experimental`.

the `no-index` prevents double documented code which is listed within the module via `__all__`

eruvanos: Removed from this PR, we start without that.

PR #1987, opened 2024-02-24
Expose batch drawing text example in the web doc

### LINKED ISSUE #1571: Add batch drawing Text example to docs

The examples page currently shows a "Slow But Easy" and a "Fast" text drawing example. These use `arcade.draw_text()` and `arcade.Text()` respectively. We have an example code for batch drawing `arcade.Text()` objects, but it is not exposed in the docs anywhere. This example needs promoted as the fastest way to draw text, as every draw call on individual `arcade.Text()` objects will add up really fast.

This example could maybe use some explanation of Pyglet batches as that is what it uses to do the batch draw. Perhaps a simpler example of using a pyglet batch or linking to the pyglet documentation for that is in order alongside this.

pushfoo:
Closes #1571 (marked as 3.0 mandatory)

### Changes

* Add the batch drawing example to the example gallery in the docs
* Clean up the example's docstrings and constant names
* Update the `drawing_text` and `drawing_text_objects` pages with cross-references

### DIFF
--- a/doc/example_code/how_to_examples/drawing_text_objects.rst
+++ b/doc/example_code/how_to_examples/drawing_text_objects.rst
@@ -2,16 +2,19 @@
 
 .. _drawing_text_objects:
 
-Fast Text Drawing
-=================
+Better Text Drawing with Text Objects
+=====================================
 
 .. image:: drawing_text_objects.png
     :width: 500px
     :align: center
     :alt: Screenshot of drawing with text objects
 
-This example shows how to draw text using "text objects." It is the fastest
-way to draw text. If you are looking for something simpler, see :ref:`drawing_text`.
+This example shows how to draw text using "text objects." It can be many
+times faster than using :py:func:`arcade.draw_text`.
+
+* If you want to be as efficient as possible, see :ref:`drawing_text_objects_batch`.
+* If you are looking for something simpler, see :ref:`drawing_text`.
 
 .. literalinclude:: ../../../arcade/examples/drawing_text_objects.py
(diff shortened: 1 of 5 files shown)


doc/example_code/how_to_examples/drawing_text_objects.rst
```diff
@@ -2,16 +2,19 @@
...
 
 .. image:: drawing_text_objects.png
     :width: 500px
     :align: center
     :alt: Screenshot of drawing with text objects
 
-This example shows how to draw text using "text objects." It is the fastest
-way to draw text. If you are looking for something simpler, see :ref:`drawing_text`.
+This example shows how to draw text using "text objects." It can be many
+times faster than using :py:func:`arcade.draw_text`.
```
einarf, line 14: I would emphasize the performance differences here.

* `draw_text`: 10s of objects per fram
* `Text` object: 100s of objects per frame
* `Text` object with batching: 1000s of objects per frame

The performance difference is crazy. Especially when they are more or less static.

pvcraven (approved):
I like einarf's suggested improvements. Regardless, I think this is a good PR.

doc/example_code/how_to_examples/drawing_text_objects.rst
pushfoo: For brevity's sake, it seems appropriate to add a programming guide section on text to cover this. We need that anyway to explain issue with fonts.

einarf:
Let's just merge this!

PR #1993, opened 2024-02-25
remove pause

DigiDuncan:
This is super unnecessary and nothing uses it (now). Even before this PR, it was used once. So, I'm counting this as "weird cruft that should be cut in 3.0." If it's a teaching concern, I recommend to teach them to use `time.sleep()`.

## Changes
- Removes `arcade.pause()`

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -76,7 +76,6 @@ from .window_commands import exit
 from .window_commands import finish_render
 from .window_commands import get_display_size
 from .window_commands import get_window
-from .window_commands import pause
 from .window_commands import schedule
 from .window_commands import run
 from .window_commands import set_background_color
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -351,7 +351,6 @@ __all__ = [
     'make_soft_circle_texture',
     'make_soft_square_texture',
     'open_window',
-    'pause',
     'print_timings',
     'play_sound',
     'read_tmx',
--- a/arcade/window_commands.py
+++ b/arcade/window_commands.py
@@ -27,7 +27,6 @@ _window: Optional["Window"] = None
 
 __all__ = [
     "get_display_size",
-    "pause",
     "get_window",
     "set_window",
     "set_viewport",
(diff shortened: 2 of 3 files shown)


einarf:
I don't mind removing this one, but if it's referenced in teaching material we migh just need to keep it. Outside of that I don't mind removing it. Especially if it's mainly a thing used internally in tests. 

@pvcraven Any input on this one?

DigiDuncan:
> I don't mind removing this one, but if it's referenced in teaching material we migh just need to keep it. [...]

From what I can tell, the docs don't reference `arcade.pause()` at any point. I don't have access to Paul's textbooks, obviously, but again, those textbooks would be outdated post-3.0 anyway and would require updating.

pushfoo:
TL;DR: We might want to keep the method but rethink what it does

> I don't have access to Paul's textbooks

I think we do: https://learn.arcade.academy/en/latest/

The [git history](https://github.com/pvcraven/arcade_book/commits/master/) shows Paul C's been updating it, but it doesn't seem to reference the method.

### What we have now

`arcade.pause()`'s current behavior seems a little counter-intuitive: it pauses all execution, not just gameplay. For just getting started, this sort of makes sense.

### What should we have?

Pause behavior in real-time games usually takes the following approach:

1. Game state ceases updating
2. Animation and sound effect playback may pause
3. The UI remains interactive, but may change to a menu / pause screen
4. Music and ambiance may keep playing

### How could we support that?

1. Have `arcade.pause()` call a `Window.pause()` method
2. The `Window`'s handler can respond directly, or defer to a `View`'s implementation
3. Users can override this behavior or defer to other handlers which do things like pause a subset of event scheduling

pvcraven (approved):
I'm ok removing this function.

pushfoo:
Removing the top-level function works too. It doesn't stop any of that other implementation from happening later. :+1:

DigiDuncan:
I really like the idea of an `on_pause` or something like it on `Window`/`View`. Newcomers having to separate their game-loop code into a big `if self.paused` check can be annoying and counter-intuitive, even for experience developers. Something for the future!

pushfoo:
Those ideas tie into a larger rework of events / input that we don't have time for rn. This PR comment thread also isn't the place to discuss it imo. We should merge this.

einarf:
Today it's `pause()`'s funeral. Bring flowers.

PR #1998, opened 2024-02-26
Swizzle support for colors

DigiDuncan:
We talked about this in the Discord once. And why not, right? It's a cheap function to write.

Could probably have docs or something!

Takes a string of r, g, b, and a's and returns a tuple of those properties in order.

It's an [OpenGL thing.](https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Swizzling)

### DIFF
--- a/arcade/types.py
+++ b/arcade/types.py
@@ -421,7 +421,7 @@ class Color(RGBA255):
         ret = []
         for c in swizzle_string.lower():
             if c not in 'rgba':
-                raise Exception("Swizzle string must only contain characters in [RGBArgba].")
+                raise ValueError(f"Swizzle string must only contain characters in [RGBArgba], not {c}.")
             ret.append(getattr(self, c))
         return tuple(ret)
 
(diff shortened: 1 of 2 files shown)


arcade/types.py
```diff
@@ -419,6 +417,14 @@ def random(
 
         return cls(r, g, b, a)
 
+    def swizzle(self, swizzle_string: str) -> Tuple[int, ...]:
+        ret = []
+        for c in swizzle_string.lower():
+            if c not in 'rgba':
+                raise Exception("Swizzle string must only contain characters in [RGBArgba].")
```
pushfoo:
```suggestion
                raise ValueError("Swizzle string must only contain characters in [RGBArgba].")
```
I'd also consider outputting the offending value.

pushfoo (changes requested):
This is almost ready. I think the current implementation is faster and easier than `__getattribute__` hacks, even if `color_instance.brga` would be cleaner code. 

Changes I'd like:
1. Use `ValueError` insteaad of `Exception`
2. Consider mentioning the offending value in the exception
3. Add tests

arcade/types.py
DigiDuncan: Done the error improvement, tests to come!

DigiDuncan: Thanks, @pushfoo, for the tests!

pushfoo (approved):
This looks good now.

Issue #933, opened 2021-08-31
AssertionError: When the parameters 'multiline' and 'wrap_lines' are True, the parameter 'width' must be a number


alderven:
Run this code:
```
import arcade

arcade.open_window(800, 600, 'Hello')
arcade.draw_text(text='Привет!', start_x=400, start_y=300, font_size=36, align='center')
arcade.run()
```

Getting error:
```
Traceback (most recent call last):
  File "C:/draw_text.py", line 4, in <module>
    arcade.draw_text(text='Привет!', start_x=400, start_y=300, font_size=36, align='center')
  File "C:\Program Files\Python38\lib\site-packages\arcade\text_pyglet.py", line 138, in draw_text
    label = pyglet.text.Label(
  File "C:\Program Files\Python38\lib\site-packages\pyglet\text\__init__.py", line 447, in __init__
    super(Label, self).__init__(document, x, y, width, height,
  File "C:\Program Files\Python38\lib\site-packages\pyglet\text\__init__.py", line 268, in __init__
    super(DocumentLabel, self).__init__(document,
  File "C:\Program Files\Python38\lib\site-packages\pyglet\text\layout.py", line 889, in __init__
    self._wrap_lines_invariant()
  File "C:\Program Files\Python38\lib\site-packages\pyglet\text\layout.py", line 1155, in _wrap_lines_invariant
    assert not self._wrap_lines or self._width, \
AssertionError: When the parameters 'multiline' and 'wrap_lines' are True, the parameter 'width' must be a number.

Process finished with exit code 1
```

**Note:** Works OK on Arcade 2.5.7 or if I remove `align='center'`

Python: 3.8
Arcade: 2.6.0
OS: Win 10

caffeinepills:
To align something, you need to add a width to align it against. 

In this case an anchor is probably more in line with what you were wanting.  Try adding `anchor_x='center'`.

einarf:
Hmm. Not sure if it's possible to make that compatible between versions.

pvcraven:
The 2.5.7 API can get the width of the text and auto-center it based on that field width. The new 2.6 API does not have that ability. The width has to be specified.

I don't think there's an ability to get the width of the field?

caffeinepills:
[...]

PR #2000, opened 2024-02-27
Clean up arcade.Text

eruvanos:
Change arcade.Text and draw_text to match general argument names,
remove interference between align and multiline.

Fix:
- #933 
- #1839 
- https://github.com/pyglet/pyglet/issues/915

Covers PR:
- #1864

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -202,6 +202,9 @@ class Text:
         if align not in ("left", "center", "right"):
             raise ValueError("The 'align' parameter must be equal to 'left', 'right', or 'center'.")
 
+        if multiline and width == 0:
+            raise ValueError("The 'width' parameter must be set when 'multiline' is True.")
+
         adjusted_font = _attempt_font_name_resolution(font_name)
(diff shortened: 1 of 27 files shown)


eruvanos:
I will remove the additional indent in the method signatures.

arcade/text.py
```diff
@@ -583,18 +580,18 @@ def position(self, point: Union[Point, Point3]):
...
+        font_size: float = 12,
+        width: int = 0,
+        align: str = "left",
+        font_name: FontNameOrNames = ("calibri", "arial"),
+        bold: bool = False,
+        italic: bool = False,
+        anchor_x: str = "left",
+        multiline: bool = False,
+        texture_atlas: Optional[arcade.TextureAtlas] = None,
+        background_color: Optional[RGBA255] = None,
```
einarf: Knitpicking here, but these should probably not be indented. We've kind of used someting close to the black formatter everywhere else.

eruvanos (commented):
Fixed wrong indent formats

PR #2002, opened 2024-02-28
Add pre-commit setup

eruvanos:
This adds pre-commit hooks for
- ruff
- ruff formater
- mypy
- pyright
- black

These can be optionally enabled following the contribution guide.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,32 @@
+default_language_version:
+    python: python3.9
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    # Ruff version.
+    rev: 'v0.2.2'
+    hooks:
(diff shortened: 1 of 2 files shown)


eruvanos:
Following the [ruff formater docs](https://docs.astral.sh/ruff/formatter/) we could even drop the black hook.

pushfoo:
**TL;DR:** auto-formatters may be ok if we limit them, but what about doc & redundancy?

### Auto-formatters

I'm glad this is an optional setup step. There are cases where ignoring PEP-8 improves readability, especially constant blocks.

Both [black](https://stackoverflow.com/questions/58584413/black-formatter-ignore-specific-multi-line-code) and [ruff's formatter](https://docs.astral.sh/ruff/formatter/#format-suppression) support `# fmt: off` + `# fmt: on` to switch formatting on and off inside a file.

### What about doc?

Should running doc build before commit be part of this too?

### Redundancy?

We currently have `./make.py` as a replacement for a `Makefile`. I use this constantly. It might be an extra step of indirection, but would it make sense to call it to maintain single source of truth?

Example:

```shell
# Runs ruff, mypy, and pyright. We might want to split the latter into strict-lint
./make.py lint
```

eruvanos:
We skip it (for now) to reduce complexity for contributions.

PR #2007, opened 2024-03-03
Fix color tables in docs 1878

### LINKED ISSUE #1429: Add gray checkerboard background under color swatches
Update the color table generation with a two-tone gray checkerboard underneath the transparent or partially opaque color swatches.

I wasn't able to get this working in #1428 due to alignment issues when trying to use Z-indexed divs.

### What would it help with?
`TRANSPARENT_BLACK` is currently rendered as a black border around a rectangle filled with the current background color of the page. This may be confusing for beginners.

Implementing this change would make the transparency of all colors easier to understand since the checkerboard pattern is an intuitive indicator for the absence or transparency of pixel data in an RGBA image.

sabadam32:
Fixes #1429


pushfoo:
One tiny question: what's the source of the checker grid image? Did you make it, or pull it from somewhere?

sabadam32:
I made `checker.png` myself

pushfoo (approved):
This builds locally and looks good. Thank you for reducing the size of the checkerboard image.

If you rephrase the PR's description to include the following, it'll auto-close the linked issue when merged:
```
Fixes #1878
```
I'm not sure if the current phrasing will get picked up by GitHub's automation.

As to parsing vs using Python's various built-ins, I think we should leave that as-is until after 3.0. I'm painfully aware of how ugly doc build is right now, and it'll take a long time to fix.

einarf:
This still WIP or done?

pushfoo:
@einarf I haven't built it recently, but I'm satisfied with it as-is, and was as of a few versions ago.

pushfoo (approved):
Ran build. It looks good. :+1:

PR #2008, opened 2024-03-03
Quick nuitka fix

DragonMoffon:
Nuitka automatically converts strings which look like paths ("C:/foo/bar/file.type") into a weird class. This PR adds a small try block to catch this issue in the resolver. It does not fix the issue everywhere, but it should work for the most part.

### DIFF
--- a/arcade/resources/__init__.py
+++ b/arcade/resources/__init__.py
@@ -95,6 +95,12 @@ def resolve(path: Union[str, Path]) -> Path:
         else:
             path = Path(path)
 
+    try:
+        path = Path(path.resolve(strict=True))
+    except Exception:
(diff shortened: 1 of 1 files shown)


arcade/resources/__init__.py
```diff
@@ -95,6 +95,12 @@ def resolve(path: Union[str, Path]) -> Path:
         else:
             path = Path(path)
 
+    try:
+        path = Path(path.resolve(strict=True))
+    except AttributeError:
+        # WARNING: This is due to an issue caused by Nuitka overriding strings into janky path objec
```
pushfoo:
```suggestion
        # WARNING: This is due to an issue caused by Nuitka overriding strings into janky path object
```

PR #2014, opened 2024-03-04
GUI:  finish public api of arcade.gui.property

eruvanos:
properties provide instance and value to listener (keep support for simple callbacks)

### DIFF
--- a/arcade/gui/property.py
+++ b/arcade/gui/property.py
@@ -25,6 +25,8 @@ class Property(Generic[P]):
     """
     An observable property which triggers observers when changed.
 
+... code-block:: python
+
         def log_change(instance, value):
             print("Something changed")
 
--- a/arcade/gui/property.py
+++ b/arcade/gui/property.py
@@ -49,7 +49,7 @@ class Property(Generic[P]):
     def __init__(
         self,
         default: Optional[P] = None,
-        default_factory: Union[Callable[[], P], Callable[[Any, Any], P], None] = None,
+        default_factory: Optional[Callable[[Any, Any], P]] = None,
     ):
         if default_factory is None:
             default_factory = lambda prop, instance: cast(P, default)
--- a/arcade/gui/property.py
+++ b/arcade/gui/property.py
@@ -25,7 +25,7 @@ class Property(Generic[P]):
     """
     An observable property which triggers observers when changed.
 
-... code-block:: python
+.. code-block:: python
 
         def log_change(instance, value):
             print("Something changed")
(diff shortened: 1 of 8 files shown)


arcade/gui/property.py
```diff
@@ -18,21 +18,39 @@ class _Obs(Generic[P]):
     def __init__(self, value: P):
         self.value = value
         # This will keep any added listener even if it is not referenced anymore and would be garbage collected
-        self.listeners: Set[Callable[[], Any]] = set()
+        self.listeners: Set[Callable[[Any, Any], Any]] = set()
```
pushfoo: Can we make this a generic somehow? As a warning, although I haven't used `ParamSpec` extensively, I know it can be limited in some ways if you use either older versions of the `typing_extensions` backports.

arcade/gui/property.py
```diff
@@ -18,21 +18,39 @@ class _Obs(Generic[P]):
...
 
 
 class Property(Generic[P]):
     """
     An observable property which triggers observers when changed.
 
+        def log_change(instance, value):
```
pushfoo, line 30: Both Paul Craven and I prefer the explicit `.. code-block:: python` style.
```suggestion
... code-block:: python

        def log_change(instance, value):
```

pushfoo (changes requested):
My overall impression is good and I like callbacks. I have one change to suggest and a question.

arcade/gui/property.py
eruvanos: the first arg can be what ever instance the property is defined in, the second is the new value, which could be resolved by Property[P] i guess.

arcade/gui/property.py
```diff
@@ -18,21 +18,41 @@ class _Obs(Generic[P]):
...
         # This will keep any added listener even if it is not referenced anymore and would be garbage collected
-        self.listeners: Set[Callable[[], Any]] = set()
+        self.listeners: Set[Callable[[Any, P], Any]] = set()
 
 
 class Property(Generic[P]):
     """
     An observable property which triggers observers when changed.
 
+... code-block:: python
```
pushfoo: My bad, typo in my last comment. It should be a double `.`, not a triple. :sweat_smile: 
```suggestion
.. code-block:: python
```

pushfoo (changes requested):

pushfoo:
TL;DR: After fixing my suggestion typo, merge as-is if you or einarf think it's fine

I'm not yet sure on the type annotations for this. My instinct is that there is something we can do to polish it, but I may also be wrong. Getting it merged and working takes priority over perfect types, however. It's not breaking anything else as far as I can tell, and the stricter type checkers aren't something we're in full compliance with yet anyway.

PR #2018, opened 2024-03-12
Fix GUI ellipsis type issues with mypy 1.9.0 / pyright

pushfoo:
### Changes

* Replace ellipsis with `Optional` + `None` syntax


eruvanos:
I see it appears in Python >= 2.10 -.-

Maybe there is a way to fix that type hints for the background.

eruvanos:
Fixed with #2014 , thanks for your code, I took the padding changes

PR #2021, opened 2024-03-13
Name scale properties on Sprites to things that make sense

DigiDuncan:
## What does it do?
- Allows setting `scale` to a `float` for backwards compatible codefeel/easy proportional scaling
 - Getting `scale` returns a `tuple[float, float]`
- Renames `scale` -> `scale_x`
- Renames `scale_xy` -> `scale`
- Adds `scale_y`

## Why should this be done?
- **The current implementation lies to you.** `scale` as it is returns `scale_x`, which is incorrect if x/y are scaled independently.
- **There is precedent.** `Sprite` currently returns `position`¹ as a 2-tuple, and `center_x` and `center_y` as floats.
- **Learning wrong is worse than not learning.** The current implementation is more confusing for children in the long run. Most game libraries represent `scale` in both dimensions, and Arcade should aim to provide transferable knowledge.
- **It's confusing for a developer.** If you *are* a games dev, understanding why `.scale` returns x, `scale_y` returns y, `scale_xy` returns a tuple, but setting `scale` sets both x and y, is very confusing and makes code harder to read and reeks havoc on codefeel.

¹I'd argue `position` should be `center` as well, but that's for another day.


pushfoo:
TL;DR: This will be a non-breaking change & efficiency upgrade with pyglet 2.1, which seems to be [coming sooner than expected.](https://discord.com/channels/438622377094414346/701585557637038152/1218081219145895936)

Benefits:
1. `sprite_instance.scale *= 2.0` will work with scalars as it does now
2. `sprite_instance.scale *= 2.0, 3.0`:
    * It'll just work with arbitrary sequences (pyglet 2.1 feature)
    * Cleaner
    * More efficient than multiplying `center_x` and `center_y` individually ([both iterate over and update all referencing `SpriteList`s](https://github.com/pythonarcade/arcade/blob/3722e4491f66a0b2ebb8e864b3688c9d8ab0ccba/arcade/sprite/base.py#L82-L116))
    * More reliable than doing math on the individual axes yourself

Steps required:
- [x] Bump to pyglet 2.1 (#2043)
- [x] Make `BasicSprite.scale` return pyglet 2.1's tuple-based `Vec2`s
- [ ] Improve doc to use/review position to teach tuples, then lead into scale
    1. cute diagram with scale x and scale y, stretching a snake image
    2. scale example with interactive GUI sliders to scale sprite
    3. add links to healthbar examples
- [ ] Revert any example temp fixes to pass CI / typing

einarf:
What about the initializer? Should `scale` still only be a `float` there?

pushfoo:
Imo, it could be either without an `isinstance` check:
```py
if hasattr(scale, "__len__"):
    self._scale = Vec2(*scale)
else:
    self._scale = Vec2(scale, scale)
 ```

einarf:
I resolved the conflicts. There was a lot of changes to unit test so these needs to be looked into.

pushfoo:
> ["yeah just merge" - einarf](https://discord.com/channels/458662222697070613/1257099454461251664/1257189981697605644)

PR #2022, opened 2024-03-13
Add command to install latest build to docs (#1316)

### LINKED ISSUE #1316: Documentation: Add explanation of how to install bleeding edge arcade builds
### What documentation needs to be added

Clepto reminded a user and of how to install bleeding edge arcade builds:

```
pip install -I https://github.com/pythonarcade/arcade/archive/refs/heads/development.zip
```
We should mention this somewhere in the doc.

### Where should it go?

Either as a tip box in the current install documentation, or in a dedicated "Advanced" section under the install root page with warnings about instability.

### Why should we include it?

It has been forgotten at least twice before, and it will make installing bleeding edge arcade builds much faster than trying to play with git.

lbcai:
Fixes #1316

### DIFF
--- a/arcade/gui/events.py
+++ b/arcade/gui/events.py
@@ -21,8 +21,8 @@ class UIMouseEvent(UIEvent):
     Covers all mouse event
     """
 
-    x: float
-    y: float
+    x: int
+    y: int
 
     @property
     def pos(self):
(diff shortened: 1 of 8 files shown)


pushfoo:
@lbcai Did you have an issue with this PR? I was hoping to get to reviewing this in depth tonight.

PR #2025, opened 2024-03-15
created a depth of field example

DragonMoffon:
a super quick example for someone on Discord. I thought it might be nice to have as an example in base arcade. It relies on the experimental post-processing blur class.

### DIFF
--- a/arcade/experimental/depth_of_field.py
+++ b/arcade/experimental/depth_of_field.py
@@ -1,0 +1,167 @@
+from typing import Tuple, Optional
+from contextlib import contextmanager
+from math import cos, pi
+
+from PIL import Image
+
+from arcade.gl import geometry, NEAREST, Texture2D
+from arcade.experimental.postprocessing import GaussianBlur
+from arcade import get_window, draw_text
+
+
+class DepthOfField:
+
+    def __init__(self, size: Optional[Tuple[int, int]] = None):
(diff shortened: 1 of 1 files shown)


pushfoo:
I've made a PR (https://github.com/DragonMoffon/arcade/pull/3) to clean this up a bit. It won't be perfect yet, but this should make it legible faster than going through the code review process. Once it's merged, I think it should be good.

pushfoo:
TL;DR: I want to rephrase this but need more time to think. Also, the original page is broken from a UX / readability perspective.

I'm currently in favor of keeping your changes, but in a rephrased form. I need more time to think about the best way because the original page is already hard to read. That's not your fault. The original page is just kind of broken.

The problem with the page you're trying to improve is:

1. We say "don't do X"
2. We then do the exact opposite

For beginners who just want to add UI, this is overwhelming. Three separate code versions would be better, but the example gallery is already crowded.

The third option is to investigate unifying examples into single pages with tab plugins, and there's a discord thread about that in the doc improvement channel. However. there's work that Clepto or I need to do first:

1. Fix the readthedocs build
2. Either:
   * Verify that PDF output still works with one of the known Furo-compatible tab plugins
   * Come up with a workaround

pushfoo (approved):
LGTM. It's experimental and readable enough for us to finish polishing it later.

PR #2026, opened 2024-03-16
Add example for positioning flat buttons to docs

### LINKED ISSUE #1127: Update example for flat text buttons

The example for [flat text buttons](https://api.arcade.academy/en/latest/examples/gui_flat_button.html?highlight=UIFlatButton).

### Where is it located?

https://github.com/pythonarcade/arcade/blob/development/arcade/examples/gui_flat_button.py

It would be good to show how to align the buttons at a desired coordinate. As there is no mention of it in the whole docs of flat text button.

lbcai:
Fixes #1127


pushfoo:
TL;DR: I want to rephrase this but need more time to think. Also, the original page is broken from a UX / readability perspective.

I'm currently in favor of keeping your changes, but in a rephrased form. I need more time to think about the best way because the original page original page is just kind of broken.

The problem with the page you're trying to improve is:

1. We say "don't do X"
2. We then do the exact opposite

For beginners who just want to add UI, this is overwhelming. Three separate code versions would be better, but the example gallery is already crowded.

The third option is to investigate unifying examples into single pages with tab plugins, and there's a discord thread about that in the doc improvement channel. However. there's work that Clepto or I need to do first:

1. Fix the readthedocs build
2. Either:
   * Verify that PDF output still works with one of the known Furo-compatible tab plugins
   * Come up with a workaround for the PDF build

eruvanos:
@pushfoo 
Should we take the changes, I like the additional explanations.

The other issues you mention are valid, independent of these additions, right?

pushfoo (approved):
TL;DR: Let's merge it as-is.

We'll be giving these examples another look anyway. Since @sabadam32's findings suggest tab plugins should work for these example pages, this might get moved and rephrased to a tab.

PR #2027, opened 2024-03-16
Fix homepage css

### LINKED ISSUE #2023: Documentation Issue: Contributing section on home page has alignment issues
## Bug / Documentation: 

TL;DR: The `Contributing` heading on the home page is ugly in full screen and half-screen views

### Replication steps

1. Check out the repo
2. `pip install -Ie .[dev]`
3. `./make.py serve`
4. Go to http://localhost:8000
5. Scroll down

####  Full screen
image

#### Half screen
image

lbcai:
Fixes #2023 

- Adjusted the top header where the sprinter and the star button are
- Made the Contributing section always have 1 column since there are only 2 items there. But if 2 columns are preferred, I'll change that back
- Added a little margin between sections
- Made Social and Learning Resources section swap to 1 column if screen is small enough, just like the other sections
- Adjusted star button for dark mode users. Before:
image
After:
image
It's not currently possible to make the iframe background match the color of the dark mode docs. To fix that, we could use this other solution (https://buttons.github.io/) that involves loading a Javascript script, but it seems like this background issue is only present in Firefox and not in Chrome so I imagine it's low priority.

### DIFF
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -4,61 +4,110 @@ body {
 /* TOC - Main landing page */
 /* Show two columns if screen is wide enough */
 @media screen and (min-width: 700px) {
+    /* For regular items */
     .toctree-wrapper > ul {
         columns: 2;
     }
+    /* For Social and Learning Resources sections */
     .toc-outside-links {
         columns: 2;
     }
 }
-.item-framing-box {
+
+/* Wrap individual main page items for easier group manipulation
+ of contents and images. See PR #2027 for main page css changes */
+.main-page-item-wrapper {
     align-items: flex-start;
     display: flex;
     margin: 10px;
 }
-.item-framing-box .single-col-box {
+.main-page-item-wrapper > .toctree-wrapper {
     width: 100%;
 }
-.item-framing-box .single-col-box ul {
+
+/* single-col-box for items on main page with 2 bullet points, use if
+ wanting to avoid having 2 columns with 1 item each inside */
+.main-page-item-wrapper > .single-col-box {
+    width: 100%;
+}
+.main-page-item-wrapper > .single-col-box ul {
     columns: 1;
 }
-.item-framing-box-main {
-    align-items: center;
-    display: flex;
-    margin: 10px;
-}
-.item-framing-box .item-framing-sub-box {
+
+/* For Social and Learning Resources to make
+title + list appear like other categories */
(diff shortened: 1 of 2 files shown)


doc/_static/css/custom.css
```diff
@@ -7,15 +7,64 @@ body {
     .toctree-wrapper > ul {
         columns: 2;
     }
+    .toc-outside-links {
+        columns: 2;
+    }
+}
+.item-framing-box {
```
pushfoo: 1. Can we add a brief comment above this group of rules to explain what they're doing and why? It could also link the issue / PR to avoid having the extra step of git blame.
2. Is there a more specific prefix we can use than `item-framing-`?

doc/_static/css/custom.css
```diff
@@ -7,15 +7,64 @@ body {
...
+    align-items: flex-start;
+    display: flex;
+    margin: 10px;
+}
+.item-framing-box .single-col-box {
```
pushfoo: I'm not yet sure on this approach. My first reaction is to use child rules rather than applying to all descendants, but I don't yet fully understand your intent.

pushfoo (changes requested):
TL;DR: Is it possible to split the GitHub stars fix into a separate, smaller PR to merge it in faster?

Visually, this definitely seems to work and I like it. However:
* We're mixing multiple concerns right now
* Adding some clarification comments might help us in the future
* Some of the CSS class names are a little vague, but that might be OK with comments

doc/_static/css/custom.css
lbcai: You're right, nice tip. I swapped them to child selectors on seeing that child selectors are more efficient than descendant selectors and less likely to cause someone else issues down the line.

doc/_static/css/custom.css
lbcai: I renamed some things, moved some things around to group them more nicely, and added comments. Hope it's more understandable now. Let me know if I should make some more changes. There isn't a github star button change in this PR, I just edited the size and position for visual purposes.

pvcraven (approved):

pvcraven:
The PR makes things better. Bit more code churn than I'd like.

Looking at it, I think that whole toc-outside-links can be removed. It was originally created when I didn't know how to do links (or was it supported at that point?) in a TOC, so I faked one.

See: [this other PR](https://github.com/pythonarcade/arcade/pull/2037)

Think this PR should be merged, then go back and simplify.

PR #2028, opened 2024-03-19
Save screenshot

DragonMoffon:
Current is broken and needs fixing on the backend


einarf:
As discussed: We alrady have `get_image` and ``get_pixel` so we should instead improve those. Possibly rethink these functions completely.

DragonMoffon:
Yup agreed.

PR #2029, opened 2024-03-19
Improved behavior of "visible" setter and getter properties in BasicSprite.

DragonMoffon:
This is so the alpha state is stored when setting visibility. Originally, it overrode the alpha, which was stinky.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -38,6 +38,7 @@ class BasicSprite:
         "_color",
         "_texture",
         "_hit_box",
+        "_visible",
         "sprite_lists",
         "_angle",
         "__weakref__",
@@ -49,6 +50,7 @@ class BasicSprite:
         scale: float = 1.0,
         center_x: float = 0,
         center_y: float = 0,
+        visible: bool = True,
         **kwargs: Any,
     ) -> None:
         self._position = (center_x, center_y)
@@ -57,6 +59,7 @@ class BasicSprite:
         self._width = texture.width * scale
         self._height = texture.height * scale
         self._scale = scale, scale
+        self._visible = visible
         self._color: Color = Color(255, 255, 255, 255)
         self.sprite_lists: List["SpriteList"] = []
 
@@ -306,16 +309,15 @@ class BasicSprite:
             sprite.visible = not sprite.visible
 
         """
-        return self._color[3] > 0
+        return self._visible
 
     @visible.setter
     def visible(self, value: bool):
-        self._color = Color(
-            self._color[0],
-            self._color[1],
-            self._color[2],
-            255 if value else 0,
-        )
+        if self._visible == value:
+            return
+
+        self._visible = value
+
         for sprite_list in self.sprite_lists:
             sprite_list._update_color(self)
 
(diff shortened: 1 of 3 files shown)


arcade/sprite/base.py
```diff
@@ -346,27 +348,14 @@ def color(self) -> Color:
...
-            self._color = Color(color[0], color[1], color[2], self._color[3])
-        else:
+        if color == self._color:
+            return
+
+        if len(color) != 3 and len(color) != 4:
             raise ValueError("Color must be three or four ints from 0-255")
 
+        self._color = Color(color[0], color[1], color[2], self._color[3] if len(color) != 4 else color[3])
+
```
pushfoo: I'd consider either using `Color.from_iterable` or just inlining its contents for speed.

The Color classmethod is cleaner, but using pyglet-style unpacking has the following benefits:

* Implicit length check while supporting non-sequence iterables for speed / flexibility
* Skips repeated method calls for index access and `len()` calls

https://github.com/pythonarcade/arcade/blob/5d8f82acfb94a0c048d4e577279494cbbac9fe92/arcade/types.py#L176-L188
The main room for improvement is clearer exceptions. The lines below handle checking for too many values:
https://github.com/pythonarcade/arcade/blob/5d8f82acfb94a0c048d4e577279494cbbac9fe92/arcade/types.py#L181-L183
However, we currently leave the underflow option uncaught. If it doesn't hurt perf, it might make sense to catch the the potential `ValueError` when unpacking, then raising a clearer one since sprites are core beginner-facing classes. For example, this would be sort of vague when setting a color if you're loading a value stored from config or somewhere else in the code base.
```py
>>> r, g, b, *a = (1,2,)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected at least 3, got 2)

```

tests/unit/sprite/test_sprite.py
```diff
@@ -324,12 +324,10 @@ def test_visible():
     # Make invisible
```
pushfoo, line 328: If we're changing the behavior, we should explicitly test for the new state persistence.
```suggestion
    # Set an alpha value which the visible property shouldn't change
    sprite.alpha = 127

    # Make invisible
```

tests/unit/sprite/test_sprite.py
```diff
@@ -324,12 +324,10 @@ def test_visible():
...
     sprite.visible = False
     assert sprite.visible is False
-    assert sprite.alpha == 0
```
pushfoo, line 327:
```suggestion
    assert sprite.alpha == 127
```

tests/unit/sprite/test_sprite.py
```diff
@@ -324,12 +324,10 @@ def test_visible():
...
 
     # Make visible again
     sprite.visible = True
     assert sprite.visible is True
-    assert sprite.alpha == 255
```
pushfoo, line 332:
```suggestion
    assert sprite.alpha == 127
```

arcade/sprite/base.py
```diff
@@ -306,16 +309,15 @@ def visible(self) -> bool:
...
 
     @visible.setter
     def visible(self, value: bool):
-        self._color = Color(
-            self._color[0],
-            self._color[1],
-            self._color[2],
-            255 if value else 0,
-        )
+        if self._visible == value:
```
pushfoo, line 332: If ``value`` is an int, this version will store it instead of a bool. I'm not sure what the best approach here is. The shortest would be as follows:
```suggestion
        if (value := bool(value)) == self._visible:
```
You could also split out the assignment into two lines:
```py
value = bool(value)
if self._visible == value:
```

arcade/sprite/base.py
```diff
@@ -57,6 +59,7 @@ def __init__(
         self._width = texture.width * scale
         self._height = texture.height * scale
         self._scale = scale, scale
+        self._visible = visible
```
pushfoo: Since we implicitly converted truthy values in the old setter, I'd do it here too.
```suggestion
        self._visible = bool(visible)
```

pushfoo (changes requested):
TL;DR: There are nasty, subtle details which need changes to take care of.

The general changes are good, however. I also think we might have uncovered room for improvement in `Color.from_iterable` since we leave an unpack underflow uncaught, and it results in a somewhat vague exception.

arcade/sprite/base.py
```diff
@@ -346,26 +349,19 @@ def color(self) -> Color:
 
     @color.setter
     def color(self, color: RGBA255):
```
pushfoo: This'll need an import change at the top, but I think we should do this while we're at it.
```suggestion
    def color(self, color: RGBOrA255):
```

pushfoo (changes requested):
We should fix this while we're here.

PR #2030, opened 2024-03-21
Make ColorLike match RGB/RGBA in both float and int

pushfoo:
TL;DR: Make `ColorLike` mean something instead of being broken

Ty @DigiDuncan for noticing this issue.

* Make `ColorLike` match RGB and RGBA in both int and float
* Move it to top of file with other color aliases


pushfoo:
Superceded by #2065

PR #2033, opened 2024-03-24
Simplify animation

einarf:
* Remove the append/remove keyframe methods. Animations should simply just a list of keyframes on creation
* Added `__slots__`


einarf:
This is so simple I'll just merge it now.

PR #2036, opened 2024-03-25
GUI: Add UIView

eruvanos:
@einarf this might be already enough.


einarf:
Seems the docs failed again. We really need to fix this system soon.

eruvanos:
I am aware of the build, wouldn't merge it. Still in draft state. But I agree, that it is not so nice, that an added file causes such issues all the time.

einarf:
Let's just sneak this one in for 3.0. I would like to see if I can fix the doc issues first. This one is a great test!

einarf:
Let's just merge this for now. It's such a simple addition. There are other ways to test the doc stuf

PR #2039, opened 2024-03-29
Add screenshot 1342

### LINKED ISSUE #1342: Build improved screenshot API

sabadam32:
Fixes #1342


tash apply
pushfoo, line 1: This looks like it was accidentally included.

arcade/application.py
```diff
@@ -930,6 +931,19 @@ def on_mouse_leave(self, x: int, y: int):
         """
         pass
 
+    def save_screenshot(self, location: Optional[str] = None) -> Path:
```
pushfoo: 1. I think `location` would be less ambiguous if renamed as `path` to avoid ambiguity about screen location
2. It may be a good idea to support `Union[Path, str]` here instead

arcade/window_commands.py
```diff
@@ -362,3 +363,8 @@ def some_action(delta_time):
     :param delay: Delay in seconds
     """
     pyglet.clock.schedule_once(function_pointer, delay)
+
+
+def save_screenshot(location: Optional[str] = None) -> Path:
```
pushfoo: Same comments as earlier for the `Window` method.

tash apply
pushfoo, line 1: Is this supposed to be here? I think this is accidental.

arcade/application.py
```diff
@@ -930,6 +930,19 @@ def on_mouse_leave(self, x: int, y: int):
...
+
+        img = self.ctx.get_framebuffer_image(self.ctx.screen)
+        if not location:
+            output_dir = Path().parent.absolute()
+        else:
+            output_dir = Path(location).absolute()
+
+        filename = f"{self.caption.lower().replace(' ', '_')}_{time.monotonic_ns()}.png"
+        full_file_path = output_dir / filename
+        img.save(full_file_path, 'PNG')
```
pushfoo: I'm not sure on this behavior. This is hard-locking the path and format.

tests/unit/window/test_screenshot.py
pushfoo, line 1: This should probably use the [`tempfile` module](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) and its temp dir helper to avoid breaking things.

pushfoo (changes requested):
I can't speaks to the GL side, but there are some immediate changes I see as room for improvement. Included file in here + some possible improvements.

arcade/application.py
sabadam32: So what do we want to do here?
My thinking based on the discussion was that only input for the time being would be a path to store the files.  Then we need to auto-gen a filename based on some format and save it in that path.  I figured that the window title and timestamp would do the trick.  We can make it fancier if you want.  We can make params with sane defaults for filename generation for example, but honestly whenever I use screenshot functionality I would not expect to control it at that level.
Thoughts?

tests/unit/window/test_screenshot.py
sabadam32: yep. 😪

arcade/application.py
sabadam32: I guess we could just make it save the image with whatever the user passes in.  They control the file and path creation and just pass in the full deal.  Is that more what you were thinking

einarf:
What about the old `arcade.get_image()` and `arcade.get_pixel()`? We also talked about having screenshot feature in cameras. I think `get_image()` also have some history related to pixel ratio > 1.0. I don't quite remeber if we ended up scaling the screenshot itself.

pushfoo:
> also have some history related to pixel ratio > 1.0

Is this HiDPI or strange edge cases where non-square pixels exist? To my knowledge, that only exists on ancient platforms, but I figured I should double check.

arcade/__init__.py
```diff
@@ -34,6 +35,50 @@ def configure_logging(level: Optional[int] = None):
         LOG.addHandler(ch)
 
 
+def get_timestamp(
+        how: str = "%Y_%m_%d_%H%M_%S_%f%Z",
+        when: Optional[types.HasStrftime | datetime] = None,
```
pushfoo: 1. Did `HasStrftime` not match `datetime`? Was it only pyright which complained?
2. To my understanding, the `|` syntax is a 3.9+ feature, but we support 3.8+

arcade/__init__.py
```diff
@@ -9,7 +9,7 @@
 # Error out if we import Arcade with an incompatible version of Python.
 import sys
 import os
-from datetime import datetime
+from datetime import datetime, _TzInfo
```
pushfoo: My initial PR might have been a little busted due to my local nodejs breaking my pyright (I know pyright shouldn't depend on node, but it does :cry:). This should work (I think):
```suggestion
from datetime import datetime, tzinfo
```
The Python source for the datetime module shows it as a real ABC, and it should be less brittle than `_TZInfo`.

arcade/__init__.py
```diff
@@ -37,8 +37,8 @@ def configure_logging(level: Optional[int] = None):
 
 def get_timestamp(
         how: str = "%Y_%m_%d_%H%M_%S_%f%Z",
-        when: Optional[types.HasStrftime] = None,
-        tzinfo: Optional[datetime.tzinfo] = None
+        when: Optional[types.HasStrftime | datetime] = None,
+        tzinfo: Optional[_TzInfo] = None
```
pushfoo: I think I removed a top-level `datetime` import but left behind the broken property `datetime.tzinfo` line earlier. With the import suggestion at the top, I think this should work.
```suggestion
        tzinfo: Optional[tzinfo] = None
```

pushfoo (changes requested):
TL;DR: `_TzInfo` is protected and ugly, but so was my previous PR due to pyright breakage. Sorry about that.

I'd run a local pyright check on the changes I proposed. My pyright is broken because of nodejs issues on my distro, sadly. Sorry about my previous PR being a little broken because of it.

arcade/__init__.py
sabadam32: No it didn't.  pyright complained for sure, but I don't remember if the build failed due to this check

pushfoo: Are you sure? I see references to line 41 in the GitHub CI builds on the previous commits. That's the other line it seems.

pvcraven:
I like having the additional docs.

Although, in the past, when people have wanted to do a screenshot we've just pointed them to:

```python
image = get_image()
image.save('screenshot.png', 'PNG')
```
This seems like a lot to add to the library for screenshots when two-lines can also run a screenshot.

If we do add it, can we get some unit tests?

pushfoo:
TL;DR:
1. We have some unit tests already
   * [`tests/unit/window/test_screenshot.py`](https://github.com/pythonarcade/arcade/pull/2039/files#diff-67eac368d01775f184827e6e616b514aa47ff06869622ba8bcd3ded36df15e1e)
   * [`tests/unit/window/test_get_timestamp`](https://github.com/pythonarcade/arcade/pull/2039/files#diff-67eac368d01775f184827e6e616b514aa47ff06869622ba8bcd3ded36df15e1e)
2. It seems since it's all at once instead of over 3+ years
3. `save_image` is clear and discoverable

#### The Precedent is 2-3 Feature Exposures

> This seems like a lot to add to the library for screenshots

In the past, we slowly added features to the top-level namespace, `Window`, and `View`. Two of the top examples:

| Feature | `Window` | `View` | `Top-level` |
|------------|------------|--------|--------------|
| Clearing | `Window.clear()` | `View.clear()` | `arcade.start_render()` (`arcade.get_window.clear()`) |
| Setting BG color | `Window.background_color` | `View.background_color` | `arcade.set_background_color` |

#### Why Do It For Screenshots?

1. `View` is expected to share core API features with `Window`
2. `arcade.get_image` sets the precedent for exposing screenshots at the top level

#### Unit Tests

> If we do add it, can we get some unit tests?

Can you elaborate? We already have some here:
* [`tests/unit/window/test_screenshot.py`](https://github.com/pythonarcade/arcade/pull/2039/files#diff-67eac368d01775f184827e6e616b514aa47ff06869622ba8bcd3ded36df15e1e)
* [`tests/unit/window/test_get_timestamp`](https://github.com/pythonarcade/arcade/pull/2039/files#diff-67eac368d01775f184827e6e616b514aa47ff06869622ba8bcd3ded36df15e1e)

If you'd like, we add some to explicitly cover `View`.

einarf:
We can revisit this in 3.x. Let's keep this PR around.

There's things like docs in there that can be salvaged.

PR #2042, opened 2024-04-04
Fix sprite rotation in sprite_bullets_aimed

pushfoo:
This fixes an example missed in #1624:

* Fix direction update in arcade.examples.sprite_bullets_aimed
* Rephrase a few comments


pushfoo:
The CI failures are from pyright issues which appear related to text and GUI changes.

Issue #2045, opened 2024-04-05
camera_platform.py - Camera shake no longer working


einarf:
It seems this one broke. Likely a simple fix.

einarf:
Fix merged

PR #2046, opened 2024-04-06
Updating outdated camera code in example

DragonMoffon:
Firstly I updated the platformer tutorial to use the new Camera2D.

I also fixed a bug in the platformer_camera example.


DragonMoffon:
Fixes issue #2045

PR #2043, opened 2024-04-04
Switch to Pyglet 2.1 or dev releases

pushfoo:
TL;DR: Get new pyglet bug fixes + immutable vecs for #2021

### Why?

Per Discord discussion with @Cleptomania, this change:

1. enables [backward-compatible `BasicSprite.scale` behavior](https://github.com/pythonarcade/arcade/pull/2021#issuecomment-1999386042) in #2021:
    ```py
   # Will work with scalars as it does now
   sprite_instance.scale *= 2.0
   # Cleaner & more efficient than setting individual scale components
   sprite_instance.scale *= 2.0, 3.0
   ```
2. seems to show it works fine according to preliminary tests
3. matches how we've never made a release that wasn't on a preview pyglet build

### Changes

1. Use `pyglet==2.1dev2`
2. Update references to `pyglet.canvas` to refer to `pyglet.display`
3. Use new layout syntax per 2.1 styles

## Testing Steps Taken

- [ ] Human spot-check of examples (GUI, Text, and Media) on
  - [x] Linux
  - [ ] Windows
  - [ ] Mac

- [ ] Run unit & integration tests on
  - [x] Linux
  - [ ] Windows
  - [ ] Mac

## Follow-up Work

* Update #2021 with `BasicSprite` changes
* Thorough example / functionality shakedown per [the 2.1 migration guide](https://pyglet.readthedocs.io/en/development/programming_guide/migration.html)
* (Optional) Make arcade match the x, y, z style of pyglet's text / etc?

PR #2050, opened 2024-04-06
Fix Camera2D render target size bug & improve durability

pushfoo:
### Changes

1. Fix apparent bug where render target sizes are ignored by `Camera2D`
2. Use more durable keyword argument passing for dataclass arguments
3. Make `Camera2D.from_raw_data` inheritance-safe
4. Correct `Camera2D.viewport`'s docstring
5. Add tests for the changes above
6. Move the orthographic projector tests to an appropriately named file

#### Vec types
Once #2043 is merged, we should probably use its `Vec*` types liberally

#### Fancy / Named Tuple types?

Would `NamedTuple` or similar be better for internal `viewport` representation? Benefits I can see:
1. Provides named access which is cleaner and more legible
2. Named access may be faster for named props according to Ben's recent benchmarks for pyglet vectors (needs verification)

LRBT and similar types might be useful.

### Validation?

Currently we don't do any pyglet-style implicit unpacks or length checks. Maybe we should?


pushfoo:
The remaining type issues in the GitHub based tests all appear to be GUI stuff.

pushfoo:
TL;DR: Changes didn't break anything, it's just the GUI pyright issues again

The changes are responses to Discord feedback from DragonMoffon. GitHub based tests still report the same GUI breakage as before, and those are unrelated to this PR.

PR #2051, opened 2024-04-06
Feature: relative position helpers

DigiDuncan:
Really basic relative position helpers for a Window.

Adds two functions, `relative()` and `absolute()`, which allow to convert to and from screenspace coordinates.

Inspired by my adventures in Ren'Py.

If there's any issue with this, I apologize, my blood sugar is off today 😅

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -915,29 +915,29 @@ class Window(pyglet.window.Window):
         """
         pass
 
-    def absolute(self, normalized_pos: tuple[float, float]) -> tuple[float, float]:
+    def absolute(self, fractional_pos: tuple[float, float]) -> tuple[float, float]:
         """Convert a relative fractional pair to absolute screen pixels.
 
         As in the rest of arcade and OpenGL, the default coordinate system
         places the origin at the bottom left.
 
-        :param normalized_pos: A position where the x and y are fractional
+        :param fractional_pos: A position where the x and y are fractional
             values of the window size.
         :returns: A tuple of absolute pixel X and Y relative to bottom left.
         """
-
         w, h = self.size
-        return (normalized_pos[0] * w, normalized_pos[1] * h)
+        frac_x, frac_y = fractional_pos
+        return frac_x * w, frac_y * h
 
-    def relative(self, pos: tuple[float, float]) -> tuple[float, float]:
+    def relative(self, pixel_pos: tuple[float, float]) -> tuple[float, float]:
         """Convert absolute screen pixels to a relative fractional pair.
 
-        :param pos: a tuple of X and Y pixel positions within the window.
+        :param pixel_pos: a tuple of X and Y pixel positions within the window.
         :returns: A tuple of fractional X and Y relative to the bottom left.
         """
-
         w, h = self.size
-        return (pos[0] / w, pos[1] / h)
+        pixel_x, pixel_y = pixel_pos
+        return pixel_x / w, pixel_y / pixel_y
 
 
 def open_window(
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -915,6 +915,32 @@ class Window(pyglet.window.Window):
         """
         pass
 
+    def absolute(self, normalized_pos: tuple[float, float]) -> tuple[float, float]:
+        """Convert a position in normalized form to pixel-based, absolute form.
+
+        :param normalized_pos: a tuple of two floats, representing a position,
+        where 0.0 -> 1.0 in the x represents the far left and right of the window,
+        and 0.0 -> 1.0 in the y represents the very bottom and the very top.
+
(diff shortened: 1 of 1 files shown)


arcade/application.py
```diff
@@ -915,6 +915,30 @@ def on_mouse_leave(self, x: int, y: int):
...
+
+    def relative(self, pixel_pos: tuple[float, float]) -> tuple[float, float]:
+        """Convert absolute screen pixels to a relative fractional pair.
+
+        :param pixel_pos: a tuple of X and Y pixel positions within the window.
+        :returns: A tuple of fractional X and Y relative to the bottom left.
+        """
+        w, h = self.size
+        pixel_x, pixel_y = pixel_pos
+        return pixel_x / w, pixel_y / pixel_y
```
einarf: `pixel_y / pixel_y` should probably be `pixel_y / h` unless I'm completely misunderstanding the purpose of these methods... not to mention the risk of `ZeroDivisionError`.

arcade/application.py
```diff
@@ -915,6 +915,30 @@ def on_mouse_leave(self, x: int, y: int):
...
+    def absolute(self, fractional_pos: tuple[float, float]) -> tuple[float, float]:
+        """Convert a relative fractional pair to absolute screen pixels.
+
+        As in the rest of arcade and OpenGL, the default coordinate system
+        places the origin at the bottom left.
+
+        :param fractional_pos: A position where the x and y are fractional
+            values of the window size.
+        :returns: A tuple of absolute pixel X and Y relative to bottom left.
+        """
```
einarf, line 927: This docstring just seems to academic and lacks contex/example. It's unclear what "fractional pair" is. Also unsure is the method name itself is clear.

arcade/application.py
```diff
@@ -915,6 +915,30 @@ def on_mouse_leave(self, x: int, y: int):
...
+        w, h = self.size
+        frac_x, frac_y = fractional_pos
+        return frac_x * w, frac_y * h
+
+    def relative(self, pixel_pos: tuple[float, float]) -> tuple[float, float]:
+        """Convert absolute screen pixels to a relative fractional pair.
+
+        :param pixel_pos: a tuple of X and Y pixel positions within the window.
+        :returns: A tuple of fractional X and Y relative to the bottom left.
```
einarf, line 936: This docstring just seems to academic and lacks contex/example. It's unclear what "fractional pair" is. Also unsure is the method name itself is clear.

einarf:
If the idea here just being able to express/convert 0.0 -> 1.0 screen coordinates into pixel coordiantes regardless of aspect ratio? Is this really a common use case in arcade? I think this change lacks contex and more explanation.

DigiDuncan:
@einarf

> If the idea here just being able to express/convert 0.0 -> 1.0 screen coordinates into pixel coordiantes regardless of aspect ratio?

Yes. This is mainly for UI positioning or thing meant to be "pinned" to the sides of the windows.

For example, if you have a 1280x720 window, and you want something in the dead center, `(0.5, 0.5)` <-> `(640, 360)`
what this means is it doesn't matter how much you resize the window, you can always say `window.absolute((0.5, 0.5))` and get the correct place to place your element. Or if I want something to be 95% to the right at all time, `window.absolute((0.95, 0.5))`, etc.

It's based slightly off how Ren'Py handles positioning; it makes these calculations screen dimension-agnostic.

arcade/application.py
DigiDuncan: solved!

pushfoo:
> If the idea here just being able to express/convert 0.0 -> 1.0 screen coordinates into pixel coordinates regardless of aspect ratio?

TL;DR: Not just 0.0 to 1.0 since it's useful for OOB events

When I asked Duncan about whether this is intentionally unclamped, he pointed out that it's intentional. I can see the value of this since values may fall outside window and other bboxes, especially during mouse drag events.

This means "normalized" not quite right in my eyes. "Relative" is also too vague and misleading since it's not actually encoding the relative position to anything but the origin.

> Is this really a common use case

TL;DR: I think it would be if we built it out, especially for the UI

Percent-like layout would be useful for general UI layout imo. I'm not sure of how to handle it from an API standpoint yet, but I see its consistent usefulness in CSS percent units. 
In our case, I think the decimal form would be more useful since we already seem to have axis aligned UV values for UI element textures. Those area all quads anyway, right?

einarf:
Definitely not againts it, but method names and docstrings are very vague. I had to actually read the code to even start understanding what the methods are about. A little more text and an `Example::` block is a helpful start.

> "There are only two hard things in Computer Science: cache invalidation and naming things" -- Phil Karlton

einarf:
Let's close this for now and evaluate if we want something like this in 3.x

PR #2062, opened 2024-04-13
Eliminate pyright errors

pushfoo:
TL;DR: Non-`#type: ignore` fixes for GUI type issues through annotations + int conversion

### Why

1. Build failure details on every commit are:
   * wasting everyone's time since we always have to inspect the build failures to see what they mean
   * confusing and scaring less experienced contributors
2. Until pyglet 2.1 finishes the `float` conversion of their labels & layouts, we need to account for it:
    * pyglet 2.0.X (our current version) will probably keep the `int` behavior / storage for layouts
    * pyglet 2.1 + dev releases (our future version) will switch to `float` storage
      * text layout & other positioning may floor / round values to whole pixels
      * animatable / data storage would remain float to avoid error loss / jitter
    * pyglet hasn't implemented the new behavior yet
4. LRWH rectangles have uses outside of the GUI (#1853)

### Changes

1. Add `AsFloat = Union[float, int` type to:
   * Make pyright happy
   * Convey meaning to readers
2. Annotate + document `arcade.gui.Rect`
3. Add tests for existing `arcade.gui.Rect.scale` behavior
5. Add `rounding=None` behavior to `arcade.gui.Rect.scale` + tests
6. Add `int` conversion shims to stop pyright issues
   * We haven't yet switched to a pyglet `2.1dev*`release
   * The float storage behavior hasn't beein implemented on their side yet
   * would fix issues with their `.pyi` useful to reduce `#type: ignore` cruft

### Test steps taken
 
- [x] Add tests for any new functionality
- [x] CI
- [x] Human shakedown of the GUI after changes
  - [x] Linux (pushfoo)
  - [x] Mac (@eruvanos)
  - [x] Windows (DigiDuncan)

### DIFF
--- a/arcade/gui/widgets/__init__.py
+++ b/arcade/gui/widgets/__init__.py
@@ -31,7 +31,7 @@ from arcade.gui.events import (
 from arcade.gui.nine_patch import NinePatchTexture
 from arcade.gui.property import Property, bind, ListProperty
 from arcade.gui.surface import Surface
-from arcade.types import RGBA255, Color, Point
+from arcade.types import RGBA255, Color, Point, AsFloat
 
 if TYPE_CHECKING:
     from arcade.gui.ui_manager import UIManager
@@ -52,7 +52,7 @@ class Rect(NamedTuple):
     width: float
     height: float
 
-    def move(self, dx: float = 0.0, dy: float = 0.0) -> Self:
+    def move(self, dx: AsFloat = 0.0, dy: AsFloat = 0.0) -> Self:
         """Returns new Rect which is moved by dx and dy"""
         x, y, width, height = self
         return Rect(x + dx, y + dy, width, height)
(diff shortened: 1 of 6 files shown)


eruvanos:
Thanks for the contribution. 

@pushfoo Is this ready to merge? For me it looks fine, the new functionality within Rect seems to be useful and the AsFloat makes the usage of values clear.

pushfoo:
TL;DR: Give it a look over on Mac and Windows for the text scrolling text and text drawing examples?

I just went through those examples under Linux. It looks fine here as far as a quick glance can tell. @DigiDuncan @DragonMoffon could you two help out look at those examples under Windows? Also, I think Eruvanos has a mac.

Btw, scale changes are actually somewhat conservative. I'm considering more in the future but the current one seemed small enough to merge now.

PR #2064, opened 2024-04-13
Correct Camera2D Class comments

MiCurry:
This commit updates the Camera2D Class comments. It incorrectly stated that camera_data held the viewport, but it does not. The viewport is now defined in `OrthographicProjectionData`.


pushfoo (approved):
TL;DR: Merge ASAP, changes are small and good. :+1:

PR #2061, opened 2024-04-12
Add validation for projection dimensions in Camera2D + projections

pushfoo:
TL;DR:
1. Leave full doc rework until a later PR
2. Add validation which prevents delayed `ZeroDivisionError` which @MiCurry discovered
3. Add tests for validation

### Changes

- [x] Add validation to `Camera2D.from_raw_data`
  - [x] Add validation
  - [x] Add tests
- [x] Rename internal `*Data` attributes for clarity per Discord discussion
    * `_data` -> `_camera_data`
    * `_projection` -> `_projection_data`
- [x] Add `near` and `far` validation to `__init__`
- [x] Add validation to `.projection`

### Why?

1. Any of the following in an ortho projection cause a divide be zero:
  * `left == right`
  * `top == bottom`
  * `near == far`
4. It's hidden until you call `Camera2D.use` for the first time

Example traceback sanitized from [the Discord thread](https://discord.com/channels/458662222697070613/1228115132345745468/1228166379933274173):
```
  File "E:\Projects\SpaceGame\space.py", line 443, in on_draw
    self.cameras[player].use()
  File "C:\Users\User\AppData\Roaming\Python\Python312\site-packages\arcade\camera\camera_2d.py", line 697, in use
    self._ortho_projector.use()
  File "C:\Users\User\AppData\Roaming\Python\Python312\site-packages\arcade\camera\orthographic.py", line 134, in use
    _projection = self._generate_projection_matrix()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\User\AppData\Roaming\Python\Python312\site-packages\arcade\camera\orthographic.py", line 106, in _generate_projection_matrix
    return Mat4.orthogonal_projection(*_true_projection, self._projection.near, self._projection.far)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\python312\Lib\site-packages\pyglet\math.py", line 627, in orthogonal_projection
    sx = 2.0 / width
         ~~~~^~~~~~~
ZeroDivisionError: float division by zero
```
### Questions

- [x] ~~Should `ZeroProjectionDimension` subclass `ZeroDivisionError` or `ValueError`?~~ It's a `ValueError` since the math hasn't taken place yet.

### Follow-up work

- Add validation to future properties after rework

Issue #2057, opened 2024-04-08
Camera2D near and  near and far types


alejcas:
Hi, two related things:

1. It appears that calling Camera2D will default to near = 0.0, but calling Camera2D.from_raw_data will default to near = -100
2. near and far have float types but are set as integers by default by Camera2D.from_raw_data

I don't send a PR because I don't know if it should be 0.0 or -100.

DigiDuncan:
Pinging @DragonMoffon, they're the ones handling Camera2D primarily.

DragonMoffon:
For Camera2D it should be a float and -100.0 is correct.

pushfoo:
The commit Dragon made wasn't cherry-pickable. I asked him in Discord whether it would be okay to roll into that  branch. He said yes, so:
1. I redid it on my branch (0f75155215d6714348798e92e1f3b97d9268dadd)
2. It's credited to him

alejcas:
Closing this as it's fixed now

pushfoo:
TL;DR: I disagree since it's not merged and there will be force pushes breaking commit hashes.

It's not merged or released yet, but you can use that commit hash if you'd like assuming you mirror the branch to your repo. Btw, I strongly suggest mirroring and pinning if you want to use that branch ASAP because there will be commit hash changes.

alejcas:
> TL;DR: I disagree since it's not merged and there will be force pushes breaking commit hashes.
> 
> It's not merged or released yet, but you can use that commit hash if you'd like assuming you mirror the branch to your repo. Btw, I strongly suggest mirroring and pinning if you want to use that branch ASAP because there will be commit hash changes.

You are right, so reopening again

pushfoo:
@alejcas Can you pull the latest `development` commits to confirm this is fixed?

alejcas:
> @alejcas Can you pull the latest `development` commits to confirm this is fixed?

It is!

PR #2065, opened 2024-04-18
Initial cleanup of arcade.types

pushfoo:
## Changes

TL;DR: Spring cleaning to make 3.0-critical fixes easier

1. Make `arcade.types` a folder module
2. Move `Color` and related to `arcade.types.color`
3. Delete or reduce use of redundant or infrequent aliases and types:
   * Remove `NamedPoint`
     * only used in shapes_perf.py
     * Doesn't appear to be needed, but we can use `Vec2` once non-advertised API issues are resolved
   * Reduce usage of `IPoint` in favor of clearer `Size2D[int]`
   * Rename `arcade.types.Vector` to `Velocity`
      * Remove non-velocity uses
      * Leave cleanup of particles for later since it's related to a separate 3.0-mandatory issue (#1137)
6. Small pyglet 2.1 & pyright-related changes in `arcade.math`:
   * Remove `Vec*` from `V_2D` and `V_3D` aliases since `pyglet.math.Vec*` types are now tuples
   * Use `arcade.types.AsFloat` in a few places, such as `arcade.,math.lerp`
7. Update doc build to account for new type module structure

## Why?

TL;DR: Multiple 3.0-mandatory issues need good rectangle abstraction(s)
1. Issues listed below share a need for rectangle abstractions (or the all-important LRWH one if we have more)
2. We can refactor GUI's `Rect` into our rectangle abstraction
3. To make this easier for devs, we need to split and clean the over-stuffed `arcade.types` module
  * Navigating the current module is painful
  * None of the following really overcome the huge volume of stuff in the module:
     * PyCharm / VSCode's module structure tree views
     * Command palette jump-to
     * Old-school search by string / regex
  * Less experienced devs **can** help with type cleanup if we:
    1. Split the module to be more legible
    2. Structure our docs / types well
    3. Give them clear instructions and guidance on how to proceed

### The 3.0 issues which need good rectangles

The following 3.0-mandatory issues would all be massively helped along by sharing rectangle handling code to avoid bugs:

#### 1. Cameras

Due to lack of good rectangle abstractions, cameras individually compute and expose both raw and scaled values by:
1. Numerically indexing tuple elements individually (ie, `value[0]`)
4. Mixing the data, often with similarly-accessed attributes from camera `*Data` classes.

As a result of this, we have:
1. Hard-to-read code
4. 4+ known bugs and we keep finding more:
   * #2061  
   * #2057
   * A known issue where [some scaling math is backwards](https://discord.com/channels/458662222697070613/1228115132345745468/1228192811728240652)
6. Wasted function calls and `.` accesses due to individual recalculation of values

For example, if we want to check what's in the player's view, scaled projection boundaries currently waste 4 function calls and repeated `.` accesses recalculating the values:

https://github.com/pythonarcade/arcade/blob/d625a01de1584771966576081a038bb9a5defd27/arcade/camera/camera_2d.py#L364-L469

Instead, we should be recalculating the scaled rectangle bounds once relative to the projection centerpoint and returning it. For bound checking, we'll:
* always need all the values (`left <= x <= right and bottom <= y <= top`)
* can encapsulate bound checking inside a rectangle type, as [we already do for points](https://github.com/pythonarcade/arcade/blob/d625a01de1584771966576081a038bb9a5defd27/arcade/gui/widgets/__init__.py#L62)

#### 2. The 3.0-mandatory create_text_sprite rework

I've paused work on #1410 since I don't want to buggily rewrite bounds checking and scaling code.

#### 3. (Not actually 3.0) Screenshots

To complete #2039 to satisfaction, we'll want region handling. Rectangles can do that.

#### Rectangles are very nasty

First, we have at least 4 different ways of using / defining "rectangle" data. We won't fix this in this PR, but we'll clear room for it.

Two of the rectangle definitions are very bad:

1. The current `Rect` type alias inside `arcade.types`:
    * is labelled as `LBWH` in the comments
    * gets used as `LRBT` elsewhere
    * wastes space which could be expressed in a clean annotation / alias
7. `arcade.math.rand_in_rect` takes `bottom_left: Point, width: float, height: float`

We also have two good ones:

1. The LBWH `Rect` inside the GUI
   * It's a reasonable start
   * It implements `__contains__` as point bound-checking tool
   * It has a scaling helper which may need point-relative scaling

2. Separate left, right, bottom, top arguments to functions in `draw_commands`
   * These are beginner friendly so the format is okay
   * They're also great debug tools

#### The rest is cluttered

This PR tries to clean up some of the various redundant or one-off definitions cluttering `arcade.types`. It also moves `Color` and related defs into their own submodule.

### 3.0 Critical Work

1. Nailing down what our abstraction should look like, probably involving:
   * A `Rect` type which computes not just axis bound values (LRBT), but also width and height
     * You usually want the whole viewport bounds during real game development
     * width & height are useful too: if something's size is bigger than the viewport, skip detailed bounding checks
  * A `Protocol` with properties
   * Using `typing.NamedTuple` for base classes due to these being fastest for named access according to the pyglet team's tests
2. Refactoring our GUI's `Rect` into the decided reusable component
3. Following up with that rectangle in:
   * Current PRs
   * Future ones
2. Refactor `arcade.particles`:
   * (3.0 mandatory) #1137
   * While we're there, consider cleaning up `arcade.types.Velcocity`
     * Only remaining uses are in the particle emitters
     * It's consistent enough that this PR will get even bigger if we try to fix it here

### As time allows

1. Moving Arcade-specific exceptions out of `utils`
   * This might be good for cameras
   * Might not be entirely useful since we may want camera exceptions to go in the camera module
2. The math module needs cleaning
   * In-shape-bound randomization functions might be best on types?
   * If pyglet 2.1 vector issues are resolved without vendoring, we'll want to do some refactoring


DigiDuncan:
Praying to the merging gods we can get this in ASAP. Super important. Also, glorious write up, push.

DragonMoffon (approved):

PR #2069, opened 2024-04-20
 Clean up draw_commands in areas potentially helpful for #2065

pushfoo:
### Changes

TL;DR: Try to improve readability ahead of camera refactor research & prototyping

### Why?

TL;DR: Help @DragonMoffon & @DigiDuncan when working on cameras

1. Good news: I finished investigating our pyglet 2.1 `Vec*` issues much sooner than expected
  * Looks like a bad merge + overzealous `__eq__` check
  * Need to follow-up with pyglet team
  * Probably easily fixed with a `pyglet==2.1dev3` release
2. Rects would help make cameras less brittle (see #2065's follow-up / description)
  * We have a repeated root cause of projection / bounding box bugs due to lack of nice rect abstractions
  * The bugs due to this keep piling up as I perform intensive code review (newest is that up vector breaks properties)
3. Digi & Dragon's feasibility checks / "census" for rect abstraction may have glossed over function internals
   * Signatures inspected
   * Unsure about implementations, especially on hard to read code
4. The comments here a requests for further refactoring
5. Improving readability lowers friction for delivering those refactors in future PRs

### Where?

TL;DR: Function bodies where things help draw rects or might use quads

In `draw_commands.py`, wherever things look like:

1. Shared drawing helpers
7. Directly rectangle related
8. Appear to contain quad-like structures (see `draw_polygon_outline`)

### How?

TL;DR: Make the code understandable before even attempting to fix how fast it runs

1. Replace long ugly inlined logic with `Color`'s:
   1. `from_iterable` classmethod
   9. `normalized` property
7. Try to fail fast by putting these higher up:
   1. Window & context fetch
   2. Storing local refs to the context's shader & buffer singletons
   3. `Color.from_iterable(color)` validation
   4. Allocations of new `array.array`s
3. General readability improvements, including:
   1. Split nested one-liner expressions
   2. Store re-used quantities in local variables
   3. Add comments
   4. Remove redundant conversions such as the `tuple` in (`array.array('B', tuple(generator))`
   5. `*` unpacking positional point arguments instead of indexing

### DIFF
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -489,7 +489,7 @@ def draw_lines(point_list: PointList,
     # Validate & normalize to a pass the shader an RGBA float uniform
     color_normalized = Color.from_iterable(color).normalized
 
-    line_pos_array = array.array('f', tuple(v for point in point_list for v in point))
+    line_pos_array = array.array('f', (v for point in point_list for v in point))
     num_points = len(point_list)
 
     # Grow buffer until large enough to hold all our data
--- a/arcade/draw_commands.py
+++ b/arcade/draw_commands.py
@@ -542,16 +542,17 @@ def draw_points(point_list: PointList, color: RGBA255, size: float = 1):
 
     # Validate & normalize to a pass the shader an RGBA float uniform
     color_normalized = Color.from_iterable(color).normalized
+    num_points = len(point_list)
 
     # Resize buffer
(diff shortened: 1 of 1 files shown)


pushfoo:
1. I ran `python -m arcade.examples.drawing_primitives` and nothing seems off
2. If needed, I run do pixel comparison between branch outputs later

PR #2072, opened 2024-04-24
flip camera2D init and from_raw_data

alejcas:
This is the camera init/from_raw_data code flip we talk on discord.

Nota that from_raw_data cameras won't reference the same data objects but create new ones using data from the provided objects.

I have also fixed examples and tests.

I tested using the example codes and all worked

### DIFF
--- a/arcade/camera/camera_2d.py
+++ b/arcade/camera/camera_2d.py
@@ -179,7 +179,7 @@ class Camera2D:
         # build a new camera with defaults and after reuse provided camera objects
         new_camera = cls(render_target=_render_target, window=_window)
         new_camera._camera_data = _camera_data
-        new_camera._projection_data = projection_data
+        new_camera._projection_data = _projection_data
         new_camera._ortho_projector = OrthographicProjector(
             window=_window,
             view=_camera_data,
--- a/arcade/camera/camera_2d.py
+++ b/arcade/camera/camera_2d.py
@@ -14,9 +14,9 @@ from arcade.camera.data_types import (
 from arcade.gl import Framebuffer
 
 from arcade.window_commands import get_window
+
 if TYPE_CHECKING:
     from arcade.application import Window
-
 
 __all__ = [
     'Camera2D'
@@ -120,13 +120,13 @@ class Camera2D:
         )
 
     @classmethod
-    def from_raw_data(cls, *,
-                      camera_data: Optional[CameraData] = None,
-                      projection_data: Optional[OrthographicProjectionData] = None,
-                      render_target: Optional[Framebuffer] = None,
-                      window: Optional["Window"] = None) -> Self:
+    def from_camera_data(cls, *,
+                         camera_data: Optional[CameraData] = None,
+                         projection_data: Optional[OrthographicProjectionData] = None,
+                         render_target: Optional[Framebuffer] = None,
+                         window: Optional["Window"] = None) -> Self:
         """
-        Create a Camera2D with provided CameraData or/and OrthographicProjectionData object.
+        Create a Camera2D with reusing provided CameraData or/and OrthographicProjectionData object.
 
         :param camera_data: A :py:class:`~arcade.camera.data.CameraData`
             describing the position, up, forward and zoom.
@@ -175,20 +175,17 @@ class Camera2D:
             near=-100.0, far=100.0,
             viewport=(0, 0, width, height)
         )
-        _projection = (_projection_data.left, _projection_data.right,
-                       _projection_data.bottom, _projection_data.top)
 
-        return cls(
-            viewport=_projection_data.viewport,
-            position=_camera_data.position[:2],
-            up=_camera_data.up[:2],
-            zoom=_camera_data.zoom,
-            projection=_projection,
-            near=_projection_data.near,
-            far=_projection_data.far,
-            render_target=_render_target,
-            window=_window
+        # build a new camera with defaults and after reuse provided camera objects
+        new_camera = cls(render_target=_render_target, window=_window)
+        new_camera._camera_data = _camera_data
+        new_camera._projection_data = _camera_data
+        new_camera._ortho_projector = OrthographicProjector(
+            window=_window,
+            view=_camera_data,
+            projection=_projection_data
         )
+        return new_camera
 
     @property
     def view_data(self) -> CameraData:
@@ -239,12 +236,12 @@ class Camera2D:
 
         Useful for checking if a :py:class:`~arcade.Sprite` is on screen.
         """
-        return self._camera_data.position[0] + self._projection_data.left/self._camera_data.zoom
+        return self._camera_data.position[0] + self._projection_data.left / self._camera_data.zoom
 
     @left.setter
     def left(self, _left: float) -> None:
-        self._camera_data.position =\
-            (_left - self._projection_data.left / self._camera_data.zoom,)\
+        self._camera_data.position = \
+            (_left - self._projection_data.left / self._camera_data.zoom,) \
             + self._camera_data.position[1:]
 
     @property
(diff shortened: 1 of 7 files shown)


DragonMoffon:
While I understand why it was done the way it was, this fundamentally breaks a core assumption of the Cameras. The CameraData and OrthographicProjectionData objects have to be shared between objects.
It would be better to create the camera with default values and then override the attributes later

alejcas:
I can rewrite from_raw_data to build a camera with the arguments (c data and p data).

~~but i think it should be called “from_camera”:~~

~~`Camera2D.from_camera(camera: Camera2D)`~~

or “from_camera_data” with same arguments as “from_raw_data” but reusing the internal data objects.

~~what do you think?~~

rewrite done on fb025ec

arcade/camera/camera_2d.py
```diff
@@ -49,104 +49,40 @@ class Camera2D:
...
     contents are exposed via properties rather than directly to prevent this.
 
-    :param window: The Arcade Window to bind the camera to.
-            Defaults to the currently active window.
-    :param camera_data: A :py:class:`~arcade.camera.data.CameraData`
-        describing the position, up, forward and zoom.
-    :param projection_data: A :py:class:`~arcade.camera.data.OrthographicProjectionData`
-        which describes the left, right, top, bottom, far, near planes and the viewport
-        for an orthographic projection.
+    :param viewport: A 4-int tuple which defines the pixel bounds which the camera with project to.
```
MiCurry: 'with project to.' -> 'will project to'. Or 'will project onto'.

alejcas: I didn't write that but it should be "will project to" I'll change it

arcade/camera/camera_2d.py
```diff
@@ -49,104 +49,40 @@ class Camera2D:
...
+    :param position: The 2D position of the camera in the XY plane.
+    :param up: The 2D unit vector which defines the +Y-axis of the camera space.
```
MiCurry: I like the description used in `CameraData` better here:

``` Python
        up: A 3D vector which describes which direction is up (+y).
```

At least for my brain, I like seeing the word 'up' (with the additional mention of the positive y direction). It helps me understand that this controls the normal vector of the camera.

arcade/camera/camera_2d.py
```diff
@@ -49,104 +49,40 @@ class Camera2D:
...
+            i.e. a zoom of 2.0 halves the size of the projection, doubling the perceived size of objects.
+    :param projection: A 4-float tuple which defines the world space
+                bounds which the camera projects to the viewport.
+    :param near: The near clipping plane of the camera.
+    :param far: The far clipping plane of the camera.
     :param render_target: The FrameBuffer that the camera uses. Defaults to the screen.
         If the framebuffer is not the default screen nothing drawn after this camera is used will
         show up. The FrameBuffer's internal viewport is ignored.
+    :param window: The Arcade Window to bind the camera to.
+        Defaults to the currently active window.
```
MiCurry, line 65: Should we keep these comments written the same as the comments in `CameraData` and `OrthographicProjectionData`? 🤷‍♂️ Just something to think about.

arcade/camera/camera_2d.py
```diff
@@ -165,26 +101,75 @@ def from_raw_data(
...
+        )
+
+    @classmethod
+    def from_camera_data(cls, *,
+                         camera_data: Optional[CameraData] = None,
+                         projection_data: Optional[OrthographicProjectionData] = None,
+                         render_target: Optional[Framebuffer] = None,
+                         window: Optional["Window"] = None) -> Self:
+        """
+        Create a Camera2D with reusing provided CameraData or/and OrthographicProjectionData object.
```
MiCurry: How about:
```
Create a Camera2D from provided CameraData or/and OrthographicProjectionData objects.
```

Because it's possible that someone may use this function without another Camera. They may just create their own `CameraData` and/or `OrthographicProjectionData` objects.

arcade/camera/camera_2d.py
```diff
@@ -165,26 +101,75 @@ def from_raw_data(
...
+                    f"projection height is 0 due to equal {bottom=}"
+                    f"and {top=}"))
+            near, far = projection_data.near, projection_data.far
+            if near == far:
+                raise ZeroProjectionDimension(
+                    f"projection depth is 0 due to equal {near=}"
+                    f"and {far=} values"
+                )
+
+        # build a new camera with defaults and after reuse provided camera objects
```
MiCurry: How about: 

```
 build a new camera with defaults and then apply the provided camera objects.
 ```

**Comment by MiCurry** [comment]
I'm wondering if we might run into some shallow vs deep copy issues with `from_camera_data`. Specifically if someone tries to use the camera data from one camera to initialize another:

``` python
camera1 = Camera(...)
camera2 = Camera.from_camera_data(camera1.view_data, camera1.projection_data)
```

In that case I think adjusting either camera would affect the other, since they both contain references to each other. We should be able to easily test this in a unit test, which I'll make if I get some freetime today.

arcade/camera/camera_2d.py
alejcas: But in camera 2D it's a 2D vector...
I'll combine both with 2d

arcade/camera/camera_2d.py
alejcas: Yes, they should be there. A normal user will use a camera2D and not know nothing about what is CameraData.

arcade/camera/camera_2d.py
alejcas: Ok

arcade/camera/camera_2d.py
alejcas: ok

alejcas:
> I'm wondering if we might run into some shallow vs deep copy issues with `from_camera_data`. Specifically if someone tries to use the camera data from one camera to initialize another:
> 
> ```python
> camera1 = Camera(...)
> camera2 = Camera.from_camera_data(camera1.view_data, camera1.projection_data)
> ```
> 
> In that case I think adjusting either camera would affect the other, since they both contain references to each other. We should be able to easily test this in a unit test, which I'll make if I get some freetime today.

I think this is the intended purpose.

arcade/camera/camera_2d.py
```diff
@@ -165,26 +101,75 @@ def from_raw_data(
...
+        )
+
+    @classmethod
+    def from_camera_data(cls, *,
+                         camera_data: Optional[CameraData] = None,
+                         projection_data: Optional[OrthographicProjectionData] = None,
+                         render_target: Optional[Framebuffer] = None,
+                         window: Optional["Window"] = None) -> Self:
+        """
+        Create a Camera2D from provided CameraData or/and OrthographicProjectionData objects.
```
pushfoo: TL;DR: Something like this for explaining the purpose of the function?
```suggestion
        """Make a ``Camera2D`` directly from data objects.

        This :py:class:`classmethod` allows advanced users to:

        #. skip or replace the default validation
        #. share ``camera_data`` or ``projection_data`` between cameras

       .. warning:: Be careful when sharing data objects!

                    **Any** action on a camera which changes a shared
                    object changes it for **every** camera which uses
                    the same object.

       .. list-table::
          :header-rows: 1

          * - Shared Value
            - Example Use(s)

          * - ``camera_data``
            - Mini-maps, reflection, and ghosting effects.

          * - ``projection_data``
            - Simplified rendering configuration

          * - ``render_target``
            - Complex rendering setups

```

pushfoo (changes requested):
**TL;DR: The docstring change to clarify the shared data object behavior is the most we should do in this PR.**

@MiCurry the shared behavior is intended, but I agree we need to document it better and think about the CameraAPI a bit more.

However, this PR isn't the place for those changes. We can add a minimal docstring fix since @DragonMoffon and @DigiDuncan are finishing up prototyping improvements which will likely cause merge conflicts with further API changes.

arcade/camera/camera_2d.py
alejcas: I don't understand the dots before the warning or the list-table, and also the alignment of those docstrings.
I'll submit the change anyway in a few moments

MiCurry: @pushfoo yeah that looks good to me!

pushfoo (approved):
LGTM. A bunch of changes are coming, so we'll fix lingering API & doc issues in subsequent PRs.

Issue #2074, opened 2024-04-24
Properly handle copy/deepcopy 


einarf:
Some are using `copy` or `deepcopy` to clone arcade objects. Especially sprites. This can potentially add a lot of different bugs and weird issues. Possibly we can implement `__copy__` and `__deepcopy__` just raising an exception with a useful error message for now?

einarf:
Fix merged

PR #2076, opened 2024-04-26
Add exception-raising copy dunders via reusable decorator

pushfoo:
TL;DR: Use decorator to stub `__copy__` and `__deepcopy__` per #2074 + add a bunch of tests

### Changes

1. Add decorator which adds stubs with informative messages
2. Apply it to the following + add tests:
   * `SpriteList`
   * `BasicSprite` + tests for subclasses:
      * `Sprite`
      * `SpriteSolidColor`
      * `SpriteCircle`
   * `ShapeElementList` and `Shape`

3. Apply the decorator to the base `UIWidget`
   * This should cover all child widgets
   * Tests omitted since adding / removing the decorator as needed is fast and easy
   * It seems unlikely that people will be trying to copy widgets


DragonMoffon (approved):

einarf:
👍

PR #2088, opened 2024-05-22
Reinvent the Rectangle, or The First Step Towards Rectangle Unification.

DigiDuncan:
_Take that, [Thales of Miletus](https://www.britannica.com/biography/Thales-of-Miletus)._

For years, Arcade has suffered from Rectangle fragmentation. The [GUI](https://github.com/pythonarcade/arcade/blob/development/arcade/gui/widgets/__init__.py), [GL](https://github.com/pythonarcade/arcade/blob/development/arcade/gl/geometry.py), [Camera](https://github.com/pythonarcade/arcade/blob/development/arcade/camera/data_types.py), [draw commands](https://github.com/pythonarcade/arcade/blob/development/arcade/draw_commands.py), they all redefine and reinvent the rectangle. Need proof? **[We have proof.](https://docs.google.com/spreadsheets/d/1UIl_JXl_omVicYap7bkeQLxSGzDRtb2NluJvyWWVJWo/edit?usp=sharing)** But why have many rectangle when one **`Rect`** do trick?

With `Rect`, you can:
- Get every side, vertex, and size you'd ever want
- Reorient and resize the `Rect` to your heart's content
- Align it in frankly an absurd amount of ways
- Constrain and expand the heck out of it
- Union and intersect it with other `Rect`s (now with bonus `|` and `&` support)
- Find if points are in and on them
- Turn them into GL-friendly (and even GL-unfriendly!) tuples

They have a beautiful `__str__` method, backwards-compatible aliases, dead-simple property names, and implement *every way rectangles were used in the library*.

You can create them from `**kwargs`, or using one of several DX-friendly helper methods, including `LRBT`, `LBWH`, `XYWH`, `XYRR` (you're welcome, @einarf!), and `Viewport` (you're welcome @DragonMoffon!)

And they're *fast*. All eights values are pre-calculated on `__init__` (despite only needing four values to create), and they subclass `NamedTuple`, so dot-access is Sonic-level fast.

Immutable. Extendable. Human-usable.

Merging in this PR allows a full rework of every rectangle in Arcade, making them easy to use for contributors, developers, and end users.

And as a bonus, it cleans the heck out of `draw_commands`.


benjamin-kirkbride:
Could we get a mutable rect as well?

arcade/types/rect.py
```diff
@@ -0,0 +1,511 @@
...
+        return (
+            Vec2(left, bottom),
+            Vec2(left, top),
+            Vec2(right, top),
+            Vec2(right, bottom)
+        )
+
+    @property
+    def lbwh(self) -> RectParams:
+        """Provides a view into the Rect in the form of a tuple of (left, bottom, width, height)."""
```
eruvanos: Provides a tuple in the format of (left, bottom, ...)

arcade/types/rect.py
```diff
@@ -0,0 +1,511 @@
...
+def LBWH(left: AsFloat, bottom: AsFloat, width: AsFloat, height: AsFloat) -> Rect:
+    """Creates a new Rect from left, bottom, width, and height parameters."""
+    right = left + width
+    top = bottom + height
+    x = left + (width / 2)
+    y = bottom + (height / 2)
+    return Rect(left, right, bottom, top, width, height, x, y)
+
+
+def XYWH(x: AsFloat, y: AsFloat, width: AsFloat, height: AsFloat, anchor: Vec2 = AnchorPoint.CENTER) -> Rect:
```
eruvanos, line 460: I  am not sure about the uppercase function names.

@einarf what do you think?

arcade/types/rect.py
```diff
@@ -0,0 +1,511 @@
...
+    This object is immutable by design. It provides no setters, and is a NamedTuple subclass.
+
+    Attempts to implement all Rectangle functions used in the library, and to be a helpful
+    tool for developers storing/maniulating rectangle and rectangle-like constructs.
+
+    Rectangles cannot rotate by design, since this complicates their implmentation a lot.
+
+    You probably don't want to create one of these directly, and should instead use a helper method, like
+    :py:func:`~arcade.types.rect.LBWH`, :py:func:`~arcade.types.rect.LRBT`, :py:func:`~arcade.types.rect.XYWH`,
+    :py:func:`~arcade.types.rect.AnchoredXYWH`, or :py:func:`~arcade.types.rect.Viewport`.
```
eruvanos: :py:func:`~arcade.types.rect.AnchoredXYWH` gone

eruvanos (changes requested):

arcade/types/rect.py
DigiDuncan: Resolved in latest commit.

arcade/types/rect.py
DigiDuncan: I like this better, changed in latest commit.

arcade/types/__init__.py
```diff
@@ -143,8 +165,8 @@
 EMPTY_POINT_LIST: PointList = tuple()
 
 
-Rect = Union[Tuple[int, int, int, int], List[int]]  # x, y, width, height
-RectList = Union[Tuple[Rect, ...], List[Rect]]
+IntRect = Union[Tuple[int, int, int, int], List[int]]  # x, y, width, height
```
eruvanos, line 168: is xy bottom left or center?

eruvanos: Should this be a type if we have Rect? Rect could also be generic.

eruvanos:
Just agreed (with @DigiDuncan) that RectInt should be removed after we merged this huge PR.

eruvanos:
> Could we get a mutable rect as well?

Right now we stick to the immutable version only. From my point of view, the convenience benefits are not big enough to introduce a second Rect option, which is mutable.

eruvanos:
So summary:

I am fine with the PR. @einarf @pvcraven please give feedback what you think about the naming of the Rect constructor functions (XYWH, LBWH, ...) being uppercase.

I would merge the PR over the next day, if you give feedback afterwards. We can still do it in another PR.

> Background: The camera changes are blocked and they are one of the major points holding us back from 3.0 so I would like to unblock them.

eruvanos (approved):

eruvanos:
@DigiDuncan Can you check, if you can make the PR mergable?

PR #2089, opened 2024-05-23
Fixing the no culling version of the SpriteList shader

DragonMoffon:
It was never adequately updated to use the new atlas shader lib so it grabs the incorrect parts of the atlas. 

image


DigiDuncan (approved):
This desperately needs to be merged in!

PR #2094, opened 2024-05-26
Camera Touchups for arcade 3.0RC1

DragonMoffon:
Full change list:

- Added `Camera2D` to arcade's top level now that its `__init__` has been fixed
- Added generic functions to generate View, and Projection matrices from camera PoDs.
- Added generic functions to project and unproject coordinates using Matrices
- Added `StaticCamera`, which holds view and projection matrices directly for manual/temporary cameras
- Added Unproject and Project methods to all projectors
- Added new properties to `Camera2D` for key points around the visible boundary
- Added method to `Camera2D` to quickly test if a point is visible
- Moved the current and default camera to `ArcadeContext` so the default camera better behaves with framebuffers, and changing viewports
- Removed `map_screen_to_world_coordinate`, use `unproject` instead.
- Switching views now resets the active framebuffer to the screen and the active camera to the default
- `Window.viewport` now returns the screen's viewport, not the currently active viewport.
- The render target in `Camera2D` is now optional
- Fixed `Camera2D` activating the render target twice in `Camera2D.activate`
- Fixed `Camera2D` ignoring the viewport when initialising with a RenderTarget
- Fixed TextureAtlas not using a View matrix when rendering into a texture


DigiDuncan (approved):

DigiDuncan (commented):
~~Dragon in the Discord requested we wait to merge this since they found a bug.~~

Solved!

Issue #2097, opened 2024-05-31
Fix in method _on_player_eos (arcade/sound.py)


albecst:
I've been creating a 2D videogame with Arcade, and I've had some trouble stopping the sounds. Sometimes, randomly, the game would crash when we stopped (and the sound of the player's footsteps would stop), showing us this error:
```
File "...\Python\Python310\lib\site-packages\arcade\sound.py", line 87, in _on_player_eos media.Source._players.remove(player)
ValueError: list.remove(x): x not in list
```

You must check if there is a player (player: media.Player = media.Player()) to remove, so I'm about to open a PR to fix this.

Cleptomania:
This is with Arcade 2.6, I'm wondering if this was a problem of the player getting cleaned up early by Pyglet that maybe doesn't happen in the current version of pyglet.

Regardless, a try/except for the ValueError is probably the best approach here.

PR #2098, opened 2024-05-31
Fix crashing in method _on_player_eos (pythonarcade#2097)

albecst:
**Context:**

**Where:** *arcade/sound.py (method: _on_player_eos())*
As I explain in (#2097) when trying to stop the sound of the footsteps of the character I control, sometimes (randomly) the game crashed, showing the message I attach in the issue.

**Added:**
- Check if "player" exists.

**Testing:**
- When I added this comprobation in my project I didn't have any problems again, and we got rid of a bug that was closing the entire game.

### DIFF
--- a/arcade/sound.py
+++ b/arcade/sound.py
@@ -96,10 +96,11 @@ class Sound:
         media.Source._players.append(player)
 
         def _on_player_eos():
-            media.Source._players.remove(player)
+            if player in media.Source._players:
+                media.Source._players.remove(player)
             # There is a closure on player. To get the refcount to 0,
             # we need to delete this function.
-            player.on_player_eos = None  # type: ignore  # pending https://github.com/pyglet/pyglet/issues/845
+            player.on_player_eos = None # type: ignore  # pending https://github.com/pyglet/pyglet/issues/845
 
         player.on_player_eos = _on_player_eos
         return player


pushfoo:
To follow up on and confirm what @Cleptomania mentioned in #2097, what arcade version are you running? Try running `python -m arcade` and pasting the output here.

If it's not a 3.0.0 dev preview, the `development` branch may not be the right place for this. This branch is for an upcoming 3.0.0 release which uses  a much later version of pyglet than the Arcade 2.6.X releases.

albecst:
I didn't know that, my arcade version is: Arcade 2.6.17

```
Arcade 2.6.17
-------------
vendor: NVIDIA Corporation
renderer: NVIDIA GeForce RTX 3070/PCIe/SSE2
version: (4, 6)
python: 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
platform: win32
```

pushfoo:
TL;DR: Your PR may still be relevant, so please keep it open for the moment.

I'm going to ask pyglet's sound codec specialist about this. Your PR has highlighted two interesting things:

1. There's a comment nearby about some `None` issues (https://github.com/pythonarcade/arcade/issues/2097) and
2. `Source._players` is a list, so your code's `in` check is O(N)

The second item seems unavoidable with the current upstream implementation, but that makes me wonder: should the upstream code be different? It seems like this should be a set or something since a player should only be addable once.

Cleptomania:
If we do still need this fix, we can put it in a try/catch for raising the ValueError from the list removal to avoid the extra check for the item in the list. It'll make the exception path slower, but not slower than doing a list check.

albecst:
```"It seems like this should be a set or something since a player should only be addable once."```

That could be true, it's strange, maybe we need to change a few more things.

Answering to @Cleptomania , it's true that we can put it in a try/catch for the ValueError, instead of checking the "player" in the list.
That's up to you, you're the ones who understand the most about Arcade here! 😊

I just wanted to flag that, because it's not a superficial mistake that we can avoid, on several occasions I've had the game shut down (and if you don't have any kind of save in your game, you'll lost your progress...).

**Comment by Square789** [comment]
The error can be consistently replicated by just a few hacks in arcade's and pyglet's source
```py
    # --- arcade.Sound ---
    def stop(self, player: media.Player) -> None:
        with (c := media.get_audio_driver().worker._awful):
            c.wait()
        ...
    # --- pyglet's PlayerWorkerThread ---
    def __init__(self):
        ...
        self._awful = threading.Condition()

    def run(self):
        while True:
            ...
            with self._condition:
                with self._awful:
                    self._awful.notify()
                ...
```
and then calling `stop` close to when the sound ends.
Two problem flows i can think of are basically:
- The PlayerWorkerThread runs, the sound is finished and an `on_eos` event is dispatched
- The main thread receives this event, but runs the arcade program's `update` function first, which will eventually stop the sound and then doom the pending `on_eos` callback to fail.

or
- The PlayerWorkerThread takes over just as the `update` function is running, somewhere before `player.pause()` in `Sound.stop`
- The sound ends and an `on_eos` event is dispatched
- The main thread continues running, removes the sound from `Source._playing` and again sets the event handler up for failure

I'm not sure whether this list is needed at all. In pyglet, its only purpose is to store indirectly created players and delete that through an `atexit` handler once python terminates. It's half-baked as it does not delete explicitly created pyglet players: those may or may not be deleted since anything involving `__del__` and interpreter shutdown is extremely unreliable. But it's not causing problems and some cleanup is probably better than no cleanup.
The list could be turned into a set for faster membership detection and deletion.
Cleptomania's idea is the best option for arcade right now, could be replaced with `discard` if the set idea comes through.

pushfoo:
TL;DR: Would you be up for implementing the try/catch?

> Cleptomania's idea is the best option for arcade right now, could be replaced with `discard` if the set idea comes through.

@AlbeTroll I brought this up on Discord with Clepto, and I think there's general agreement: the try/catch route is probably the best. Do you want to try to implement it in this PR branch?

Cleptomania:
I've handled this in #2134

PR #2101, opened 2024-06-04
Add position rect to camera

alejcas:
Camera has rects for viewport and projection but not for position.
I use a lot the camera rect to obtain the map space rect currently on screen.

### DIFF
--- a/arcade/camera/camera_2d.py
+++ b/arcade/camera/camera_2d.py
@@ -13,7 +13,7 @@ from arcade.camera.data_types import (
 from arcade.gl import Framebuffer
 from pyglet.math import Vec2, Vec3
 
-from arcade.types import Point, Rect, LBWH
+from arcade.types import Point, Rect, LBWH, LRBT
 from arcade.types.vector_like import Point2
 from arcade.window_commands import get_window
 
@@ -433,6 +433,12 @@ class Camera2D:
 
         x, y = new_left
         self.position = x - up[1] * left, y - up[0] * left
+
+    def position_rect(self) -> Rect:
+        """ Get the LRBT rect of the camera position """
+        bx, by = self.bottom_left
+        tx, ty = self.top_right
+        return LRBT(left=bx, right=tx, bottom=by, top=ty)
 
     def point_in_view(self, point: Point2) -> bool:
         """
--- a/arcade/camera/camera_2d.py
+++ b/arcade/camera/camera_2d.py
@@ -434,11 +434,21 @@ class Camera2D:
         x, y = new_left
         self.position = x - up[1] * left, y - up[0] * left
 
-    def position_rect(self) -> Rect:
-        """ Get the LRBT rect of the camera position """
-        bx, by = self.bottom_left
-        tx, ty = self.top_right
-        return LRBT(left=bx, right=tx, bottom=by, top=ty)
+    def aabb(self) -> Rect:
+        """
+        Retrieve the axis-aligned bounds box of the camera's view area.
+        If the camera isn't rotated , this will be precisely the view area,
+        but it will cover a larger area when it is rotated.
+        """
+        tr_x, tr_y = self.top_right
+        tl_x, tl_y = self.top_left
+        br_x, br_y = self.bottom_right
+        bl_x, bl_y = self.bottom_left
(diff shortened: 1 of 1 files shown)


arcade/camera/camera_2d.py
```diff
@@ -434,6 +434,12 @@ def center_left(self, new_left: Point2):
         x, y = new_left
         self.position = x - up[1] * left, y - up[0] * left
 
+    def position_rect(self) -> Rect:
+        """ Get the LRBT rect of the camera position """
+        bx, by = self.bottom_left
+        tx, ty = self.top_right
+        return LRBT(left=bx, right=tx, bottom=by, top=ty)
```
DragonMoffon: Because the Camera2D can rotate, it doesn't make sense to return a rect as you have defined it. However, I think this is a valuable feature.

```python
def aabb(self) -> Rect:
    """
    Retrieve the axis-aligned bounds box of the camera's view area. 
    If the camera isn't rotated , this will be precisely the view area, 
    but it will cover a larger area when it is rotated.
    """
    tr_x, tr_y = self.top_right
    tl_x,, tl_y = self.top_left
    br_x, br_y = self.bottom_right
    bl_x, bl_y = self.bottom_left
    left = min(tl_x, tr_x, bl_x, br_x)
    right = max(tl_x, tr_x, bl_x, br_x)
    bottom = min(tl_y, tr_y, bl_y, br_y)
    top = max(tl_y, tr_y, bl_y, br_y)
    return LRBT(left=left, right=right, bottom=bottom, top=top)
```

DragonMoffon (changes requested):
Because the Camera2D can rotate, it doesn't make sense to return a rect as you have defined it. This is a useful feature, but there is a few changes I would recomend.

arcade/camera/camera_2d.py
alejcas: True! Changed as you suggested

DigiDuncan:
Like this idea a lot. Would love to see it get in, and as the person spearheading `Rect`, I definitely love hearing that it's getting practical use!

alejcas:
> Like this idea a lot. Would love to see it get in, and as the person spearheading `Rect`, I definitely love hearing that it's getting practical use!

Using Rects as if there was no tomorrow!

DragonMoffon (approved):
nice!

Cleptomania (changes requested):
I'm just sanity checking that the behavior makes sense.

If the camera can rotate, should the returned Rect be accounting for the rotation(as in, returning a Rect that is rotated) as opposed to returning a larger rect that covers the entire region of the actual rotated Camera plane?

I'm not sure what the behavior actually should be for that or which one makes more sense, but just wanna make sure if this has been considered or not.

DigiDuncan:
> I'm just sanity checking that the behavior makes sense.
> 
> If the camera can rotate, should the returned Rect be accounting for the rotation(as in, returning a Rect that is rotated) as opposed to returning a larger rect that covers the entire region of the actual rotated Camera plane?

`Rect`s can not be rotated by design, so it essentially has to for this behavior to do *anything*.

However, for a large amount of use cases, having this information is helpful (games where the camera does not meaningfully rotate, which is a lot of them.) But providing *something* if the camera rotates makes sense too.

Cleptomania:
> > I'm just sanity checking that the behavior makes sense.
> > If the camera can rotate, should the returned Rect be accounting for the rotation(as in, returning a Rect that is rotated) as opposed to returning a larger rect that covers the entire region of the actual rotated Camera plane?
> 
> `Rect`s can not be rotated by design, so it essentially has to for this behavior to do _anything_.
> 
> However, for a large amount of use cases, having this information is helpful (games where the camera does not meaningfully rotate, which is a lot of them.) But providing _something_ if the camera rotates makes sense too.

Lot of discussion about this topic happened on discord, but the TL;DR is that Rect does seem to be what other engines(namely Godot and Unity) are using for an Axis Aligned Bounding Box. There was a worry that Rect not having rotation would be misleading and not in line with other industry standard engines, but that seems to not be the case.

Regardless that's not super relevant to this PR, so this is probably fine as is.

Cleptomania (approved):

PR #2106, opened 2024-06-12
Move and slightly improve get_points_for_thick_line

pushfoo:
### Changes

* Move get_points_for_thick_line to drawing_commands, where it's used
* Remove empty drawing_support module
* Remove quick index reference to drawing_support
* Stop repeating operations in get_points_for_thick_line

### Y tho?
Indirectly helps us with quick index script issues.


arcade/__init__.py
```diff
@@ -68,7 +68,6 @@ def configure_logging(level: Optional[int] = None):
 # pyglet.options['win32_gdi_font'] = True
 
 # Imports from modules that don't do anything circular
```
FengchiW, line 70: Do we want to keep this comment here?

arcade/draw_commands.py
```diff
@@ -58,10 +57,49 @@
...
+    r1_x = start_x + shift_x
+    r1_y = start_y + shift_y
+    r2_x = start_x - shift_x
+    r2_y = start_y - shift_y
+    r3_x = end_x + shift_x
+    r3_y = end_y + shift_y
+    r4_x = end_x - shift_x
+    r4_y = end_y - shift_y
+
+    return (r1_x, r1_y), (r2_x, r2_y), (r4_x, r4_y), (r3_x, r3_y)
```
FengchiW, line 100: I know we're trying to move to returning vec2's is this something we want in this function too?

pushfoo: Probably. Same as the previous comment, but it also depends on whether we make this function internal or not. The docstring currently marks it as such, but it's exposed and named as if it isn't.

PR #2118, opened 2024-06-20
feat(gui): Use new rect type, delete GUIRect

eruvanos:


arcade/gui/property.py
```diff
@@ -23,26 +23,26 @@ def __init__(self, value: P):
 
 class Property(Generic[P]):
     """
-    An observable property which triggers observers when changed.
+        An observable property which triggers observers when changed.
```
pushfoo:
```suggestion
    """An observable property which triggers observers when changed.
```

arcade/gui/property.py
```diff
@@ -23,26 +23,26 @@ def __init__(self, value: P):
...
-        # > Something changed
+            my_obj.name = "Hans"
+            # > Something changed
 
-    :param default: Default value which is returned, if no value set before
-    :param default_factory: A callable which returns the default value.
-                            Will be called with the property and the instance
+        :param default: Default value which is returned, if no value set before
+        :param default_factory: A callable which returns the default value.
+                                Will be called with the property and the instance
```
pushfoo: This seems over-indented?

pushfoo (changes requested):
TL;DR: I only have minor docstring complaints.

PR #2137, opened 2024-06-26
Cleanup Log Levels

benjamin-kirkbride:
It is debatable if we need these at all, but at least they annoy me less like this :sweat_smile:


benjamin-kirkbride:
call me when we use black and isort

PR #2141, opened 2024-06-26
Black formatting + config

einarf:
Manually applied black formatting to most of the codebase manually inspecting and testing the changes,

A basic black config was also added in `pyproject.toml` setting line width to 120 and excluding several directories.


arcade/camera/grips/strafe.py
```diff
@@ -18,8 +18,4 @@ def strafe(data: CameraData, direction: Tuple[float, float]) -> Tuple[float, flo
     _pos = data.position
 
     offset = _right * direction[0] + _up * direction[1]
-    return (
-        _pos[0] + offset[0],
-        _pos[1] + offset[1],
-        _pos[2] + offset[2]
-    )
+    return (_pos[0] + offset[0], _pos[1] + offset[1], _pos[2] + offset[2])
```
pushfoo: This is separation is kind of meaningful until this file is cleaned up more.
```suggestion
# fmt: off
return (
        _pos[0] + offset[0],
        _pos[1] + offset[1],
        _pos[2] + offset[2]
    )
# fmt: on
```

arcade/experimental/clock/clock_window.py
```diff
@@ -777,16 +780,18 @@ def show_view(self, new_view: AnyView):
         """
         if not isinstance(new_view, View):
             raise TypeError(
-                f"Window.show_view() takes an arcade.View,"
-                f"but it got a {type(new_view)}.")
+                f"Window.show_view() takes an arcade.View," f"but it got a {type(new_view)}."
```
pushfoo: If we have one line, then no need for two separate f-strings. This should be merged into one.
```suggestion
                f"Window.show_view() takes an arcade.View, but it got a {type(new_view)}."
```

pushfoo (changes requested):
Initial look done on subfolders up to those starting with e.

arcade/camera/grips/strafe.py
einarf: This one is pretty straight forward to read in both forms imho. Subjective. I'll just commit the suggestion.

PR #2153, opened 2024-06-29
Initial conversion to PEP 585

Cleptomania:
Does an initial conversion of types to be in line with [PEP 585](https://peps.python.org/pep-0585/)

This is a full sweep of the library, places where `from __future__ import annotations` was not present, it was added in order for this to still work on Python 3.8.

After support for 3.8 is dropped, if the `__future__` import is not needed for other reasons, it could be dropped.

### DIFF
--- a/.github/workflows/selfhosted_runner.yml
+++ b/.github/workflows/selfhosted_runner.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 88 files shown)


Cleptomania:
Actually this is going to drop 3.8 support officially due to some use cases that the `from __future__ import annotations` does not cover, such as assigning subscripted tuples to variables(see `arcade/types/color.py` as an example of where we do this, but there are more).

PR #2154, opened 2024-06-29
Make `Camera2D` unique from `OrthographicProjector` for better state safety

DragonMoffon:


einarf:
Did you check how this looks when you generate docs? Likely viewport will be an undocumented member?

DragonMoffon:
Goodpoint will do that in a mo

PR #2169, opened 2024-07-01
Adding a stub Clock class and providing a fixed update event in `Window` and `View`

DragonMoffon:
Changes:
* clock.py with a `Clock` and `FixedClock` object
* `on_fixed_update` in `Window` and `View`
* deleted old experimental clock and window
* an example showing of value interpolation for smoother fixed updates


einarf:
I think this is pretty good and other libraries / engines do this. Possibly we need to get control over all the properties attached to the view and window.

PR #2170, opened 2024-07-01
Remove deprecation warnings from Rect as the final API for x/y/center is not decided

DigiDuncan:
See title.


pushfoo (approved):
Yep, I agree: this API needs work.

Issue #2116, opened 2024-06-19
Broken examples (collection for 3.0)


MiCurry:
Some examples are broken due to the 3.0 camera updates. So far there are:

- [x] `arcade/examples/procedural_caves_bsp.py` - Camera 'Jitters'
- [x] `arcade/examples/procedural_caves_cellular.py` - `AttributeError: module 'arcade.camera' has no attribute 'controllers'`
- [x] `arcade/examples/full_screen_example.py` - When swapping from fullscreen.
- [x] `arcade/examples/gl/custom_sprite.py`
- [x] `arcade/examples/camera_platform.py` (resize doesn't work right) 
- [x] `arcade/examples/easing_example_1.py` (looks wonky?)
- [x] `arcade/examples/easing_example_2.py` (due to other 3.0 changes)
- [x] `arcade/examples/turn_and_move.py` due to rotation direction changes
- [x] `arcade/examples/minimap_camera.py`
- [x] `arcade/examples/light_demo.py`

A few broken examples are listed in this comment. Others are listed in the comments below.

# procedural_caves_bsp.py

Running `python -m arcade.examples.procedural_caves_bsp` the camera 'Jitters':

https://github.com/pythonarcade/arcade/assets/2590700/ed9a40db-a610-4155-9c8d-a8b1131f820c

# procedural_caves_cellular.py

Running `python -m arcade.examples.procedural_caves_cellular` ends up in an attribute error:

```Python
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/miles/programs/arcade/arcade/examples/procedural_caves_cellular.py", line 350, in <module>
    main()
  File "/Users/miles/programs/arcade/arcade/examples/procedural_caves_cellular.py", line 346, in main
    arcade.run()
  File "/Users/miles/programs/arcade/arcade/window_commands.py", line 146, in run
    pyglet.app.run(window._draw_rate)
  File "/usr/local/lib/python3.11/site-packages/pyglet/app/__init__.py", line 80, in run
    event_loop.run(interval)
  File "/usr/local/lib/python3.11/site-packages/pyglet/app/base.py", line 165, in run
    timeout = self.idle()
              ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pyglet/app/base.py", line 226, in idle
    self.clock.call_scheduled_functions(dt)
[...]

PR #2171, opened 2024-07-01
Fix error when resizing the minimap in minimap_camera example

MiCurry:
The `minimap_camera.py` example was broken when the viewport was switched to use RECTs. Before this change, pressing 'i' or 'k' to resize the minimap would throw an error.

Thus, this PR updates the `minimap_camera.py` example to resize the minimap using LBWH when pressing 'i' or 'k'.

This PR addresses one example in #2116

### DIFF
--- a/arcade/examples/minimap_camera.py
+++ b/arcade/examples/minimap_camera.py
@@ -10,6 +10,7 @@ python -m arcade.examples.minimap_camera
 import random
 
 import arcade
+from arcade import LBWH
 
 SPRITE_SCALING = 0.5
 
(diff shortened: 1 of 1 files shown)


MiCurry:
Is LBWH the best way to resize a Rect?

einarf:
This is exactly the right solution. 

Two possible improvements;:
* Just use `arcade.LBWH` directly instead of explicitly importing it so users easily see where this member belongs
* Increase the window size to 720p (1280 x 720). All examples should ideally use this now. 
* BONUS: make the bottom text area a little bit bigger for better readability

(If you find more examples not using 720p you can get some "free" PRs from that :D)

MiCurry:
@einarf great suggestions. I added them!

pvcraven:
Thanks!

PR #2190, opened 2024-07-03
Adding a sub-event so users don't need to call `super().on_resize()`

DragonMoffon:


einarf:
Can we always assume `_on_resize` is called before `on_resize` here? This seems to be the case when testing.

einarf:
I messed up and pushed tweaks to branch in arcade repo. New PR is here https://github.com/pythonarcade/arcade/pull/2237

alejcas:
> Can we always assume `_on_resize` is called before `on_resize` here? This seems to be the case when testing.

Yes, that will always be the case. I test it  and commented it on discord

Issue #2186, opened 2024-07-03
SpatialHash's init should raise a clear TypeError on wrong cell_size type


pushfoo:
### Actual behavior:

1. Passing a `float` value to `SpatialHash` on creation raises a `ValueError`:
```python
>>> from arcade.sprite_list.spatial_hash import SpatialHash
>>> my_hash = SpatialHash(float('inf'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/src/arcade/arcade/sprite_list/spatial_hash.py", line 39, in __init__
    raise ValueError("cell_size must be an integer")
ValueError: cell_size must be an integer
```

2. Passing `"a string value"` raises an unclear `TypeError`:
```python
>>> from arcade.sprite_list.spatial_hash import SpatialHash
>>> my_hash = SpatialHash(float('inf'))
>>> my_hash = SpatialHash("a string value")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/src/arcade/arcade/sprite_list/spatial_hash.py", line 36, in __init__
    if cell_size <= 0:
TypeError: '<=' not supported between instances of 'str' and 'int'
```

### Expected behavior:

All incorrect types raise `TypeError("cell_size must be an int (integer)")`

### Steps to Correct

- [x] Move the `isinstance` check first
- [x] Make it raise a `TypeError` instead
- [x] Add tests in `tests/unit/spritelist/test_spatial_hash.py` in the `test_create()` function which use `pytest.raises` to check both cases above

PR #2201, opened 2024-07-04
Fixing the SpatialHash issue 2186

osm3000:
This PR is addressing the issue raised in https://github.com/pythonarcade/arcade/issues/2186

Steps (as kindly explained in the issue)

- [X] Move the isinstance check first
- [X] Make it raise a TypeError instead
- [X] Add tests in tests/unit/spritelist/test_spatial_hash.py in the test_create() function which use pytest.raises to check both cases above


einarf:
Thank you 👍

Issue #2139, opened 2024-06-26
Autoformatting


benjamin-kirkbride:
How do people feel about implementing (and enforcing) autoformatting in Arcade? Specifically:
- Black for code (120 char line length)
- Ruff for imports

Some considerations:
- we would likely need to disable this for sphinx purposes on some files
- git blame can be preserved (inconveniently) using this: https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

Cleptomania:
As far as ruff goes, we already use it for linting, so using it for the import sorting is trivial.

Really just need to exclude the examples code from autoformatting as far as exclusions go, maybe the experimental directory because there are technically examples in that? Not sure autoformatting matters for any of those though.

I don't think we care about git blame preservation

einarf:
We definitely don't care about blame preservation.

Cleptomania:
Things that probably need to happen for this:

1. A single, giant PR for initial conversion
2. Added to the CI to check for formatting(probably not having CI actually do any formatting though)
3. Add a tool to make.py to run it(CI could just use this probably)
4. Update contributing docs for it

benjamin-kirkbride:
Isn't step one actually some kind of broader consensus? Who needs to sign off?

benjamin-kirkbride:
Autoformatting the imports may break a bunch of stuff at first, due to.. load bearing import order

Cleptomania:
Those steps were just like, assuming it was happening, we probably have very little load bearing import order left. The 3.0 work cleaned up a lot of Arcade import. I would probably be in favor of excluding `__init__.py` files from the formatting though, the top level one is probably a minefield

eruvanos:
gui is already formated with ruff and I use pre-commit hooks to enforce it on my side.

pushfoo:
TL;DR: I am very wary of this idea.

The most important part of this (imo) is knowing how to turn it off. The following probably need to be excluded:

* `make.py`
* examples (as [Clepto said above](https://github.com/pythonarcade/arcade/issues/2139#issuecomment-2190397460))
* `arcade.experimental`
* Sphinx stuff
* The top-level `utils` folder can be fragile (includes more Sphinx stuff)

We're refactoring the utils folder one step at a time (#2109), but it's probably best to leave it alone for now.

[...]

PR #2205, opened 2024-07-04
fmt(all): set line-length to 100(like black config) and ruff fmt all

eruvanos:
#2139


einarf:
I already covered this in a pass a few days ago when I also fixed example line lengths.
The examples are limited to ~95 while the rest of the code is limited to 100.

We should probably have coordinated this better to avoid double work 😓

PR #2183, opened 2024-07-02
Bump to pyglet == 2.1dev3

pushfoo:
EDIT: Moved to #2226

Big changes:

1. Modified code to use `pyglet.options` object appropriately
2. No more `__eq__` deletion tricks
3. Partly document and type the pyglet text stack's apparent secret undocumented support for partial bolding (yes, I tested this)

Other changes:
- [x] Sound annotations
- [x] Align & `pyright`'s spec-violating ideas o on `Literal` for text
- [x] Document the apparent undocumented support for string bold values across non-win32 platforms
- [x] Use `| None` in a few places
- [x] Temp fix `float` vs `int` for window sizes (Apple supports these, apparently  to simplify touch events)

To resolve:
- [ ] Verify that the workaround for pyglet font loading changes work
- [ ] Figure out the current actual state of `Label`'s font return values
- [ ] What's going in with italic?
- [ ] The GUI stack & event handlers
- [ ] Whatever's going with `Screen` types and pyglet's import trickery

PR #2221, opened 2024-07-08
`Text` `int` typing for positions to `float`

DigiDuncan:
Pyglet types these attributes as `int` but happily accepts `float`, and we use float for position everywhere else and ¯\\_(ツ)_/¯

It was messing up my typing in Charm frankly so I figured other people would run into it

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -67,7 +67,9 @@ def _attempt_font_name_resolution(font_name: FontNameOrNames) -> FontNameOrNames
         elif isinstance(font_name, tuple):
             font_list = font_name
         else:
-            raise TypeError("font_name parameter must be a string, or a tuple of strings that specify a font name.")
+            raise TypeError(
+                "font_name parameter must be a string, or a tuple of strings that specify a font name."
+            )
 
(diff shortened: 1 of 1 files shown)


einarf:
@eruvanos was the one changing the typing on these in the past. Should be discussed with him.

Moving deeper I would also check with the pyglet people why these were typed as `int`. It's something we can change in pyglet for 2.1 if it makes sense.

pushfoo (approved):
**TL;DR:** :+1: Do it

This was partly covered in #2183 but that will take forever to merge, so I'll just rebase.

> Moving deeper I would also check with the pyglet people why these were typed as `int.`

I've been [discussing this with them on Discord](https://discord.com/channels/438622377094414346/1257652007364329473):

* Mac and Linux seem to allow float font sizes
* Windows does not seem to
* [The annotation seems like it should be float even if some renderers round to int](https://discord.com/channels/438622377094414346/1257652007364329473/1257690509892386896)

einarf:
Thanks for this fix. The squiggly lines in my editor have been bothering me 🤣

PR #2228, opened 2024-07-09
Sprite.x and .y

DigiDuncan:
In the future, if/when an anchor point is implemented on a sprite, this will be important, but for now, it's a shorthand for `center_x` and `center_y`.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -122,6 +122,36 @@ class BasicSprite:
         self.position = (self._position[0], new_value)
 
     @property
+    def x(self) -> float:
+        """Get or set the x position of the sprite.
+
+        Currently, this is equivalent to `.center_x`, but will change in the future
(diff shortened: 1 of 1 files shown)


einarf:
If you are using the black addon in VC Code you need to change the workspace configuration for black to read `pyproject.toml` to get the 100 line length instead of the default 80ish. Possibly black on command line will pick it up automatically.

einarf:
Possibly we should brew a little bit on this. There are other shortcuts possibly needed. For example setting the sprite position using the bottom left corner without caring about the hitbox. One can also argue sprite position should be bottom left corner in the first place like most other libraries, but using the center point right now gives us many shortcuts/optimizations. All these things are closely tied to hitboxes, collision/physics. If we do end up adding anchor support the benefits of the center point is thrown out of the window and we're back at the drawing board.

DigiDuncan:
Makes sense. I'll close this for now until post-3.0.

PR #2236, opened 2024-07-10
Remove font loading on arcade init

einarf:
* Don't load fonts in `arade.__init__`. This have been a headache for many users trying to distribute they games
* `arcade.resources.load_system_fonts()` can be called instead when you need them

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -406,19 +406,6 @@ __version__ = VERSION
 
 # Piggyback on pyglet's doc run detection
 if not getattr(sys, "is_pyglet_doc_run", False):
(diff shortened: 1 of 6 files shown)


DigiDuncan:
Looks good to me! I've been wanting this for a while.

DragonMoffon (approved):
I wasn't thorough, but every draw call I saw that needed one of the system fonts is covered.

einarf:
> I wasn't thorough, but every draw call I saw that needed one of the system fonts is covered.

Integration tests would go BOOM if anything was forgotten, thankfully.

PR #2247, opened 2024-07-12
Docs: CSS overhaul for new RTD Sphinx theme, Part 2

ryyst:
New stylistic features include:

### Specific to new dark mode
- All code blocks recolored
- All sphinx blocks recolored
- API reference links recolored in navigation
- API quick_index table recolored

All recolorizations were made to "match" the light mode as much as possible, with minimal artistic intervention. Eg. string symbol is red in light mode -> make a red-ish variant in dark.

### General
- New logo area!
- Examples page facelift!
- Minor color adjustments to light mode sphinx blocks

### Quick previews

image
image

### DIFF
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -103,21 +103,31 @@ html[data-theme="dark"] .wy-nav-top a {
     content: none;
 }
 
-/* Swap the ordering of default RTD logo/title elements and position accordingly. */
+/* Redesign the top-left logo area. */
 .wy-side-nav-search .icon-home {
     display: flex;
-    flex-direction: row-reverse;
+    flex-direction: column-reverse;
     align-items: center;
-    justify-content: left;
 
     padding: 16px 12px;
+    padding-bottom: 32px;
 
     color: var(--color-emphasis) !important;
+
+    border-radius: 16px;
 }
 .wy-side-nav-search .icon-home img.logo {
     margin: 0 !important;
-    padding: 0px;
-    padding-right: 10px;
+    padding: 0;
+    max-width: 100px;
+    padding-bottom: 16px;
+}
+.wy-side-nav-search > div.version {
+    font-size: 85%;
+    /* Same monospace as our `pre` blocks */
+    font-family: SFMono-Regular,Menlo,Monaco,Consolas,Liberation Mono,Courier New,Courier,monospace;
+    margin-top: -38px;
+    margin-bottom: 24px;
 }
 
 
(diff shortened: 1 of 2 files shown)


doc/_static/css/custom.css
```diff
@@ -181,6 +210,167 @@ html[data-theme="dark"] .wy-nav-top a {
...
+    color: #afafaf !important;
+}
+html[data-theme="dark"] pre .gp {
+    /* The REPL `>>>` token */
+    color: var(--color-emphasis) !important;
+}
+html[data-theme="dark"] pre .hll {
+    /* Line highlights background */
+    background-color: #323331 !important;
+}
```
pushfoo, line 336: Would it be possible to add a comment with a link to the upstream source / doc explaining what these classes do?

doc/_static/css/custom.css
```diff
@@ -95,21 +114,31 @@ html[data-theme="dark"] .wy-nav-top a {
...
-    padding: 0px;
-    padding-right: 10px;
+    padding: 0;
+    max-width: 100px;
+    padding-bottom: 16px;
+}
+.wy-side-nav-search > div.version {
+    font-size: 85%;
+    /* Same monospace as our `pre` blocks */
+    font-family: SFMono-Regular,Menlo,Monaco,Consolas,Liberation Mono,Courier New,Courier,monospace;
```
pushfoo, line 139: If it's the same, can we CSS variable it? [This post suggests](https://stackoverflow.com/questions/53846582/can-i-use-css-variables-in-a-font-list-and-have-it-work-in-legacy-browsers) it was an issue for legacy browsers, but 6 years ago.

pushfoo (changes requested):
**TL;DR:** Looks good but I have questions about comments to cross-ref and more CSS vars

doc/_static/css/custom.css
ryyst: Certainly would be possible, but honestly I have no idea which system is actually generating this syntax highlighting in the first place. This whole mess was basically just reverse-engineered with the browser inspection tool, which I reckon will be the easiest way in the future as well, as you can just click-select the specific symbol you're interested in currently.

doc/_static/css/custom.css
ryyst: Kinda same answer here as above: the "same" in this comment block just refers to the general styling rules I found on the page itself via inspection; no idea which system/theme/framework sets those fonts. Thus making a variable out of this wouldn't actually affect any other part of our custom.css than this single line, so on a glance the idea doesn't really make sense to me. (although sure it can also be a small stylistic choice)

Also this is sort of a rare case in styling anyway. Normally the monospace-ness of an element should already be dictated semantically by the element itself (`pre` `code` etc.).

einarf:
> **TL;DR:** Looks good but I have questions about comments to cross-ref and more CSS vars

Discord.

PR #2268, opened 2024-07-16
Improve header fonts

DigiDuncan:
We aren't a Western movie, and we're cooler than "Default Read The Docs." Let's get this place classed up a bit, yeah?

Replace Roboto Slab with [Josefin Sans](https://fonts.google.com/specimen/Josefin+Sans).

arcade_before_after


DigiDuncan:
Header (`h1`) updated to be a bit thicker.
firefox_jFFxW9Zm35

PR #2275, opened 2024-07-17
Docs: Fix a copy-paste error in 'Fastest Text Drawing' example

ryyst:
Essentially this made the example code exactly the same as in the 'Better Text Drawing' example.


einarf:
Whoops. Thank you!

PR #2304, opened 2024-07-23
Build docs in build/

einarf:
* Build docs in `build/` located in project root, not in the doc directory, This is not standard and it makes searching doc files a real plain.
* Re-enable the prettyspecialmethods plugin
* Fix broken color tables


pushfoo:
Haven't run it yet, but LGTM :+1:

PR #2305, opened 2024-07-23
docs(gui): convert to google doc and little cleanup

eruvanos:


pushfoo:
This needs follow-up on typing if possible before 3.0.

PR #2309, opened 2024-07-24
Attempt to fix doc build broken by #2304

pushfoo:
**TL;DR** #2304 broke readthedocs because it doesn't use make.py nor allow a good way to do so

### What broke?

Color lookup filename read since we don't set the relative path: https://readthedocs.org/projects/arcade-library/builds/25093429/

1. #2304 alters `make.py` to set the build directory
2. It does not update `readthedocs.yaml`
3. This doesn't matter since the only real control rtd offers comes in the form of:
    1. An environment variable control panel (aka bus factor) if it even works
    2. [Overriding the entire build process](https://blog.readthedocs.com/build-customization/) :upside_down_face: 

### Changes

1. Try to extract the root dir from the Sphinx object
3. Convert to absolute path
4. Pass the new path to the color table generation function instead
5. Explain why further in the source
6. Add some diagnostic printout with comments

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """Sphinx configuration file"""
-from typing import Any
+from functools import cache
+from typing import Any, NamedTuple
 import docutils.nodes
 import os
 import re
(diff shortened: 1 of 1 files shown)


pushfoo:
@einarf CI is [extra :hot_pepper:  today](https://github.com/pythonarcade/arcade/actions/runs/10076937039/job/27858346245?pr=2309):
1. textures are failing to allocate
```
self = <arcade.texture_atlas.atlas_default.DefaultTextureAtlas object at 0x7f48583ed570>
image_data = <ImageData width=1024, height=600, hash=ee8ea5d6ce2ceee8d128117e02145a5ae110bbd600e7514fc2110cb8d0c517b9>

    def _allocate_image(self, image_data: "ImageData") -> tuple[int, int, int, AtlasRegion]:
        """
        Attempts to allocate space for an image in the atlas or
        update the existing space for the image.
    
        This doesn't write the texture to the atlas texture itself.
        It only allocates space.
    
        Returns:
            The x, y texture_id, TextureRegion
        """
        image = image_data.image
    
        # Allocate space for texture
        try:
            x, y = self._allocator.alloc(
                image.width + self._border * 2,
                image.height + self._border * 2,
            )
        except AllocatorException:
>           raise AllocatorException(
                f"No more space for image {image_data.hash} size={image.size}. "
                f"Curr size: {self._size}. "
                f"Max size: {self._max_size}"
            )
E           pyglet.image.atlas.AllocatorException: No more space for image ee8ea5d6ce2ceee8d128117e02145a5ae110bbd600e7514fc2110cb8d0c517b9 size=(1024, 600). Curr size: (1024, 1024). Max size: (16[384](https://github.com/pythonarcade/arcade/actions/runs/10076937039/job/27858346245?pr=2309#step:6:385), 16384)
```
2. Dictionaries are changing themselves in the middle of a test:
```
self = <arcade.texture_atlas.atlas_default.DefaultTextureAtlas object at 0x7f48583ed570>

    @property
    def unique_textures(self) -> list[Texture]:
        """
        All unique textures in the atlas.
    
        These are textures using an image with the same hash
        and the same vertex order. The full list of all textures
        can be found in :py:meth:`textures`.
        """
        # Grab the first texture from each set
        textures: list[Texture] = []
>       for tex_set in self._unique_textures.values():
E       RuntimeError: dictionary changed size during iteration
```

PR #2310, opened 2024-07-25
Switch arcade.math.lerp to use a Protocol bound to a TypeVar

pushfoo:
**TL;DR:** Add `HasAddSubMul` protocol and update `Lerp` functions to return `Vec*` where appropriate

### Changes

* Add `HasAddSubMul` protocol type + `L` TypeVar in `arcade.math`
* Annotate `arcade.math.lerp` cleanly to allow arbitarary lerpables
* Make `lerp_2d` return `Vec2`
* Make `lerp_3d` return `Vec3`
* Docstring updates
* Cross-references for `lerp_angle`

### Follow-up work

- [ ] Should we have `lerp` check for an `object.lerp` method?
- [ ] How do we integrate that into type annotations?

One option for the latter is something like this:
```python
# Assume HasAddSubMul is the same as now

class HasLerp(Protocol[_T, _R]):
    def lerp(other: _T) -> _R:
```
- [ ] Weigh upstreaming the positional slash for maximum fidelity to math built-ins?

### DIFF
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -84,7 +84,7 @@ def lerp_2d(v1: V_2D, v2: V_2D, u: float) -> tuple[float, float]:
     return (
         v1[0] + ((v2[0] - v1[0]) * u),
         v1[1] + ((v2[1] - v1[1]) * u),
-     )
+    )
 
 
 def lerp_3d(v1: V_3D, v2: V_3D, u: float) -> tuple[float, float, float]:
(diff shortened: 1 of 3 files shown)


pushfoo:
As an update to this, it looks like only pyglet vector objects have the `lerp` method. The matrices do not.

DragonMoffon (approved):

PR #2315, opened 2024-07-26
Update easing_example_2.py

feiyuhuahuo:
update() takes no arg


einarf:
Like with the other pull request. You are probably using an other version of arcade here. `update()` definitely takes delta time in 3.0

einarf:
See https://github.com/pythonarcade/arcade/blob/fa6e9387cb9801081f378a4b4e978028ea531bd8/arcade/sprite/base.py#L581-L591

Issue #2120, opened 2024-06-22
GUI: all text blinks when UIInputText caret changes visibility


eruvanos:
## Bug Report

Text blinks when Caret of input fields are visible.

### Steps to reproduce/example code:

```
python -m arcade.gui.examples.exp_hidden_password
```

einarf:
Likely related to https://github.com/pythonarcade/arcade/issues/1158

einarf:
The `hidden_passord` module doesn't exist for me but other examples using input don't have this problem. Be free to reopen.

eruvanos:
Example was renamed (by me 😅)

`python -m arcade.gui.examples.exp_hidden_password`

einarf:
Definitely something fishy here. The blinking is still a problem in latest dev.

einarf:
Is it possible to make a minimal pyglet only version of this so we can narrow it down?

einarf:
Modified pyglet text input example don't show any blinking

```py
"""Demonstrates basic use of IncrementalTextLayout and Caret.

A simple widget-like system is created in this example supporting keyboard and
mouse focus.
"""

import pyglet
import pyglet.text.caret
import pyglet.text.layout


class TextWidget:
    def __init__(self, text, x, y, width, batch):
        self.document = pyglet.text.document.UnformattedDocument(text)
        self.document.set_style(0, len(self.document.text), dict(color=(255, 255, 255, 255)))
        font = self.document.get_font()
        height = font.ascent - font.descent

        self.layout = pyglet.text.layout.IncrementalTextLayout(self.document, x, y, 0, width, height, batch=batch)
        self.caret = pyglet.text.caret.Caret(self.layout, color=(255, 255, 255))
        # Rectangular outline
        pad = 2
        self.rectangle = pyglet.shapes.Rectangle(x - pad, y - pad, width + pad, height + pad, (200, 200, 220), batch)

    def hit_test(self, x, y):
        return (0 < x - self.layout.x < self.layout.width and
                0 < y - self.layout.y < self.layout.height)


class Window(pyglet.window.Window):
    def __init__(self, *args, **kwargs):
        super(Window, self).__init__(400, 140, caption='Text entry', *args, **kwargs)

        self.batch = pyglet.graphics.Batch()
        self.labels = [
            pyglet.text.Label('Name', x=10, y=100, anchor_y='bottom',
                              color=(255, 255, 255, 255), batch=self.batch),
[...]

PR #2317, opened 2024-07-26
"Fix" password blinking but break alignment and efficiency

pushfoo:
**TL;DR:** Prove it's something to do with layout / double-render trigger in the ugliest way possible. \:D

EDIT: this breaks the universe in a slightly different way. Editing the username now shows a cursor in both?

"Fix" #2120 by:
1. Draw the text as transparent
2. Remove layout manipulation during draw
3. Use `draw_text` + `Surface.activate` context manager to render badly over the surface

Anyone who wants to commit / PR into / replace this branch is welcome to.

### DIFF
--- a/arcade/gui/experimental/password_input.py
+++ b/arcade/gui/experimental/password_input.py
@@ -2,11 +2,60 @@ from __future__ import annotations
 
 from typing import Optional
 
+import arcade.color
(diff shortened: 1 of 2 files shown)


eruvanos:
I think we can drop this one, the blinking is already fixed, it was an issue with triggered rendering even when no changes where applied.

pushfoo:
Superceded by #2330.

PR #2327, opened 2024-07-29
Encapsulate relative position logic in UITextInput.on_event

pushfoo:
* Add relative position helpers
* Use position helpers to clean up UITextInput.on_event
* Some light refactoring for readability
* Improve related docstrings and comments

### DIFF
--- a/arcade/gui/widgets/__init__.py
+++ b/arcade/gui/widgets/__init__.py
@@ -11,7 +11,6 @@ from typing import (
     Tuple,
     List,
     Dict,
-    Type,
 )
 
 from pyglet.event import EventDispatcher, EVENT_HANDLED, EVENT_UNHANDLED
--- a/arcade/gui/widgets/__init__.py
+++ b/arcade/gui/widgets/__init__.py
@@ -1,7 +1,18 @@
 from __future__ import annotations
 
 from abc import ABC
-from typing import NamedTuple, Iterable, Optional, Union, TYPE_CHECKING, TypeVar, Tuple, List, Dict
+from typing import (
+    NamedTuple,
+    Iterable,
+    Optional,
+    Union,
+    TYPE_CHECKING,
+    TypeVar,
+    Tuple,
+    List,
+    Dict,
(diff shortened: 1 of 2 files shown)


pushfoo:
@DigiDuncan Let me know if the relative position helper seems to belong in `Rect` or if I somehow missed one like it already existing.

arcade/gui/widgets/__init__.py
```diff
@@ -167,6 +178,20 @@ def on_update(self, dt):
...
+            mouse_event: Any :py:class:`UIMouseEvent`.
+        Returns:
+            ``None`` if outside the widget or coords relative to
+             the widget's bottom left.
+        """
+        pos = mouse_event.pos
+        if not self.rect.point_in_rect(pos):
+            return None
+        return Vec2(pos[0] - self.left, pos[1] - self.bottom)
+
```
DragonMoffon, line 194: Update to use the new `Rect` method.

DragonMoffon (changes requested):

arcade/gui/widgets/__init__.py
pushfoo: I didn't update it yet because I want to give it more thought after rereading the UI code and considering the following:
1. 3.9 features will be our new minimum
2. The new rect method seems to perform more ops
3. It may add even more call overhead

eruvanos:
Hi, I am unsure about the state of the changes, due to the conflicts and the lag of an open issue, I would maybe close this PR?
@pushfoo let me know, if you thing the UITextInput requires the refactoring, from my point of view, right now it is kind of explicit. But I am always open for a second opinion.

pushfoo:
> I would close this PR
TL;DR: Agreed.

This thing was an exploratory kludge. I want to review your recent UI event proposals for controllers before attempting this + thoroughly review `Rect` since:
1. I need to in general
2. Dragon's "use rect" suggestion seemed to consume more operations at first glance
3. The fact I said "seemed" means we need to bench it

PR #2329, opened 2024-07-29
feat(window): add kwargs to open_window

eruvanos:
Do you think it is fine to add kwargs to the open_window? this would allow things like fullscreen etc


einarf:
Definitely reasonable.

einarf (approved):

PR #2343, opened 2024-08-07
updated ViewportProjector to use Rect, and created unit tests

DragonMoffon:


arcade/camera/default.py
```diff
@@ -123,16 +123,16 @@ def use(self) -> None:
...
         # If the viewport is correct and the default camera is in use,
         # then don't waste time resetting the view and projection matrices
-        if self._ctx.viewport == self.viewport and self._ctx.current_camera == self:
+        if self._ctx.viewport == self.viewport.viewport and self._ctx.current_camera == self:
             return
 
         # If the viewport has changed while the default camera is active then the
         # default needs to update itself.
         # If it was another camera's viewport being used the default camera should not update.
         if self._ctx.viewport != self.viewport and self._ctx.current_camera == self:
```
eruvanos: Is this not the same like line 126?

DragonMoffon: No the viewport stored by the ArcadeContext is a 4-tuple while the camera's is a Rect.

eruvanos: But shouldn't it also be changed to `self._ctx.viewport != self.viewport.viewport`

DragonMoffon: No, this is an early escape so when the camera's viewport matches the context's it doesn't need to re-calcuate and apply the matrices.

eruvanos: We are talking about line 132 right?

eruvanos: Maybe I am just confused, but why can you compare `self._ctx.viewport == self.viewport.viewport` in line 126.
and then `self._ctx.viewport != self.viewport` in 132?

should not both lines be adjusted?

eruvanos (dismissed):

arcade/camera/default.py
DragonMoffon: oh shoot. I see what you are seeing. Sorry I was only paying attention to line 126. Sorry about that u.u

PR #2345, opened 2024-08-07
feat(section): untie sections and arcade

eruvanos:
I would like to untie arcade and sections, we have a lot of discussions about Sections in general, this PR would ease the path, what ever we decide later.


alejcas:
How are `Section` `on_show_section` and `on_hide_section` called for the first time?

Now that `on_show_view` and `on_hide_view` no longer exists, when the view is displayed the section manager will no longer call each section `on_show_section`. 

Some functionality relies on this. 

Maybe `SectionManager.enable()` should call on_show_section on each one:

```
for section in self.sections:
    if section.enabled:
        section.on_show_section()
```
Then disable the opposite.

eruvanos:
SectionManager sets `section.enabled = True` 

The property calls `on_show_section()` already.
https://github.com/pythonarcade/arcade/blob/23e975244b0bad42db2d39068834f0cb2bdf366b/arcade/sections.py#L149

So that should be covered.

alejcas:
> SectionManager sets `section.enabled = True`

No, it won't work because section is enabled by default and if you set enabled to True it won't trigger anything:

https://github.com/pythonarcade/arcade/blob/23e975244b0bad42db2d39068834f0cb2bdf366b/arcade/sections.py#L150-L151

eruvanos:
Oh, I see, you are right. Will add the code you mentioned

eruvanos:
```
 for section in self.sections:
            # sections are enabled by default
            if section.enabled:
                section.on_show_section()
            else:
                section.enabled = True
                # enabling a section will trigger on_show_section
```

eruvanos:
@alejcas Would you be fine if I merge this state, if you have any concerns, just let me know.

alejcas:
> @alejcas Would you be fine if I merge this state, if you have any concerns, just let me know.

I think so!! go ahead

PR #2347, opened 2024-08-07
Fix point in polygon 

### LINKED ISSUE #2253: get_sprites_at_point broken
When a point intersects a line in the hitbox you get false positives. 

EDIT: Changes are already in `point-in-polygon-fix` branch. Unit tests for the geometry module have been broken up. They needs a bit of extending.

pushfoo:
* Fix #2253 by hardcoding a horizontal edge case (Ty @FengchiW !)
* Split up some tests
* Partial smoke test tool for collisions

Unless there's a good reason to wait, I suggest merging this once remaining test issues get straightened out.


einarf:
Remaining work:

* Sanity check fix
* Ensure the new geometry tests at least covers the basics¨
* Look into broken sprite tests

PR #2357, opened 2024-08-30
Changes for arcade 3.0 RC2 (dev34)

eruvanos:
This PR collects changes, which should make it into RC2.

- revamp UIGridLayout
  > the current implementation does properly integrate with size hints, for the huge 3.0 release introducing UILayouts a consistent size_hint handling should be ensured

- further documentation for layouts and how to create own widgets
  > the GUI got a major overhaul, so I want to make sure I spend enough time on the documentation
  
  
🏃‍♂️ I am focusing my free time on this and it should be done in just a few days. after that I am ready for RC2 and 3.0 afterwards.


eruvanos:
I already merge this, the grid revamp is the major thing I had on my list.

I have further improvements in the pipeline and a revamp of the draggable mixin, but this is not a road blocker for anything

PR #2361, opened 2024-09-06
Enable HiDPI scaling on Windows by default

einarf:
We should obey the UI scaling on windows by default. This is supported in pyglet 2.1. Users with 4k screens no longer need to look at a postage stamp of a window.

It looks like the scaling in pyglet doesn't work like on MacOS/retina were the window size stays unchanged. Only framebuffer size increases and you get a pixel ratio between them. The plan in pyglet is to support two modes:

* "window_only": you are supposed to scale the content yourself
* "window_and_content": Like MacOS/retina

Related issue: https://github.com/pyglet/pyglet/pull/1202


pushfoo:
I can't test this at the moment, but I don't see a reason not to merge it. @DragonMoffon or @DigiDuncan, can you try it on one of your systems?

einarf:
* Can be merged when 2.1.dev6 is out.
* We need to disable scaling for unit tests.

DragonMoffon:
Pyglet dev6 just released if we want to look into this again

pushfoo:
I finished #2385:
- [x] type checking / import linting
- [x] tests which occasionally fail (it's mouse position triggering extra mouse enter events)

PR #2365, opened 2024-09-14
add .tracking and px -> em conversion

DigiDuncan:
Added `.tracking` getter/setter to `Text`, as well as `.em_to_px()` and `.px_to_em()` for easy relative-to-font-size value conversion.

An example of why you may want this? To replicate setting this value in Photoshop:
image
one could write the code:
```py
# Note the Photoshop deals in micro-em, not em.
text.tracking = text.em_to_px(-.120)
```

As you can see, this is needed for some fonts to look good:
tracking_kerning_demo

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -638,14 +638,14 @@ class Text:
     def em_to_px(self, em: float) -> float:
         """Convert from an em value to a pixel amount.
 
-        1em is defined as `font_size`pt.
+        1em is defined as ``font_size``pt.
         """
         return (em * self.font_size) * (4 / 3)
 
     def px_to_em(self, px: float) -> float:
-        """Convert from an em value to a pixel amount.
+        """Convert from a pixel amount to a value in ems.
 
-        1em is defined as `font_size`pt.
+        1em is defined as ``font_size``pt.
         """
         return px / (4 / 3) / self.font_size
 
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -616,36 +616,33 @@ class Text:
     @property
     def tracking(self) -> float | None:
         """
-        Get the tracking amount for this text object, or rather,
+        Get/set the tracking amount for this text object, or rather,
         the added space between each character.
 
-        Returns a pixel amount, or None if the tracking is inconsistent.
+        The value is an amount in pixels and can be negative.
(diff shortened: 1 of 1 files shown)


arcade/text.py
```diff
@@ -613,6 +613,42 @@ def position(self, point: Point):
...
+
+    @tracking.setter
+    def tracking(self, value: float):
+        """
+        Get the tracking amount for this text object, or rather,
+        the added space between each character.
+
+        `value` is an amount in pixels and can be negative.
+        To convert from the em unit, use Text.em_to_px().
+        """
```
einarf: The setter docstring is never visible in docs. It needs to be merged with property. "Get or set" is a decent indication it's not a read-only property.

arcade/text.py
```diff
@@ -613,6 +613,42 @@ def position(self, point: Point):
...
+        self._label.set_style("kerning", value)
+
+    def em_to_px(self, em: float) -> float:
+        """Convert from an em value to a pixel amount.
+
+        1em is defined as ``font_size``pt.
```
einarf: You need a space after \`\` to make it valid rst.

arcade/text.py
```diff
@@ -613,6 +613,42 @@ def position(self, point: Point):
...
+        """
+        return (em * self.font_size) * (4 / 3)
+
+    def px_to_em(self, px: float) -> float:
+        """Convert from a pixel amount to a value in ems.
+
+        1em is defined as ``font_size``pt.
```
einarf: You need a space after \`\` to make it valid rst.

arcade/text.py
```diff
@@ -613,6 +613,42 @@ def position(self, point: Point):
         else:
             self._label.position = x, y, self._label.z
 
+    @property
+    def tracking(self) -> float | None:
+        """
+        Get the tracking amount for this text object, or rather,
+        the added space between each character.
+
+        Returns a pixel amount, or None if the tracking is inconsistent.
```
einarf: You can use google style docstring for the return here to make it more readable in docs

```
Returns:
    Explain the return value
```

einarf (changes requested):

DragonMoffon (approved):

einarf (approved):

PR #2370, opened 2024-09-22
Add simple tutorial on optimizing code with ctypes or numpy for beginners

eschan145:


DragonMoffon:
Does this page have a reference to Arcade Accelerate at all? Especially for collisions, it is where we want to lead new developers. Also, we recommend using rust over ctypes (though keep the details on ctypes as it is very helpful). It has a much nicer developer experience.

pushfoo:
TL;DR: This whole page needs some thought before I comment in-depth on this.

We may want to split performance into multiple pages. I need to look at some dependency issues for both 3.0 and the maybe-2.6.18 EOL release before coming back to this.

einarf:
This definitely needs a lot more thought into it. We need to maintain this document for years to come so covering the most relevant things based on experience from interaction with arcade users is likely the way to go?

I think the sections here are too vague or incomplete. "Algorithms and Raw Performance" however is a good idea to cover, but there are a lot more to cover other than the trivial case shown in the doc. I'm not sure how useful it is in its current state.

Optimizing using numpy is a complex topic. However the `sin` / `cos` functions shown in the document are actually slower than then standard library so I don't understand the purpose of those. If we are going to cover numpy it would need to highlight its strengths: Doing batch operations on larger arrays of data. Numpy arrays are also VERY useful in with `arcade.gl` because of the buffer protocol support meaning you can pass the raw internal memory of the array directly into textures and buffers without converting this data into python types (very costly).

The "Using cypes" section i think can instead be something more generic not necessarily mentioning specifics. There are so many options today, not only C/C++. I don't think we should go into specifics and I would be careful linking to content that might disappear in the future.

DragonMoffon:
Converting to a draft PR because there is much more work to be done on this part of the docs.

PR #2377, opened 2024-09-29
fix typing in arcade.math.clamp

DigiDuncan:


DragonMoffon (approved):
I am curious if there are other methods in arcade that could get this improved typing treatment? Stuff like the lerp function.

pushfoo:
I'm considering including something like [useful_types](https://github.com/hauntsaninja/useful_types). There was some part of it I didn't like for some reason. It was more than just another tiny dependency in our heaping pile. Looking at their PR queue, it may have been that the project is bit slow to merge.

> stuff like the lerp function

:sunglasses: Beat you to that one.
:point_down:
https://github.com/pythonarcade/arcade/blob/553894e398c00fc5a1421a95dca7ea6258b124ff/arcade/math.py#L49-L78

PR #2382, opened 2024-09-30
Bump pymunk to 6.8.1

pushfoo:
**TL;DR:** They only host the latest doc so our cross-refs will be wrong unless we do this


pushfoo:
I'm merging this now because:
1. It seems to just work when I run the joint builder demo (`python -m arcade.examples.pymunk_joint_builder`, etc)
2. Our unit tests pass
3. The [Pymunk change log](https://www.pymunk.org/en/latest/changelog.html#changelog) doesn't show any breaking changes

PR #2389, opened 2024-10-06
Sprite scale: Preserve performance

### LINKED ISSUE #2324: Review performance impact of `Sprite.scale` changes
This might just be about reverting https://github.com/pythonarcade/arcade/pull/2021

The vector sprite perf testing showed 2.x cost. We should quickly review the cost of the sprite scale changes before we release.

https://github.com/pythonarcade/arcade/blob/fa6e9387cb9801081f378a4b4e978028ea531bd8/arcade/sprite/base.py#L263-L339

We might just have to revert these changes for now and bake them into the new experimental sprites.
See : https://github.com/pythonarcade/arcade/issues/2184

### LINKED ISSUE #2399: Allow tuple scale in arcade.Sprite init, same as the setter

The `scale` argument to `arcade.Sprite.__init__` is a float (https://github.com/pythonarcade/arcade/blob/development/arcade/sprite/sprite.py#L73) but the getter for the `scale` property returns a tuple and the setter for it accepts both tuple and float (https://github.com/pythonarcade/arcade/blob/development/arcade/sprite/base.py#L310). 

I would like the `scale` argument to `__init__` to work the same way as the property setter, so it accepts both a 2-tuple and a float. 

### What would it help with?

I can easily pass the scale of one sprite to the init of a new sprite, same as I could in older versions of arcade (before the scale getter started to return a tuple).

einarf:
The `Sprite.scale` changes made the property 4-5 times slower. Instead we introduce methods to handle uniform scaling.


DragonMoffon:
Id like to try one last thing before we go ahead. Since the biggest cost with the Vec2 is around creation and GC I wonder if making a slotted murable Vec-like with support for piecewise operations would be performant enough. Like working with lists it may have a few hickups, but Ill implement `__copy__`

DigiDuncan:
> Id like to try one last thing before we go ahead. [...]

I'm curious to see this before we push this one through -- the DX is a tad clunky and it won't be forward-compatible with 3.1. I'm fine with it if Dragon's proposal doesn't get done/isn't up to snuff, though, I agree it's a better option then lying to the user as to what the underlying scale is.

einarf:
The amount of things that broke due to this change was minuscule so I'm not too worried. The new methods are also a lot faster than the scale property can be. Especially when adding and multiplying scale simply because it's one callable less compared to `+=` and `*=`. 

This also needs to be checked before merge. Ensure all subclasses of `BasicSprite` is typed to allow a tuple as well.
https://github.com/pythonarcade/arcade/issues/2399

arcade/examples/sprite_explosion_particles.py
DragonMoffon, line 1: Still has wrong method will fix soon

DragonMoffon (changes requested):

DragonMoffon (approved):

DigiDuncan (approved):
As much as the functions feel a tad clunky still, the speed can't be denied, and I'm more concerned about the return of `scale = float` for DX reasons. Seems good to me until 3.1+.

PR #2407, opened 2024-10-10
View these Examples

DragonMoffon:
SO MANY EXAMPLES

I gave up at the GL examples. I also didn't touch the platformer tutorial or any of the doc tutorials.

### DIFF
--- a/arcade/examples/gui/0_basic_setup.py
+++ b/arcade/examples/gui/0_basic_setup.py
@@ -118,6 +118,4 @@ def main():
     arcade.run()
 
 if __name__ == "__main__":
-    window = arcade.Window(title="GUI Example: Basic Setup")
-    window.show_view(GreenView())
-    window.run()
+    main()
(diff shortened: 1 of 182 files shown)


arcade/examples/gui/0_basic_setup.py
eruvanos, line 1: should the main be called?

arcade/gui/view.py
eruvanos, line 1: Why did you remove the background?
Isn't it also used in all the UI examples?

DragonMoffon: Ah because I added the background as a property of view

arcade/examples/gui/0_basic_setup.py
DragonMoffon: yeah :p

arcade/gui/view.py
eruvanos: Maybe I oversaw that ... mh .... found it, you are changing the window background_color via the property 🤔
Not sure, in that case the `on_show_view` method would have to set background everytime again.

My idea was to have a per view property, that does not require `on_show_view code`

DragonMoffon: Yeah you're right I will update View to have this behaviour

Issue #2387, opened 2024-10-04
Changing development branch breaks resource embeds in doc


pushfoo:
### Actual behavior:
Audio and music resource embeds for the doc are broken for `arcade <= 2.6.17` because resource listing embeds always reference the development branch.

### Expected behavior:
Resource embeds point to an appropriate release tag URL.

### Why this matters
**TL;DR:** Long-term reliability for Arcade release doc.

This affects:
1. The current `arcade <= 2.6.17` doc (we moved resources for 3.0)
2. The EOL 2.6.18 release discussed in the comments of #2358:
    * It exists to help existing 2.6 projects keep working
    * It will have warnings that 3.0 is imminent + links to upgrade URLs
    * It will be superseded by 3.0 shortly after
3. 3.0+ (once we release new versions)

### Steps to reproduce/example code:
1. Go to https://api.arcade.academy/en/latest/resources.html?highlight=resources#id22
2. Observe that all the sound and music embeds are broken

### Root cause
See the following line:
https://github.com/pythonarcade/arcade/blob/9a993dd744127aa628712e6352e63b702fabe46d/util/create_resources_listing.py#L21

### Proposed Fix

Update our doc build to use the following logic:

1. Detect our Arcade version
4. Does a matching git tag exist on GitHub?
   * Yes: Use a tag-based URL (Example: `https://github.com/pythonarcade/arcade/blob/2.6.17/arcade/resources/music/funkyrobot.mp3`)
   * No: default to development branch
5. Write base URL to an env variable or temp file to simplify passing data between scripts?
7. Update `create_resources_listing.py` to use the selected base URL for embeds

pushfoo:
Update:
1. There's a rough fix for the [`2.6-py-312`](https://github.com/pythonarcade/arcade/tree/2.6-py312) branch.
2. Work on the `development` will come after since:
   * It requires a bit of a different approach
   * I want to validate our changes / prepare 3.0-related doc beforehand

pushfoo:
There's still some lingering weirdness with this I need to fix.

pushfoo:
Nvm, looks like a caching issue. @ everyone else, pls confirm the page has no broken embeds on your end.

Cleptomania:
This looks fixed now

Issue #2404, opened 2024-10-09
Fix readthedocs config to stop warning that latest branch is a development branch.


pushfoo:
## Bug Report

Readthedocs now injects "helpful" JS that misleads users:
![image](https://github.com/user-attachments/assets/f2faad5b-dfb7-4fe4-bde8-499812750fbd)

We need to gain control of that.

DragonMoffon:
It also warns that the development page is out of date

pushfoo:
**TL;DR:** RTD switched on some plugins as defaults on October 7th, but I found the off switch

### Who can fix this?

Anyone with access to the new plugins control panel: @einarf, @Cleptomania, or @pvcraven.

### How can they fix this?
#### Overview
| Option | What | Why it could be     worth it |
|----------|------------------|--|
| The Good | Turn it off through the control panel | Ship 3.0 fast |
| The Bad    | Re-use their non-stable JS API | It's still injected when a project switches off the plugin |
| The Ugly   | Fuss with version number regex | It may be the best long-term option |

#### 1. The Good: turn it off fast
This'll let us move on with other 3.0-critical tasks ASAP:

1. Go to https://app.readthedocs.org/dashboard/
3. Click the Arcade project
4. Look one row down below the title
5. Follow it right until you see a "settings" button:
    ![image](https://github.com/user-attachments/assets/3374bae2-3864-46fa-a3c5-037cc983e4e2)
6. Click it
7. In the left-hand sidebar, click the `Addons` option:
    ![image](https://github.com/user-attachments/assets/7fce06fd-f764-4078-8033-6509a845547f)
8. Look to the box to the right of left-hand sidebar to find the `Notifications` tab: 
    ![image](https://github.com/user-attachments/assets/b33f2158-3eb4-4bf4-8621-8e89f914f5ff)
10. Click it
11. Uncheck this box 
    ![image](https://github.com/user-attachments/assets/9d09fd40-0773-4cda-b512-48e4b7acf112)
12. Hit save

:crossed_fingers: Hopefully, that should be it.

#### 2. Bad: re-use their JS
They inject [`notifications.js`](https://github.com/readthedocs/addons/blob/main/src/notification.js) even when a project disables version notifications in the add-ons panel. The caveat is that this probably isn't meant to be public API:
1. There's no real doc for this in the main Readthedocs documentation
2. There's no doc for it on the addons repo (see https://github.com/readthedocs/addons/)

For an example of how how the inject it:

[...]

Issue #2403, opened 2024-10-09
Verify rotation doc and fix any outdated rotation diagrams


pushfoo:
### Actual behavior:
At least some rotation diagrams are out of date:
![image](https://github.com/user-attachments/assets/426905ed-7996-45a7-89e6-5f8d5563a8c2)

### Expected behavior:
Rotation in doc matches rotation in code.

Issue #2325, opened 2024-07-28
Minimal doc changes for 3.0


einarf:
Minimal tasks for 3.0 release. Some of these can be done incrementally
- [x] Do a pass on front page
- [x] Fill out the about section
- [x] Fill out the contributing section
- [x] Fill out the community section
- [x] Resolve differences between arcade book and the skill tree
- [x] Organize the api references. These needs to be put in categories one level down
- [x] Organize the manual

Still, moving things can create future issues when users move between doc versions. We should try to establish structures as soon as possible.

pushfoo:
## WIP notes

### Community section

https://api.arcade.academy/en/development/about/intro.html#community-driven

1. (@pushfoo) Rephprase this paragraph
2. Put links first in the list as calls to action

https://api.arcade.academy/en/development/contributing/index.html#how-to-contribute

1. Turn learning resource list into links

https://api.arcade.academy/en/development/contributing/index.html#how-to-contribute-without-coding

3. Table the history links?

https://api.arcade.academy/en/development/community/community_locations.html#community-locations
1. `s/discord server/Discord server/g` (it's a brand)

Q: Move "how to get help" into "Getting Started"?

https://api.arcade.academy/en/development/community/games/sample_games.html

1. Fix example game page links (some are markdown instead of ReST )

### Book + Skilltree Page

> Resolve differences between arcade book and the skill tree

I started taking a look at this. My initial impression is that we need to split + rename this page as follows:

1. Rename it to Introduction
5. Move it up the TOC tree to be above Install
6. Remove mentions of / references to the book here because it won't be 3.0 compatible for a long while
   * Add a `.. note::` for educational users who need long-term support which links a page on 2.6.18 vs 3.0
   * Replace the `[Learn Arcade book sound chapter](https://learn.arcade.academy/en/latest/chapters/20_sounds/sounds.html)` link with the programming guide sound page
   * Fix the sound demos:
      * Add on-screen labels of some sort so it's not just squares 
      * Add arrows ( can be text-based )
   * Fix inconsistent physics engine formatting 
7. Move the skill tree to a new `Skill Tree` page

The 2.6.18 vs 3.0 page can live on either site (3.0 or 2.6.X). Here's a rough concept for a plan:

- [ ] I do some 2.6.18 polish on:
   * A "compatibility maintenance-only" page
   * Our own warnings for RTD
- [ ] We ship 2.6.18 / etc
- [ ] We update the doc for whichever versions need to point to it wherever

PR #2410, opened 2024-10-12
Fix video embeds, columns, and resource link breakage

pushfoo:
**TL;DR:** Fixes most if not all of #2387 for `development` + add video embeds plus some copyable string literals

### Changes
There are a whole bunch.

1. Most of (if not all) of #2387 is covered, but needs a thorough check before closing the issue
2. When N < 3 for a file category, reduce the number of columns
3. Add video embedding
4. Add preformatted `":resources:sounds/jump3.wav"` literals for sounds and videos
5. Add format -> codec string tables to clean up sound and video embeds

### Follow-up work
Switch to a copy-based approach for true long-term durability.


pushfoo:
Per einarf's feedback:
1. the optimal solution may be to copy our resources files into our build
2. our built-in assets total 20 megs per his estimate, but we should check
I'm currently looking at new CI failures on #2412 in doc build of all things.

pushfoo:
**TL;DR:** Feedback from Digi is incorporated, but perfect file copying will follow later since there's loads more stuff to do

Getting folder copying "right" is less important than:
| Done? | Item | Why it's more important |
|---------|----|----------------------------------|  
| :heavy_check_mark:  |  #2403 | Doc ~~may be outright wrong~~ is now fixed |
| Pending admin account override | #2404 | RTD notifications mislead users |
|  | #2325 | Basic usability |

We also have the following upstream changes to take care of now when it's ok to break API:
| pyglet text issue | how it's broken |
|-----------------------|----------------------|
| https://github.com/pyglet/pyglet/issues/1203 | Tracking and kerning should add up for individual character adjustments |
| Font weight typing stuff in pyglet | Broken typing |

PR #2430, opened 2024-10-21
Fix math point types

Rapen765:
Changed functions' return type from `tuple[float, float]` to `Point2`.
Fixed documentation of `rand_vec_magnitude` and `rand_on_circle`.

### DIFF
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -306,7 +306,7 @@ def rand_vec_magnitude(
     Returns a random vector, with random magnitude.
 
     Args:
-        angle (float): The vector angle
+        angle (float): The vector angle in radians
         lo_magnitude (float): The lower magnitude
         hi_magnitude (float): The higher magnitude
     """
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -303,7 +303,7 @@ def rand_vec_magnitude(
     hi_magnitude: float,
 ) -> Point2:
     """
-    Returns a random vector, with random magnitude.
+    Return a vector of randomized magnitude pointing in the given direction.
 
     Args:
         angle (float): The vector angle in radians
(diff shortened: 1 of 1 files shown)


arcade/math.py
```diff
@@ -304,12 +301,12 @@ def rand_vec_magnitude(
     angle: float,
     lo_magnitude: float,
     hi_magnitude: float,
-) -> tuple[float, float]:
+) -> Point2:
     """
-    Returns a random vector, within a spread of the given angle.
+    Returns a random vector, with random magnitude.
```
pushfoo:
```suggestion
    Return a vector of randomized magnitude pointing in the given direction.
```

arcade/math.py
```diff
@@ -304,12 +301,12 @@ def rand_vec_magnitude(
...
 
     Args:
-        angle (float): The angle to spread from
+        angle (float): The vector angle
```
pushfoo:
```suggestion
        angle (float): The vector angle in radians
```

pushfoo (changes requested):
**TL;DR:** I'm undecided on the `Point2` change, but I have suggestions on the doc and the doc

@DragonMoffon any thoughts on consistency for `Point2` vs the perf dangers of `Vec2`?

@Rapen765 The CI's linter complaints. Could you please try the following in your local repo clone?
1. Make sure your venv's activated
2. `./make.py format && ./make.py ruff`
3. Commit & push

DragonMoffon:
I def think Point2 should be used over tuple[float, float] everywhere. The only place it shouldn't be used is anything that returns a `tuple[int, int]` that should have a different alias (`Size2` maybe)

pushfoo (approved):

pushfoo:
Merging to nudge velocity back up for 3.0.

pushfoo:
Ty for the contribution and sorry for the wait!

PR #2434, opened 2024-10-24
Promote sprite depth examples from experimental to main

pushfoo:
1. Stop trying to set the window's read-only clock value
2. Convert them to the new `View` example style
3. Add entries for them in the examples listing


arcade/experimental/sprite_depth_cosine.py
```diff
@@ -88,11 +87,10 @@ def on_key_press(self, symbol: int, modifiers: int):
             self.text_use_depth.text = f"SPACE: Toggle depth testing ({self.use_depth})"
 
     def on_update(self, delta_time):
-        self.time += delta_time
-
         for i, sprite in enumerate(self.sprite_list):
+            # time ticks automatically now due to Window's global clock
```
DragonMoffon: We have to assume most users will have never looked at this example before so this comment doesn't make sense. (Also this might not have existed pre 3.0 so even less likely to have been seen before)

DragonMoffon (changes requested):
The window has a depth buffer by default, and spritelists innately support depth testing so this should be moved to `examples`

alejcas:
Seems like this line is wrong:

`if self.use_depth:`

You moved `use_depth` to the window so the line should check:

`if self.window.use_depth:`

Anyway I don't understand why you moved that attribute to the window. Just leave it in the view...no?

alejcas:
> Seems like this line is wrong:
> 
> `if self.use_depth:`
> 
> You moved `use_depth` to the window so the line should check:
> 
> `if self.window.use_depth:`
> 
> Anyway I don't understand why you moved that attribute to the window. Just leave it in the view...no?

I think this is what's wrong:

`window.use_depth = not window.use_depth`

arcade/experimental/sprite_depth_cosine.py
pushfoo: Removed.

pushfoo:
@alejcas fixed

arcade/examples/depth_of_field.py
```diff
@@ -171,7 +178,7 @@ def render(self):
         self._geo.render(self._render_program)
 
 
-class App(Window):
+class App(View):
```
DragonMoffon: convention is `GameView`. Even though it's not really accurate, the examples used to be a mess in this regard, so im forcing a naming convention.

arcade/examples/depth_of_field.py
```diff
@@ -234,11 +242,26 @@ def on_draw(self):
         with self.dof.draw_into():
             self.sprites.draw(pixelated=True)
 
-        # Draw the blurred frame buffer and then the focus display
-        self.use()
+        # Draw the blurred frame buffer and then the focus display\
```
DragonMoffon: accidental slash

arcade/examples/sprite_depth_cosine.py
```diff
@@ -23,21 +23,22 @@
...
 SPRITE_X_START = 150
 SPRITE_X_STEP = 50
-SPRITE_Y = HEIGHT // 2
+SPRITE_Y = WINDOW_HEIGHT // 2
 
 DOT_SIZE = 10
 
 
-class MyGame(arcade.Window):
+class DepthExample(arcade.View):
```
DragonMoffon: Same naming note as above. post 3.0, we could discuss a different convention, but for now...

arcade/examples/sprite_depth_cosine.py
```diff
@@ -88,11 +89,26 @@ def on_key_press(self, symbol: int, modifiers: int):
             self.text_use_depth.text = f"SPACE: Toggle depth testing ({self.use_depth})"
 
     def on_update(self, delta_time):
-        self.time += delta_time
-
+        # We absolute time instead of the delta, so we get it from the window's clock
```
DragonMoffon: This might be github, but `we absolute time` doesn't make sense.

DragonMoffon (changes requested):

einarf:
The reason it wasn't moved into examples was a lingering issue related to the depth range. As far as I know the depth range set in the ortho projection is `[-1.0, 1.0]` defining the `depth` that can be set in the sprite's geometry.

We could possibly change to to `[-1, 100]` to give the user a more reasonable range to work with? If the sprite depth is set outside this range it risks being clipped in primitive assembly not appearing.

Including a new reasonable depth range would make things a bit more cleaner?

einarf:
It seems @DragonMoffon changed the range to `[-100, 100]` so we should make sure to mention this range in the example and take advantage of this range if that's not already in place. Seems it was modified a bit from the original -1 to 1 range.

The default depth range in arcade is -100 (near plane) / 100 (far plane). Higher value is further away from the camera.

pushfoo:
Can we add default near and far constants to make this more legible?

einarf:
For now I think it's enough to just mention the default range.

DragonMoffon:
I didn't actually choose this range myself. There was definitely a code snippet in Arcade that I based it on. I feel like it was the internal matrix gen, but it might have been one of the gl tutorials.

In any case, in the depth example, it should probably be mentioned since it's directly settable in Camer2D's init

arcade/examples/sprite_depth_cosine.py
```diff
@@ -88,11 +89,26 @@ def on_key_press(self, symbol: int, modifiers: int):
             self.text_use_depth.text = f"SPACE: Toggle depth testing ({self.use_depth})"
 
     def on_update(self, delta_time):
-        self.time += delta_time
-
+        # We use absolute time instead of the delta, so we get it from the window's clock
```
DragonMoffon: I would still change this to 'elapsed' since that's the language used by the clocks. It is also missing a 'the'

'We use the elapsed time instead of the delta, so we get it from the window's clock'

DragonMoffon (approved):
One last, very pedantic, change. Im otherwise happy for it to be merged.

arcade/examples/sprite_depth_cosine.py
einarf: Just use the time from the clock. There are several examples can could need this change.

pushfoo:
- [x] Added doc about near / far clipping
- [x] Changed time phrasing to be lessspecific to American / Hemmingway style
- [x] Centralized near / far defaults from a single source of truth

pushfoo:
> ["yeah looks good" - DragonMoffon](https://discord.com/channels/458662222697070613/458662458198982676/1299560812221825054) (Discord)

PR #2446, opened 2024-11-12
Fix: issue-2429-Add_direction_to_ScreenShake2D

ouad002:
The fix addresses an enhancement request for the **ScreenShake2D** class in the **Arcade** library by introducing an optional **direction_deg parameter**. This allows users to specify a constant direction for the screen shake effect, expressed in degrees. If the **direction_deg** parameter is not provided, the shake continues to occur randomly, as before. The fix ensures that when **direction_deg** is set, the shake occurs along a fixed angle, providing more control over the shake's direction while maintaining the original random behavior when the parameter is not used.

### DIFF
--- a/arcade/camera/grips/screen_shake_2d.py
+++ b/arcade/camera/grips/screen_shake_2d.py
@@ -40,6 +40,9 @@ class ScreenShake2D:
         shake_frequency:
             The number of peaks per second. Avoid making it a multiple of half
             the target frame-rate. (e.g. at 60 fps avoid 30, 60, 90, 120, etc.)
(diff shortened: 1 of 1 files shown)


DragonMoffon:
The fact that this removes important methods, the double definitions of the `stop` function, the strange rewriting of docs, and the lack of style following makes me suspicious that this was, at least in part, written by generative AI. We appreciate every PR made into Arcade. However, this PR needs heavy revision before I can accept it.

DragonMoffon:
Closing in favor of #2448

PR #2450, opened 2024-11-14
Pull Request to Add Optional view Argument in Window.run()

joaomarcalst:
PR Description

This pull request aims to add an optional view argument to the Window.run() and arcade.run() methods. This feature is designed to improve developer experience by promoting the use of subclassed views instead of direct window manipulation.

Implementations and Observations
->  Addition of view Argument:
- With this PR, the run() method now accepts an optional view argument and will automatically call Window.show_view.

-> Testing:
- Unfortunately, the full implementation of tests was affected by issues with the pyglet module, specifically regarding the import of pyglet.display.base.
- The required function to initialize the pyglet environment encountered module errors, preventing complete test execution.
- 
-> Suggestions for Future Implementations:
- While this PR provides a functional base, I suggest further investigation into compatibility with pyglet to complete the testing process.

Note

For technical details and context, refer to [Issue #2400](https://github.com/pythonarcade/arcade/issues/2400), where challenges and initial steps for this PR were discussed.

PR #2452, opened 2024-11-14
Add optional view argument to Window.run() and arcade.run()

Marouane743:
This PR adds an optional view parameter to Window.run() and arcade.run(), allowing users to specify a view to display when starting the main event loop. This improves flexibility in displaying views directly.


einarf:
There is another PR currently being reviewed resolving the same issue: https://github.com/pythonarcade/arcade/pull/2450

einarf:
Closing this in favor of #2452

PR #2453, opened 2024-11-14
modified (point_in_view) function

maphineth:
I modified the point_in_view method to improve its clarity and readability. Currently, it calculates the dot product manually and uses intermediate variables. We can simplify it using vectorized operations to make it more concise and efficient.

Explanation:
- Vec2 Class: I assumed that Vec2 supports vector operations like dot().
- Readability: The right vector is calculated to represent the camera's horizontal axis, perpendicular to up.
- Clarity: The dot() method is used to project the diff vector onto the right and up vectors, making the math cleaner and more intuitive.
- Efficiency: Vectorized operations make the code easier to read and maintain while ensuring mathematical operations remain efficient.

### DIFF
--- a/arcade/camera/camera_2d.py
+++ b/arcade/camera/camera_2d.py
@@ -477,23 +477,23 @@ class Camera2D:
         return LRBT(left=left, right=right, bottom=bottom, top=top)
 
     def point_in_view(self, point: Point2) -> bool:
-        # TODO test
         """
-        Take a 2D point in the world, and return whether the point is inside the
-        visible area of the camera.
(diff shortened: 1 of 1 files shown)


DragonMoffon:
This PR has been closed as it makes no functional or performance improvements to the `point_in_view` method. It also reads like it was written using LLM. While we appreciate all PR's made into Arcade, we require them to be authored and written by people.

einarf:
Also: The existing method is optimized. Introducing Vec2 will make it 10x slower. "High performance" in python is a completely different beast and needs careful human crafting.

Issue #2454, opened 2024-11-15
Make pymunk excludable during os packaging


eruvanos:
## Enhancement request:
Building a game to distribute it, allows to exclude unused packages.
While Pymunk might not be used in the game, the occurring import errors make it unable to be excluded.

### What should be added/changed?

- catch import errors in [arcade.__init__](https://github.com/pythonarcade/arcade/blob/575d6621c10bc42e32a08c04abe7f72b2b4e16c2/arcade/__init__.py#L202)
- Use "MissingImportAlgo" within [hitbox](https://github.com/pythonarcade/arcade/blob/575d6621c10bc42e32a08c04abe7f72b2b4e16c2/arcade/hitbox/__init__.py#L15-L15C40) if pymunk is not available

DragonMoffon:
I've never actually faced this issue using Nuitka, what packaging library did you use?

eruvanos:
I used cx_freeze.
Did you exclude pymunk in your build?

DragonMoffon:
Ah maybe not

eruvanos:
@DragonMoffon could you take a look? 
https://github.com/pythonarcade/arcade/pull/2458

einarf:
Status on this. Any forward plan or thoughts?

einarf:
Possibly we need support for lazy loading to get started on this?

Cleptomania:
I finished this as part of the browser support/webgl backend. Pymunk is now in an `extras` dependency group, and you will only get errors if you actually import and use PymunkPhysicsEngine without it installed.

PR #2458, opened 2024-11-18
Pymunk import

eruvanos:
Fix for: #2454

This PR makes pymunk import optional and just prints the following during start.

```
Warning: pymunk is not installed. PymunkHitBoxAlgorithm will not be available.
Pymunk is not installed. Pymunk physics engine will not be available.
```

### DIFF
--- a/.github/workflows/selfhosted_runner.yml
+++ b/.github/workflows/selfhosted_runner.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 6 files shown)


DragonMoffon:
So this looks good, but there are a few other places that reference pymunk, and because the pymunk objects may not be available, type hints and code completion might get a little confused. 

I don't think the code i wrote is particularly good, but i do have a branch where i also tried this you could compare with.

DragonMoffon:
I think the first step to doing this properly is to actually quarantine all pymunk objects to a pymunk submodule and then create a 'NotImported' object that raises an import error in '__getattr__', '__call__' etc.

eruvanos:
@DragonMoffon is right, so we drop this PR

PR #2468, opened 2024-12-05
Add View Argument to Window.run()

### LINKED ISSUE #2400: Add optional `view` argument to `Window.run()' and `arcade.run()`
To help promote subclassing views over the window itself, we should add an optional view argument to the run methods, letting them automatically call `Window.show_view`.

It results in only a one-line change, but it makes the DX (Developer Experice) nicer.

It will also need a unit test for each method, and the examples will need updating to take advantage of the change.

EmmettBicker:
Fixes #2400 

I was looking at the Good First Issue tag and I found this one! Most of the work was already done in #2450 but that pr seems to have been abandoned, so I just applied the requested changes in here.

Please let me know if there's any issues :). This seems like a wonderful repo that I'd love to contribute to.

### DIFF
--- a/arcade/window_commands.py
+++ b/arcade/window_commands.py
@@ -16,7 +16,7 @@ from arcade.types import RGBA255, Color
 
 if TYPE_CHECKING:
     from arcade import Window
-
+    from arcade.application import View
 
 _window: Window | None = None
 
@@ -97,16 +97,22 @@ def close_window() -> None:
     gc.collect()
 
 
-def run():
+def run(view: View | None = None):
     """
-    Run the main loop.
+    Run the main loop. Optionally start with a specified view.
 
     After the window has been set up, and the event hooks are in place,
     this is usually one of the last commands on the main program.
     This is a blocking function starting pyglet's event loop meaning
     it will start to dispatch events such as ``on_draw`` and ``on_update``.
+
+    Args:
+            view: The view to display when starting the run. Defaults to None.
     """
     window = get_window()
+
+    if view is not None:
+        window.show_view(view)
 
     # Used in some unit test
     if os.environ.get("ARCADE_TEST"):
--- a/arcade/window_commands.py
+++ b/arcade/window_commands.py
@@ -97,7 +97,7 @@ def close_window() -> None:
     gc.collect()
 
 
-def run(view: View | None = None):
+def run(view: View | None = None) -> None:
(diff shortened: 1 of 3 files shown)


einarf:
I was about to tag you in this issue, but you beat me :)

EmmettBicker:
Haha, I was looking through the issues and it definitely seemed like a good place to start! Just joined the discord and am excited to contribute more

arcade/window_commands.py
```diff
@@ -97,17 +97,23 @@ def close_window() -> None:
...
-    Run the main loop.
+    Run the main loop. Optionally start with a specified view.
 
     After the window has been set up, and the event hooks are in place,
     this is usually one of the last commands on the main program.
     This is a blocking function starting pyglet's event loop meaning
     it will start to dispatch events such as ``on_draw`` and ``on_update``.
+
+    Args:
+            view: The view to display when starting the run. Defaults to None.
```
einarf: I'm not sure if sphinx-napoleon likes the additional 4 spaces here

arcade/window_commands.py
```diff
@@ -97,17 +97,23 @@ def close_window() -> None:
     gc.collect()
 
 
-def run():
+def run(view: View | None = None):
```
einarf: Might as well add a `None` return for this function while we are here

einarf (changes requested):

arcade/window_commands.py
```diff
@@ -16,7 +16,7 @@
 
 if TYPE_CHECKING:
     from arcade import Window
-
+    from arcade.application import View
```
einarf, line 19: This one should ideally be further up with the other imports. Then do a `python make.py format` to ensure isort is happy.

EmmettBicker:
On it! I only ran make.py lint and make.py test, I'll make sure to read the contributing doc more thoroughly

arcade/window_commands.py
EmmettBicker: When I tried this it caused a circular import error because application imports from window_commands

EmmettBicker: Wait whoops that was the other file! I'll change this

EmmettBicker: Oh actually I think it was this one

einarf (approved):

Issue #1135, opened 2022-03-27
Bundle a ttf for Arcade's default font


einarf:
Right now Arcade's pyglet based text are using the following default:  `("calibri", "arial")`. This will just scan the system fonts and find something that matches. This will give the user fairly different looks depending on the system and OS.

Instead we should look into bundling a ttf file with a default font for Arcade. This should be something fairly close to the current default font (s):

In a game setting this makes a lot more sense. System fonts are more for generic applications.

We can and probably should add an `ATTRIBUTIONS.md` file in the root of the repo anyway. That might be needed here. It should also be added to the package in `MANIFEST.in`

pushfoo:
This can probably be handled quickly using #1011 or a similar approach, and is arguably a duplicate of that ticket. There are metric compatible fonts for Arial, so we can get cross-platform Arial equivalents fairly easily. It may also be a good idea to include license files for the fonts as well as README files in their respective folders.

einarf:
Budle a default font and define it in `arcade.fonts.DEFAULT` + add all the kenney font names we are loading on startup.

For game development in general it's all about custom fonts.

pushfoo:
**Update / recap of today's discord discussion:**

This issue requires the following to resolve:
1. Decide which languages / character sets are a priority to support
2. Choose a font that includes the needed characters and has a favorable license (probably SIL Open Font, Public domain, or Apache)
3. Triple check the licensing situation, possibly with a legal professional

I started addressing item 1 in the doc analytics channel today.

Item 2 has been in progress on-and-off since I filed #1011. Although the Liberation font family seemed like the best bet, there could be issues with it:
* Earlier versions use a GPL variant that might be infectious if bundled with software
* Liberation 2.0.0+ fonts appear use the SIL Open Font license, which should be ok
* It's unclear what languages Liberation fonts support

eruvanos:
We have Kenny fonts bundled and I set them as default for buttons, (and they are shown if loaded)

Should we have another font, that is always be loaded?

We could pick a few fonts from the following pages (with a proper license) and then let the community vote:
- https://www.dafont.com/
- https://fonts.google.com/
- https://www.dafont.com/de/bitmap.php?l[]=10

pushfoo:
As I've said before: https://github.com/liberationfonts/liberation-fonts

They're:
1. SIL Open Font Licensed
[...]

Issue #1011, opened 2021-10-15
Add 3 SIL Open font license "default" fonts & import them on startup


pushfoo:
### What should be added/changed?

A beginner-friendly way for users to keep text consistent across operating systems

#### Core Enhancement

tl;dr An easy way to achieve this would be do the following:
1. Include three metric-compatible SIL open source license fonts for the Serif, Sans Serif, and Monospace font archetypes
2. Load them as the default constants for font archetypes.

For example, to allow users to draw text in identical Times-like fonts on all platforms, we could provide the following constant:
```py
# Each of these font names is a drop-in for the others
SERIF = (
    "Times New Roman",
    "Times",  # Present on Mac OS
    "Liberation Serif",  # V 2.0 onward is reportedly SIL open font license
)
```

Drawing with font constants would be consistent with the way we already use color constants:
```py
# The beginner-friendly way
arcade.draw_text(
   "This text should look almost identical on every platform",
   WIDTH / 2, HEIGHT / 2,
   font_name=arcade.fonts.SERIF,
   font_color=arcade.colors.RED
)

# Object oriented
text_instance = arcade.Text(
   "This text should look almost identical on every platform",
   WIDTH / 2, HEIGHT / 2,
   font_name=arcade.fonts.SERIF,
   font_color=arcade.colors.RED
)
```

Since the font named in `SERIF` are [metric-compatible with Times New Roman](https://practice.typekit.com/reference/Metric-compatible-fonts/), using the constant should produce extremely similar results across most platforms, regardless of any styling parameters passed. 

#### Expanding The Idea

Linux distros appear to use a variety of metric compatible font sets to provide replacements for common fonts, while macOS appears to use a mix of custom metric-compatible fonts and licensing the original (this might vary by OS version).

The following font archetypes are covered by most metric compatible font sets, so they are probably the best starting point for this ticket:
- Serif (Times New Roman)
- Sans Serif (Arial)
- Courier

To provide a consistent interface for all cross-platform fonts, we could also include constants that reference the Kenney fonts already in the resources folder. This way, we would have both:
- an easy interface for cross-platform fonts
- a decent selection of fonts for users to style their games with

#### Making it Reliable

To ensure that a specified style will **always** be found regardless of system fonts, we can do the following:
1. Include TTF copies of a metric compatible font set in the resources folder as a fallback
2. Add `f":resources:\fonts\{METRIC_COMPATIBLE_NAME}"` at the end of each of the font constant tuples

[...]

PR #2477, opened 2025-01-04
Add Liberation font family to Arcade

pushfoo:
**TL;DR:** Add Liberation font files per #1011 and #1135 per call with Digi and Eruvanos

### Changes
1. Add the OFL-licensed Liberation font family:
   * It's like a font file limited version of  CC: Attribution, Share-Alike for fonts
   * It doesn't spread beyond the font files
2. Move Kenney fonts into their own folder
3. Add a loading helper
4. Fix typos
5. Add comments

### Questions
1. This doesn't set them as default yet, but should it?
2. Which ones are we making the "default" Arcade font if any? (Liberation Sans?)


DigiDuncan:
I think this PR is fine, but I do like the idea of Liberation Sans being the default font.

pushfoo:
The resources listing is (once again) broken because it assumes a lot about how Arcade is built. I don't have time to fix this tonight and might not in the next few days.

pushfoo:
I've rebased this and I'm looking at an acceptable short-term fix to get 3.0 moving.

arcade/resources/__init__.py
```diff
@@ -230,14 +237,36 @@ def load_kenney_fonts() -> None:
...
+    load_font(":system:fonts/ttf/Kenney/Kenney_Pixel_Square.ttf")
+    load_font(":system:fonts/ttf/Kenney/Kenney_Rocket.ttf")
+    load_font(":system:fonts/ttf/Kenney/Kenney_Rocket_Square.ttf")
+
+
+def load_liberation_fonts() -> None:
+    """Loads generic versions of Arial, Courier, and Times New Roman.
+
+    The Liberation font family is a permissively-licensed set of built-ins
+    which should be layout-compatible with popular fonts.
```
eruvanos: Can we add the names, which should be used to reference the loaded fonts? That was an issue for me with the Kenny fonts.

pushfoo: Yep. Will do.

pushfoo: I've added it here as requested, but I believe it may go best on the resources page:
image

eruvanos:
I would be fine with merging it, looks good to me and as far as I understood it, we can go with the pre-generated image for the fonts, that is fine for now.

PR #2487, opened 2025-01-10
Upgrade to pyglet==2.1.0

pushfoo:
EDIT: don't merge till this is merged https://github.com/pyglet/pyglet/pull/1254
1. Incorporates cherry-picked changes from #2475 
2. Bumps the pyglet version to `pyglet~=2.1.0`

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -60,10 +60,11 @@ else:
 import pyglet
 
 # Enable HiDPI support
-if os.environ.get("ARCADE_TEST"):
-    pyglet.options.dpi_scaling = "real"
-else:
-    pyglet.options.dpi_scaling = "stretch"
+# Note: This seems to have changed to 'scale_with_dpi' in pyglet
+# if os.environ.get("ARCADE_TEST"):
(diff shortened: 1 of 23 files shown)


arcade/__init__.py
```diff
@@ -60,10 +60,11 @@ def configure_logging(level: int | None = None):
...
 # Enable HiDPI support
-if os.environ.get("ARCADE_TEST"):
-    pyglet.options.dpi_scaling = "real"
-else:
-    pyglet.options.dpi_scaling = "stretch"
+# Note: This seems to have changed to 'scale_with_dpi' in pyglet
+# if os.environ.get("ARCADE_TEST"):
+#     pyglet.options['dpi_scaling'] = "real"
+# else:
+#     pyglet.options.dpi_scaling = "stretch"
```
DragonMoffon: I will do a fuller review later, but this has to be fixed as part of the merging of pyglet 2.1.

eruvanos: With out that the window is 1/4 of the normal size

einarf: This section definitely cannot be removed.

PR #2488, opened 2025-01-10
Fix things that mypy points out are wrong

pvcraven:


eruvanos:
Should be covered with the 2.1 merge

DragonMoffon:
I've just double checked all of the changes this PR makes, and they are already in development so I'm closing this PR.

PR #2489, opened 2025-01-10
fix type issues

eruvanos:
This should fix all mypy issues, I used some ignores and some things might not be the best solution.

Works on my machine:
image


einarf:
As far as I can see most or all of this was moved into #2487

Issue #2499, opened 2025-01-21
Pitfall with `prevent_dispatch` arg in `Section` class


ghost:
While developing using 3.0dev, I misread the documentation and passed in `prevent_dispatch=False` to allow events to pass to multiple sections. Because of my mistake, I spent 30 minutes trying to figure out why I'm not seeing any events in my other section.

This init code:
https://github.com/pythonarcade/arcade/blob/9aa8e1a4c1afca13855e13bd6a901edb02e4ee90/arcade/sections.py#L99

Means that if the caller passes in `False`, it silently gets translates to `{true}`, which is not intuitive.

I realize this is documented and also that mypy would have caught it, but I fell for it so others probably would too. Since 3.0 is still in beta there's still the possibility of changing the API so I thought I might as well raise this point. 

To me, It also seems natural for ''True|False" values to do what you'd expect, but I understand typing simplicity makes this less attractive.

Possible solutions:
1. Use `{True} if prevent_dispatch is None else prevent_dispatch` instead
2. Add asserts on input types.

DragonMoffon:
I think this behaviour is just flawed in general thank you for raising the issue. Sections are going to be reworked in a future update (after 3.0) so finding all these hiccups is a huge help. I'm not sure why this wasn't changed when the issue was documented.
I've updated it to use the line you recommended.

ghost:
if `prevent_dispatch=False`, then `prevent_dispatch` will remain a bool. Doesn't the downstream code expect it to be an Iterable?

DragonMoffon:
yeah, you are correct I'm still trying to figure out if there is a way to solve this issue without using an `isinstance` check. Admittedly this is only in initialization so it might not be that significant.

alejcas:
> Means that if the caller passes in `False`, it silently gets translates to `{true}`, which is not intuitive.

Is defined as an Iterable. A bool is not an iterable. 

But yes user may expect to pass False or True

pushfoo:
@AnonymousCoward128746 let us know if the `isinstance` checks are too slow.

Recap of the thread here + discussion with Dragon on Discord in the 3.0 release thread:
1. If it's only creation-time for the `Section` it's probably fine
   * `isinstance` can be slow
   * lots of real-time section creation is not a good idea in general
2. There are optimization tricks we can dust off and try later after 3.0 ships

ghost:
Thanks @pushfoo. Works for me. 👍

PR #2503, opened 2025-01-23
A few quick fixes

DragonMoffon:
Fixes:
* Threaded Loading Example Fighting over `time` def during unit tests
* (#2499)

### DIFF
--- a/arcade/sections.py
+++ b/arcade/sections.py
@@ -100,7 +100,9 @@ class Section:
         assert isinstance(self.prevent_dispatch, Iterable)
         """prevents events to propagate"""
 
-        self.prevent_dispatch_view: Iterable = {True} if prevent_dispatch_view is None else prevent_dispatch_view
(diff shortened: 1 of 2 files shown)


arcade/sections.py
```diff
@@ -64,8 +64,8 @@ def __init__(
         name: str | None = None,
         accept_keyboard_keys: bool | Iterable = True,
         accept_mouse_events: bool | Iterable = True,
-        prevent_dispatch: Iterable | None = None,
-        prevent_dispatch_view: Iterable | None = None,
+        prevent_dispatch: Iterable | bool | None = None,
```
ghost, line 67: defval can be set to `{True}` now, and then L99 just becomes
`self.prevent_dispatch = prevent_dispatch`

same for _view. This would break existing code which explicitly passes in None (can't imagine such).

  Also maybe `bool | Iterable` vs. `Iterable | bool`  should be consistent for readability.

Thanks!

PR #2512, opened 2025-01-24
Update color.py

### LINKED ISSUE #2511: arcade.Color can't be unpickled
`arcade.Color` implements a custom `__new__`. According to the `pickle` documentation on [`__getnewargs_ex__()`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__) and [`__getnewargs__()`](https://docs.python.org/3/library/pickle.html#object.__getnewargs__), `arcade.Color` should implement `__getnewargs__()` to enable un/pickling.

elegantiron:
adds `__getnewargs__()` so Color can be un/pickled

Fixes #2511

### DIFF
--- a/arcade/types/color.py
+++ b/arcade/types/color.py
@@ -140,6 +140,9 @@ class Color(RGBA255):
         # https://github.com/python/mypy/issues/8541
         return super().__new__(cls, (r, g, b, a))  # type: ignore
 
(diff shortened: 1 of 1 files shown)


pvcraven:
Looks like 'black' needs to be run on `arcade/arcade/arcade/types/color.py`

elegantiron:
Updated the formatting

PR #2526, opened 2025-01-27
Update application.py

### LINKED ISSUE #2525: arcade.View hints the wrong type for height and width
The `View.width` and `View.height` properties both claim to be aliases for `Window.width` and `Window.height` respectively. They also directly return those properties. `View.width` and `View.height`, however, have type hints for `float` return types.

https://github.com/pythonarcade/arcade/blob/cdd6f66aef330856a84446f8a34d176b7e66c2f5/arcade/application.py#L1552-L1564

The definitions for `Window.width` and `Window.height`, however, hint that they return `int` types.

https://github.com/pyglet/pyglet/blob/ce6889687a4e99136d38fcd63a01b5b350ebd5ac/pyglet/window/__init__.py#L1218-L1234

```python
    @property
    def width(self) -> int:
        """The width of the window, in pixels.  Read-write."""
        return self.get_size()[0]


    @width.setter
    def width(self, new_width: int) -> None:
        self.set_size(new_width, self.height)


    @property
    def height(self) -> int:
        """The height of the window, in pixels.  Read-write."""
        return self.get_size()[1]


    @height.setter
    def height(self, new_height: int) -> None:
        self.set_size(self.width, new_height)
```

The return type hints for `View.width` and `View.height` should be updated to reflect that they return `int`s

elegantiron:
fixes type hinting to match the returned properties

fixes #2525

### DIFF
--- a/arcade/application.py
+++ b/arcade/application.py
@@ -1543,21 +1543,21 @@ class View:
         pass
 
     @property
-    def size(self) -> tuple[float, float]:
+    def size(self) -> tuple[int, int]:
         """
         An alias for `arcade.Window.size`
         """
         return self.window.size
 
     @property
-    def width(self) -> float:
+    def width(self) -> int:
         """
         An alias for `arcade.Window.width`
         """
         return self.window.width
 
     @property
-    def height(self) -> float:
+    def height(self) -> int:
         """
         An alias for `arcade.Window.height`
         """


pushfoo (changes requested):
Although it may seem counter-intuitive, the float option may be more correct:
 
1. Mac uses float pairs for their screen locations ([example](https://developer.apple.com/documentation/swiftui/scene/defaultsize(width:height:)))
2. Other platforms may as well

This affects future mobile and web support plans as well, but it's a long-standing issue which affects pyglet.

elegantiron:
> Although it may seem counter-intuitive, the float option may be more correct:

*Personally*, I just want my type checker to chill out; I don't care which is more correct.

If float is the more correct one, then should we change the `__init__` functions to type hint that they accept floats?

I'm getting the following from pylance:

```
Argument of type "float" cannot be assigned to parameter "width" of type "int" in function "__init__"
  "float" is not assignable to "int"
```

And it is complaining about the following code:

`self.title_section = TitleSection(0, 0, self.width, self.height)`

This is a view attempting to create a section that is the same size as itself, but Pylance is unhappy that the types don't match.

eruvanos:
Could we just allow float or int, both would work right?

elegantiron:
That would solve my personal issue with it (type checking), and if it works for the technical side too I'm all for it.

elegantiron:
I'm going to open a new pull request with changes to `sections.py` which hint that it accepts `int` or `float`

PR #2527, opened 2025-01-28
Update section __init__ type hints

### LINKED ISSUE #2525: arcade.View hints the wrong type for height and width

elegantiron:
Update left, bottom, width, and height to hint that they accept int or float

fixes #2525

### DIFF
--- a/arcade/sections.py
+++ b/arcade/sections.py
@@ -116,19 +116,19 @@ class Section:
         # section position into the current viewport
         # if screen is resized it's upto the user to move or resize each section
         # (section will receive on_resize event)
-        self._left: int = left
-        self._bottom: int = bottom
-        self._width: int = width
-        self._height: int = height
-        self._right: int = left + width
-        self._top: int = bottom + height
+        self._left: int | float = left
+        self._bottom: int | float = bottom
+        self._width: int | float = width
+        self._height: int | float = height
+        self._right: int | float = left + width
+        self._top: int | float = bottom + height
 
         # section event capture dimensions
         # if section is modal, capture all events on the screen
-        self._ec_left: int = 0 if self._modal else self._left
-        self._ec_right: int = self.window.width if self._modal else self._right
-        self._ec_bottom: int = 0 if self._modal else self._bottom
-        self._ec_top: int = self.window.height if self._modal else self._top
+        self._ec_left: int | float = 0 if self._modal else self._left
+        self._ec_right: int | float = self.window.width if self._modal else self._right
+        self._ec_bottom: int | float = 0 if self._modal else self._bottom
+        self._ec_top: int | float = self.window.height if self._modal else self._top
 
         self.camera: Projector | None = None
         """optional section camera"""
@@ -180,7 +180,7 @@ class Section:
         return self._draw_order
 
     @property
-    def left(self) -> int:
+    def left(self) -> int | float:
         """Left edge of this section"""
         return self._left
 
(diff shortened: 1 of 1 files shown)


elegantiron:
working on the failed checks

einarf:
We do have `AsFloat` that would likely be more specific in this case? 
https://github.com/pythonarcade/arcade/blob/7717e3fecdf08606f9f8b56a1d1df0a5ed235f13/arcade/types/numbers.py#L14

elegantiron:
[Looking at the docs for type hinting:](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#useful-built-in-types)

```python
# On Python 3.10+, use the | operator when something could be one of a few types
x: list[int | str] = [3, 5, "test", "fun"]  # Python 3.10+
# On earlier versions, use Union
x: list[Union[int, str]] = [3, 5, "test", "fun"]
```

We should either use `Union` or we should drop support for Python 3.9.

I assume dropping support for Python 3.9 is going to be significantly more involved than using `Union` to stay compatible and people will prefer using `Union` over changing the supported versions.

einarf:
> We should either use `Union` or we should drop support for Python 3.9.

I think `from __future__ import annotations` backports everything even for 3.9.

pushfoo:
> We should either use `Union` or we should drop support for Python 3.9.

**TL;DR:** Soon (in a few months)

For the moment switching _back_ to `Union` would be counter-productive:

1. It's eight months until [Python 3.9 EOLs in October of 2025](https://devguide.python.org/versions/)
2. We've been incrementally switching to `|` in anticipation of that

As a bonus, the doc cited is mypy doc. They have ideas which clash with our codebase(s). Specifically, we face the same issues as pyglet does with mypy due to `Rect` being a `NamedTuple` (from the [pyglet `pyproject.toml`](https://github.com/pyglet/pyglet/blob/83edbcc6c96cb4ebf8c95f061c018283d343540b/pyproject.toml#L18-L27)):

```toml
[[tool.mypy.overrides]]
# mypy incorrectly reports issues everywhere in pyglet.math because:
# 1. pyglet.math is almost entirely typing.NamedTuple subclasses
# 2. The mypy project has the following long-standing stance on
#    supporting collections.namedtuple and typing.NamedTuple:
#    "Duplicate of #5613, still low priority, better use dataclasses,
#    as suggested above."
#    https://github.com/python/mypy/issues/5944#issuecomment-441285456
module = "pyglet.math"
ignore_errors = true
```

pushfoo:
> We do have `AsFloat` that would likely be more specific in this case?

I think it was lazy shorthand / a way to make pyright pass in specific cases.

pushfoo (approved):
In my eyes, this has better legiblity than `AsFloat`. The tab-completion ergonomics aren't as good, but :shrug:

PR #2528, opened 2025-01-31
Remove deprecated function read_tmx

cdeil:
Looks like this `read_tmx` function was removed 4 years ago in #891 and is non-functional since.

OK to remove now?


einarf:
This should probably have followed the normal deprecation pattern with a warning on earlier versions BUT 3.0 is such a huge leap forward it probably makes more sense to remove this now.

PR #2530, opened 2025-01-31
Remove outdated type: ignore

cdeil:
For me `./make.py lint` shows no problems with those lines. OK to remove the `type: ignore` there?


einarf:
Yep! Looks like we forgot to remove a lot of these for pyglet 2.1 👍

benjamin-kirkbride:
I believe there is a way to configure mypy to complain if a type: ignore is
unnecessary

einarf:
Yes but we want everyone to use the same settings :)

PR #2532, opened 2025-01-31
Fix docstring args

cdeil:
Here I fixed some mismatches of docstring args vs function signatures pointed out by PyCharm.

I assumed function signatures should remain as-is in all cases and didn't touch them.

### DIFF
--- a/arcade/future/video/video_player.py
+++ b/arcade/future/video/video_player.py
@@ -41,7 +41,7 @@ class VideoPlayer:
 
         Args:
             size: Pass None as one of the elements if you want to use the
-            dimension(width, height) attribute.
+             dimension(width, height) attribute.
         """
         if size and len(size) == 2:
             self._width = size[0] or self.width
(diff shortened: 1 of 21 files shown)


arcade/future/video/video_player.py
```diff
@@ -41,7 +41,7 @@ def draw(self, left: int = 0, bottom: int = 0, size: tuple[int, int] | None = No
 
         Args:
             size: Pass None as one of the elements if you want to use the
-            dimension(width, height) attribute.
+             dimension(width, height) attribute.
```
einarf, line 44: This one doesn't make sense :)

This is in the future module so it's kind of incomplete. The docstring is also terrible. `left` `right` is the position from the lower left corner an `size` is the size of the rectangle for the video contents.

einarf (changes requested):
Thanks for this cleanup. It's a sign we should add some auto QA on this stuff.

einarf:
Thank you again. I resolved the remaining issues locally.

PR #2546, opened 2025-02-02
Small docs and example cleanup

cdeil:
Some more docstring arg / function signature mismatches pointed out by PyCharm.

Remove big try/except in `line_of_sight.py` in the `on_draw`. Works fine for me. Not clear why it would be there.


einarf:
Thanks again. We still need something to test our docstrings.

Issue #2214, opened 2024-07-05
./make.py format does not auto-fix all formatting issues


pushfoo:
### Actual behavior:

`./make.py format` doesn't aply `ruff` import auto-fixes

### Expected behavior:

`./make.py format` also runs `ruff`'s auto-fixes for formats

### Steps to reproduce/example code:

1. I ran `./make.py format`
2. CI reports that the `black` formatting issues were fixed, [but stale imports were left behind](https://github.com/pythonarcade/arcade/actions/runs/9809603611/job/27087913394?pr=2213):

```
Run python ./make.py ruff
arcade/draw/line.py:4:42: F401 [*] `arcade.types.Point` imported but unused
  |
3 | from arcade import gl
4 | from arcade.types import RGBA255, Color, Point, Point2List
  |                                          ^^^^^ F401
5 | from arcade.window_commands import get_window
  |
  = help: Remove unused import: `arcade.types.Point`

Found 1 error.
[*] 1 fixable with the `--fix` option.
Error: Process completed with exit code 1.
```

einarf:
Should not be a problem after switching to ruff

Issue #2165, opened 2024-06-30
Improve beginner friendliness of output from ./make.py format


pushfoo:
## Enhancement

CI formatting output could be clearer for beginners.

### Actual behavior:
![image](https://github.com/pythonarcade/arcade/assets/36696816/b72c03b3-5633-4358-aaaa-bde2f77cda38)

### Expected behavior:
Useful debug output

### Steps to reproduce/example code:

PR something or run `./make.py format --check`

Cleptomania:
What else would you want it to say? It shows the problematic file.

You’re not meant to go reformat it by hand, it’s meant to run `make.py format` and it will fix it with the auto formatter.

pushfoo:
I think it should suggest the following:
1. running `./make.py format` locally
2. `git diff` to highlight changes

The current output is confusing because the following aren't clear:
* whether `ruff` or `black` is the active formatter
* what the next steps to begin fixing the formatting issue are

pushfoo:
Making those next steps clear matters because running `ruff format` changed a few hundred lines for me.

I reverted since I remembered we're using `black` + config tweaks this project, but beginners will get confused.

cdeil:
@pushfoo and all - I took a stab at this in #2550 

One further suggestion I have is to delete all or most "Docs Extra Formats" from `make.py`. Having so much stuff is confusing for newcomers like me, and is anyone actually using them? I think this is just generating short sphinx commands that are very rarely used? Centralizing config to `pyproject.toml` and avoiding extra config like all the variables in `make.py` related to sphinx build I think would be nice  - assuming all dev/maintainer needs/requirements can be met.

Also ruff is so fast and consistent that I think we could simply run ruff format including the import sorting and get rid of the multiple commands there?

```
% ./make.py --help
                                                                                                            
 Usage: make.py [OPTIONS] COMMAND [ARGS]...                                                                 
                                                                                                            
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                  │
│ --show-completion             Show completion for the current shell, to copy it or customize the         │
│                               installation.                                                              │
│ --help                        Show this message and exit.                                                │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Docs Extra Formats ─────────────────────────────────────────────────────────────────────────────────────╮
│ changes       to make an overview of all changed/added/deprecated items                                  │
[...]

PR #2550, opened 2025-02-02
Switch from black to ruff format

cdeil:
This PR has three commits that are best reviewed one by one:

* 5f03082ad14bf14fb05be3b68109f1d5bcea4543 improves make.py a bit
* 226e4e5e226ff89faf7bb31ac62fca1a60e87991 has some ruff format changes on arcade (not many changes, and very little ones removing empty lines or adding trailing commas mostly)
* c1ac6965eba6f683a36a1734eeb7ffd6eafefa3a switches make.py, pyproject.toml and CONTRIBUTING.md from black to ruff format

Related: #2139, #2165, #2214

Looks like `.pre-commit-config.yaml` already only runs `ruff format`, no `black`.
So overall I think this is almost a drop-on replacement, just with some small gains in simplicity / consistency / speed.

Thoughts?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,14 @@ jobs:
         id: wheel
         run: |
           python -m pip install -e .[dev]
+      - name: "code-inspection: formatting"
+        if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
+        run: |
+          python ./make.py format --check
+      - name: "code-inspection: ruff-check"
+        if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
+        run: | 
+          python ./make.py ruff-check
       - name: "code-inspection: mypy"
         if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
         run: | 
@@ -44,14 +52,6 @@ jobs:
         if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
         run: | 
           python ./make.py pyright
-      - name: "code-inspection: ruff"
-        if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
-        run: | 
-          python ./make.py ruff
-      - name: "code-inspection: formatting"
-        if: ${{ (success() || failure()) && steps.wheel.outcome == 'success' }}
-        run: |
-          python ./make.py format --check
       # Prepare the Pull Request Payload artifact. If this fails,
       # we fail silently using the `continue-on-error` option. It's
       # nice if this succeeds, but if it fails for any reason, it
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,23 +81,19 @@ steps at the end of this document.
 
 ## Formatting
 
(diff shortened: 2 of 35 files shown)


einarf:
Discussing this here https://discord.com/channels/458662222697070613/1335655131500249230/1335655140421800007

cdeil:
Yes? No? Maybe?

einarf:
Probably. No resistance doing this change but properly mapping out the differences and double checking the config options is a good idea. Also need to make sure the tooling is up to date for the environments people are currently using.

Resolved a conflict with dev branch to keep this up to date.

cdeil:
There's infos on `ruff format` here:  https://docs.astral.sh/ruff/formatter/

Basically it's 99.9% compatible with Black:

> As such, the formatter is designed as a drop-in replacement for [Black](https://github.com/psf/black), but with an excessive focus on performance and direct integration with Ruff. Given Black's popularity within the Python ecosystem, targeting Black compatibility ensures that formatter adoption is minimally disruptive for the vast majority of projects.

The 0.1% deviations are explained here: https://docs.astral.sh/ruff/formatter/black/

I think you're already running `ruff format` and not `black` currently on precommit:
https://github.com/pythonarcade/arcade/blob/f4f4521d4cd263e0e5f82e791bc475044bb35c77/.pre-commit-config.yaml#L17-L18

So this PR would align devdocs / CI in line and go all-in on `ruff format` vs the current solution of using two formatters and be faster. I think that's the main/only advantage, I would be surprised if anyone cares about the tiny formatting differences between the two.

Let me know if there's anything else I can do here.

einarf:
Yup. We seems to have the majority of the maintains on board. I will adapt one of my local projects quickly first before I merging this just to get the full picture even if it's meant to be a drop-in replacement.

Will merge in the next few days.

The pre-commit I think was only a test that @eruvanos was using. It had some challenges depending on how your environment is set up. We should probably look into that as well soon.

pushfoo:
I'm up for merging this.

1. You fixed some weird fstring issues
2. The root cause for them still isn't fixed in `black` https://github.com/psf/black/issues/4389

pushfoo:
#2569 just got hit with the exact issue I mentioned (f-string handling in https://github.com/psf/black/issues/4389).

> I  will adapt one of my local projects quickly first before I merging this just to get the full picture even if it's meant to be a drop-in replacement.

@einarf any reasons not to merge this?

einarf:
> #2569 just got hit with the exact issue I mentioned (f-string handling in [psf/black#4389](https://github.com/psf/black/issues/4389)).
> 
> > I  will adapt one of my local projects quickly first before I merging this just to get the full picture even if it's meant to be a drop-in replacement.
> 
> @einarf any reasons not to merge this?

Be free to merge. I have been swamped with work.

pushfoo (approved):
TL;DR: Let's do it. :rocket: 

I have one minor personal style preference to gripe over, and that's handled by [`# fmt: off` + `# fmt: on`](https://docs.astral.sh/ruff/formatter/#format-suppression) if I really need it.

PR #2561, opened 2025-02-11
Small UI docs and example cleanup

cdeil:
Mainly a fix in `arcade/examples/gui/0_basic_setup.py` removing two broken `on_click` handlers were attached that then were overwritten and unused.


eruvanos:
Thank you

PR #2562, opened 2025-02-12
Fix broken links and formatting on the sample games page

pushfoo:


pushfoo:
EDIT: local upstream target / ref issues are now fixed.

~~something's messed up with git + not sure why there are warnings. Marking this as draft until resolved.~~

PR #2569, opened 2025-02-17
Fix arcade.version.VERSION conversion from GitHub CI format

pushfoo:
**TL;DR:** Stop swallowing last `.` + add extensive tests, error reporting, and doc

### Why?

Release time is not a fun time to be debugging weird issues:

1. cspotcode is right, the current code is broken:
    > [The version numbering bug will publish it as a minor bump instead of a patch bump](https://discord.com/channels/458662222697070613/458662458198982676/1340844291706196040)
2. We should verify that our release-critical code does what it says
3. The code should tell us what went wrong during build, not `ERROR: Unable to load version number via './VERSION'`

### What?

1. Explain what's in the can and why
    * Our release process
    * A ctrl-clickable link to the version bump GH action's config YAML
2. Aggressive tests covering various possible typos or misconfigurations
3. Explain what went wrong in more detail if/when it does

Bonus: special-case `"3.0.0"` to `"3.0"` to avoid churn in bookmarks / etc change detection due to the title changing.

### DIFF
--- a/arcade/version.py
+++ b/arcade/version.py
@@ -79,23 +79,24 @@ def _parse_python_friendly_version(version_for_github_actions: str) -> str:
     Returns:
         A Python-friendly version string.
     """
-    # Extract our raw data
+    # Quick preflight check: we don't support tuple format here!
     if not isinstance(version_for_github_actions, str):
         raise TypeError(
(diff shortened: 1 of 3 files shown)


pushfoo:
In case anyone asks why I merged this already:

1. It fixes a bug
1. It adds unit tests future implementations with even better reporting can use
3. It sets up cleaning more things from `doc/conf.py`

PR #2570, opened 2025-02-17
Make on_key_release args consistent with on_key_press

cdeil:


eruvanos:
Thank you, really appreciate such fixes!

Issue #2540, opened 2025-02-01
GUI camera match_window not working


cdeil:
I think this `camera_gui.match_window()` isn't properly working at the moment?

https://github.com/pythonarcade/arcade/blob/333ce6e7fb45323da6ebd9f566e5de8ee2bbe30e/arcade/examples/minimap_texture.py#L200-L207

If I resize the window I get this - the GUI elements aren't placed correctly any more:

<img width="1160" alt="Image" src="https://github.com/user-attachments/assets/2a925b74-4078-426d-8fe4-862bf3ca08da" />

This is with latest dev version ( 333ce6e )

```
Arcade 3.0.0.rc3
----------------
vendor: Apple
renderer: Apple M3
version: (4, 1)
python: 3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 16.0.0 (clang-1600.0.26.4)]
platform: darwin
pyglet version: 2.1.2
PIL version: 11.0.0
```

einarf:
@DragonMoffon Any clue what's happening here?

DragonMoffon:
The gui has its own internal camera that you also need to update if im not mistaken.

einarf:
> The gui has its own internal camera that you also need to update if im not mistaken.

Yes, but this is not using the actual GUI. It's just called a "gui camera" because it does ortho projection drawing minimap with sprites and some text.

einarf:
Fixed in referenced PR

DragonMoffon:
ah yeah I see whoops I should have looked into the code.

einarf:
> ah yeah I see whoops I should have looked into the code.

I may or may not have assumed the same initially 😅

PR #2572, opened 2025-02-17
Fix GUI cam on_resize examples

cdeil:
In `arcade/examples/sprite_move_scrolling.py` the call to `self.camera_gui.match_window()` doesn't work, after resize it's misplaced (see screenshot). Another case of #2540 .

Calling `self.camera_gui.match_window(position=True)` does fix the issue, but also simply removing the call completely fixes the issue, so that's what I did here.

In `arcade/examples/sprite_move_scrolling_shake.py` there was a similar case, but there the `camera_gui` was completely unused, so I removed it.

Screenshot 2025-02-17 at 12 56 55


eruvanos:
Thank you

PR #2578, opened 2025-02-21
Fix the type signature of collision check methods.

### LINKED ISSUE #2576: check_for_collision_with_list does not return the right SpriteType
Hi! First, thanks a lot for this library. I am using it this semester as a teaching vehicle at the university level. It is really nice. Unlike the platformer tutoriel, I actually mandate using mypy in strict mode for all students' code.

---

It seems to me that the type signature of `arcade.check_for_collision_with_list` is incorrect. It is defined as
https://github.com/pythonarcade/arcade/blob/5743504b43416e4a4c68bca4fe7503bca67ab2ee/arcade/sprite_list/collision.py#L189-L193
You can see that it returns a `List[SpriteType]`, where `SpriteType` is the type of the *single* `sprite`. However, in practice it returns elements of the given `sprite_list`.

Because of that, when trying to use it as follows:
```python
    enemies: arcade.SpriteList[Enemy]

        for enemy in arcade.check_for_collision_with_list(weapon, self.enemies):
            enemy.hit_by_weapon() # method only available on Enemy; not on Sprite
            return
```
I get a type error telling me `hit_by_weapon` is not a method of `Sprite`.

I think the correct signature should be
```python
def check_for_collision_with_list(
    sprite: BasicSprite,
    sprite_list: SpriteList[SpriteType],
    method: int = 0,
) -> List[SpriteType]:
```

### Workaround

Use `arcade.check_for_collision_with_lists(self, [self.enemies])` instead. For some reason the alternative method with an `Iterable[SpriteList]` has the correct type:
https://github.com/pythonarcade/arcade/blob/5743504b43416e4a4c68bca4fe7503bca67ab2ee/arcade/sprite_list/collision.py#L249-L253

sjrd:
The type of the sprites in the result never depend on the first argument single sprite. Instead, it depends on the type of sprites in the given `sprite_list`.

Previously, some signatures did not accurately reflect those properties. This commit fixes them.

Fixes #2576.

---

I don't think this can be *tested* per se? The tests don't seem to be checked by mypy, are they?

FWIW, changing `sprite_list: SpriteList` to `SpriteList[BasicSprite]` (only) *did* report type errors inside `check_collision_with_sprite_list`. It previously passed the type checker only because it was inferred as `SpriteList[Unknown]`, which it was happy to assign to `Iterable[SpriteType]`.

### DIFF
--- a/arcade/sprite/base.py
+++ b/arcade/sprite/base.py
@@ -801,7 +801,7 @@ class BasicSprite:
         x, y = point
         return is_point_in_polygon(x, y, self.hit_box.get_adjusted_points())
 
-    def collides_with_sprite(self: SpriteType, other: SpriteType) -> bool:
+    def collides_with_sprite(self, other: BasicSprite) -> bool:
         """Will check if a sprite is overlapping (colliding) another Sprite.
 
         Args:
@@ -813,7 +813,7 @@ class BasicSprite:
 
         return check_for_collision(self, other)
 
-    def collides_with_list(self: SpriteType, sprite_list: "SpriteList") -> list[SpriteType]:
+    def collides_with_list(self, sprite_list: SpriteList[SpriteType]) -> list[SpriteType]:
         """Check if current sprite is overlapping with any other sprite in a list
 
         Args:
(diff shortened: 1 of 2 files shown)


pushfoo (approved):
I'm in favor of merging since:
1. The `SpriteList` is more crucial than the physics engines
2. The physics engine annotations are sort of kludgy
3. #2580

Issue #2588, opened 2025-03-03
Restore alt text + links for shields in README.md


pushfoo:
### What's Wrong?

The shields at the top of README.md normally (The buttons which look like ![the MIT License shield](https://camo.githubusercontent.com/e5faf29067104aca718f979735273afb07bb6e686ab8cf6dbf3b40a76ea13c71/68747470733a2f2f696d672e736869656c64732e696f2f707970692f6c2f617263616465), etc) link to something relevant.

In our case, these once alt text which showed up hover as well as outbound links. For example, some of the links included:
* Our PyPI page
* http://makeapullrequest.com
* https://www.firsttimersonly.com/

### Cause

Commit [`b33564`](https://github.com/pythonarcade/arcade/commit/b031c967036d24fea70205ebd8a307ba35067107#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5):
1. converted the README.md's shields from Markdown to HTML `<img>` elements
2. wrapped the shield blocks in `<p class="center">`
3. forgot to add the link and alt text back

### Suggested fix

Refer to the diff for [`b33564`](https://github.com/pythonarcade/arcade/commit/b031c967036d24fea70205ebd8a307ba35067107#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) to add the following for each shield `<img>`:
1. An `alt="original text here"` attribute
2. A wrapping `<a href="https://www.oldlink.url">` tag around each `<img>`

PR #2593, opened 2025-03-03
Restored alt text + links for shields in README.md

MrValdez:
Fix issue #2588 

- added explicit `<a href>` tag around shield images.
- added alt and title to enable text hover.
- added back missing hover text.
- added hover text for "GitHub Contributors" and "GitHub Stars".

Note that I've added both alt and text. Under Firefox, alt is not enough to cause text to hover (https://support.mozilla.org/en-US/questions/838888).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,15 @@
 # Welcome to The Arcade Library!
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


pushfoo:
The FireFox detail is a good catch, ty!

PR #2594, opened 2025-03-03
Bandaid fix for link to pyglet module not working

### LINKED ISSUE #2585: Quick-fix the top of the install instructions page
### What's broken?
At the top of this page, we have some oddness as a result of `:py:mod` and a conflict with something else:
https://api.arcade.academy/en/latest/get_started/install.html#requirements-raspi

This is likely the \``pyglet`_\` link target in the global links include getting lost in the bandaid-rip of 3.0's realease.

### Proposed quick fix

Replace this:
https://github.com/pythonarcade/arcade/blob/8a41838a9dd89c8f35e89b22111936effbdd43d8/doc/get_started/install.rst?plain=1#L13-L15

With something like this:

```
Arcade requires a desktop, laptop, or compatible Single-Board Computer (SBC) with:

#. Python 3.9 or higher
#. Graphics drivers with support for either:
    * OpenGL 3.3+
    * GLES 3.1+ with extensions on SBCs
```

### Why a quick fix?

Sphinx extensions and links are prone to compatibility issues. The approach above seems to work, but test it locally via:

1. `./make.py html`
2. One of the following:
   * `./make.py server`
   * If you have zombie process issues / rebuild loops:
     1. Report it in detail
     2. Use `python -m http.server -d build` instead
     3. hard refresh pages manually (ctrl-f5 or ctrl-click the refresh icon in the GUI)

max-torch:
A temporary solution is implemented here to just remove the link to `pyglet` for now and instead say that you need to have graphics drivers with support for OpenGL or GLES.

Closes #2585

image

### DIFF
--- a/doc/get_started/install.rst
+++ b/doc/get_started/install.rst
@@ -10,9 +10,12 @@ Install
 
 Requirements
 ------------
-:mod:`pyglet`
+Arcade requires a desktop, laptop, or compatible Single-Board Computer (SBC) with:
 
-All systems require Python 3.9 or higher on a desktop or laptop device.
+#. Python 3.9 or higher
+#. Graphics drivers with support for either:
+    * OpenGL 3.3+
+    * GLES 3.1+ with extensions on SBCs
 
 :ref:`Web <faq_web>` and :ref:`mobile <faq_mobile>` are currently
 unsupported.


pushfoo:
I'll merge this as-is. If Clepto clarifies that it's a single extension rather than multiple, I'll get the fix in a subsequent pass on the page  once I have better requirements written up.

PR #2599, opened 2025-03-04
Docs: Fix broken pyglet windowing links

csd4ni3l:
Fixes the pyglet window style and pyglet window size and position links, which were broken due to wrong syntax.


pushfoo (approved):
Looks good, ty!

PR #2610, opened 2025-03-15
Make NinePatchTexture lazy

einarf:
Make nine-patches lazy meaning they can be created before the window

### DIFF
--- a/arcade/gui/nine_patch.py
+++ b/arcade/gui/nine_patch.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
-from typing import Optional
-
 import arcade
 import arcade.gl as gl
+from arcade.texture_atlas.base import TextureAtlasBase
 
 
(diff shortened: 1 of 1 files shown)


arcade/gui/nine_patch.py
```diff
@@ -72,39 +71,66 @@ def __init__(
...
+        print("NinePatchTexture initialized")
+        self._initialized = True
 
-        self._check_sizes()
+    def initialize(self) -> None:
+        """
+        Manually initialize the NinePatchTexture if it was lazy loaded.
+        This has no effect if the NinePatchTexture was already initialized.
+        """
+        if self._initialized:
```
alejcas, line 126: Shouldn't this be ```if not self._initialized:```?

einarf: Ops. You are right.

PR #2642, opened 2025-04-03
Make arcade.Text lazy

csd4ni3l:
This is a draft to make arcade.Text lazy.

It works in the same way as the lazy NinePatch, using the `_initialized` attribute and the `initalize` and `_init_deferred` functions. It raises a `RuntimeError` if a property of the label is accessed before initializing.

I dont know how to handle arguments correctly or efficiently, so its marked as a draft, but in its current state, its still usable.

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -317,7 +317,8 @@ class Text:
         """
         The underlying pyglet.Label instance.
         """
-        self._init_deferred()
+        if not self._initialized:
+            self._init_deferred()
         return self._label
 
     def initialize(self) -> None:
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -641,7 +641,7 @@ class Text:
             raise RuntimeError("Text must be initialized before setting the height.")
 
     @property
-    def size(self) -> tuple[int, int] | None:
+    def size(self) -> tuple[int | None, int | None]:
         """Get the size of the label."""
         if self._initialized:
             return self._label.width, self._label.height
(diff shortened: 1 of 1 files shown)


arcade/text.py
```diff
@@ -276,8 +276,10 @@ def __init__(
         z: float = 0,
         **kwargs,
     ):
-        # Raises a RuntimeError if no window for better user feedback
-        arcade.get_window()
+        self._initialized = False
+        self.arguments = [text, x, y, color, font_size, width, align, font_name, bold,
+                          italic, anchor_x, anchor_y, multiline, rotation, batch, group, z]
+        self.kwargs = kwargs
```
einarf: Should probably be a private member

arcade/text.py
```diff
@@ -288,54 +290,82 @@ def __init__(
...
+        self._initialized = True
+
     def __enter__(self):
         """
         Update multiple attributes of this text,
         using efficient update mechanism of the underlying ``pyglet.Label``
         """
-        self._label.begin_update()
+        if self._initialized:
+            self._label.begin_update()
```
einarf: Instead of checking for initialized everywhere could we make a `label` property that returns the underlying pyglet label instead and only have the initialized check there?

einarf (changes requested):

arcade/text.py
csd4ni3l: Good idea. Let me do that as soon as i can.

einarf (approved):

Issue #2580, opened 2025-02-24
Completely overhaul Arcade's physics engines, and physics engine dependencies.


DragonMoffon:
Arcade has not had proper physics engines since its inception. 

The built-in pure python implementations served their purpose in the past, but as arcade becomes more complex and used by more people they will quickly be out paced. They have several flaws that wreck their performance and go against modern physics engine design.

Pymunk was partially the solution to this, but it has binaries that limit's arcade's range of support. There has already been work done to make this optional, and in the next few updates should hopefully be fully separated.

Currently in Arcade sprites always have a hitbox which is either tied to the 'first' texture they are given or set manually. This is a huge waste as most sprites don't need a hitbox, and forces users to use sprite lists for collision checking which is a confusing and inconvenient.

I propose we provide 2-3 types of physics engine.

1st is a rotation less 'simple' engine that's perfect for platformers and top-down games like celeste, super meat boy, or civ where the physics is for movement and collisions rather than anything dynamic.
2nd is a classic 2D rigidbody and softbody engine (might need to be two systems depending on how performance is)
3rd are optional bindings to various high performance 2d libraries with an abstraction layer to fit arcade's physics paradigm. This obviously includes Pymunk, but I'm also interested in getting Box2D bindings working for more options.

eschan145:
> but I'm also interested in getting Box2D bindings working for more options.

I wouldn't recommend Box2D since none of its Python bindings are actively maintained, and the 3.0 release rewrote the engine in C, making future updates to those bindings very unlikely. I would suggest a physics engine built using compute shaders in glsl. I'm not sure if this is feasible because it would be very hard to develop, but it's just an idea.

DragonMoffon:
Ah, good point. That's unfortunate news, but there are other options.

Yeah, we could definitely try to go down that route, but I'd like some cpu side stuff first for feature parity.

PR #2647, opened 2025-04-10
Introduce SpriteSequence, a covariant supertype of SpriteList.

sjrd:
This is done by analogy to `collections.abc.Sequence`, which is a covariant supertype of `list`.

Before this commit, many parts of the codebase used `SpriteList`s without type arguments (defaulting to `Unknown`). That was the only way to allow reasonable usages of the given methods and attributes. However, doing so results in weaker typing.

Using `SpriteSequence`, we can add correct type arguments to almost all of the references that were using `SpriteList`s before.

The only missing pieces are `Scene` and `TileMap`. Unfortunately, their APIs are fundamentally unsound wrt. the type arguments of their `SpriteList`s. We cannot make it sound without breaking their APIs, so we do not change them.

As a bonus, we can now create lists of `SpriteList`s with varying type arguments, and generically call `draw` or `update` on them. Previously, the only common supertype of `SpriteList[A]` and `SpriteList[B]` was `object`, which meant it was not possible to call those methods on them.

In a sense, that ability mostly subsumes the convenience provided by `Scene`. A `list[SpriteSequence[BasicSprite]]` is almost as convenient, while being type-safe.

---

This was prompted by seeing students struggle with the absence of a common supertype for `SpriteList[A]` and `SpriteList[B]`. Students resorted to `cast`s or `Any`s (or smuggled their way out of that using `Scene`) to work around the issue. With `SpriteSequence`, this shouldn't be an issue anymore.

It turns out that the introduction of `SpriteSequence` allows to more safely type many APIs of Arcade, so hopefully this addition carries its weight.

### DIFF
--- a/arcade/sprite_list/collision.py
+++ b/arcade/sprite_list/collision.py
@@ -32,7 +32,7 @@ def get_distance_between_sprites(sprite1: SpriteType, sprite2: SpriteType) -> fl
 
 
 def get_closest_sprite(
-    sprite: SpriteType, sprite_list: SpriteList
+    sprite: BasicSprite, sprite_list: SpriteList[SpriteType]
 ) -> Tuple[SpriteType, float] | None:
     """
     Given a Sprite and SpriteList, returns the closest sprite, and its distance.
--- a/arcade/sprite_list/collision.py
+++ b/arcade/sprite_list/collision.py
@@ -17,7 +17,7 @@ from arcade.sprite import BasicSprite, SpriteType
 from arcade.types import Point
 from arcade.types.rect import Rect
 
-from .sprite_list import SpriteList
+from .sprite_list import SpriteSequence
 
 
 def get_distance_between_sprites(sprite1: SpriteType, sprite2: SpriteType) -> float:
@@ -32,7 +32,7 @@ def get_distance_between_sprites(sprite1: SpriteType, sprite2: SpriteType) -> fl
 
 
 def get_closest_sprite(
-    sprite: BasicSprite, sprite_list: SpriteList[SpriteType]
+    sprite: BasicSprite, sprite_list: SpriteSequence[SpriteType]
 ) -> Tuple[SpriteType, float] | None:
     """
     Given a Sprite and SpriteList, returns the closest sprite, and its distance.
@@ -74,7 +74,7 @@ def check_for_collision(sprite1: BasicSprite, sprite2: BasicSprite) -> bool:
     if __debug__:
         if not isinstance(sprite1, BasicSprite):
             raise TypeError("Parameter 1 is not an instance of a Sprite class.")
-        if isinstance(sprite2, SpriteList):
+        if isinstance(sprite2, SpriteSequence):
             raise TypeError(
                 "Parameter 2 is a instance of the SpriteList instead of a required "
                 "Sprite. See if you meant to call check_for_collision_with_list instead "
@@ -133,7 +133,7 @@ def _check_for_collision(sprite1: BasicSprite, sprite2: BasicSprite) -> bool:
 
 
 def _get_nearby_sprites(
-    sprite: BasicSprite, sprite_list: SpriteList[SpriteType]
+    sprite: BasicSprite, sprite_list: SpriteSequence[SpriteType]
 ) -> List[SpriteType]:
     sprite_count = len(sprite_list)
     if sprite_count == 0:
@@ -186,7 +186,7 @@ def _get_nearby_sprites(
 
 def check_for_collision_with_list(
     sprite: BasicSprite,
-    sprite_list: SpriteList[SpriteType],
+    sprite_list: SpriteSequence[SpriteType],
     method: int = 0,
 ) -> List[SpriteType]:
     """
@@ -218,7 +218,7 @@ def check_for_collision_with_list(
                 f"Parameter 1 is not an instance of the Sprite class, "
                 f"it is an instance of {type(sprite)}."
             )
-        if not isinstance(sprite_list, SpriteList):
+        if not isinstance(sprite_list, SpriteSequence):
             raise TypeError(f"Parameter 2 is a {type(sprite_list)} instead of expected SpriteList.")
 
     sprites_to_check: Iterable[SpriteType]
@@ -246,7 +246,7 @@ def check_for_collision_with_list(
 
 def check_for_collision_with_lists(
     sprite: BasicSprite,
-    sprite_lists: Iterable[SpriteList[SpriteType]],
+    sprite_lists: Iterable[SpriteSequence[SpriteType]],
     method=1,
 ) -> List[SpriteType]:
     """
@@ -290,7 +290,7 @@ def check_for_collision_with_lists(
     return sprites
 
 
-def get_sprites_at_point(point: Point, sprite_list: SpriteList[SpriteType]) -> List[SpriteType]:
+def get_sprites_at_point(point: Point, sprite_list: SpriteSequence[SpriteType]) -> List[SpriteType]:
     """
     Get a list of sprites at a particular point. This function sees if any sprite overlaps
     the specified point. If a sprite has a different center_x/center_y but touches the point,
@@ -303,7 +303,7 @@ def get_sprites_at_point(point: Point, sprite_list: SpriteList[SpriteType]) -> L
     :returns: List of sprites colliding, or an empty list.
     """
     if __debug__:
-        if not isinstance(sprite_list, SpriteList):
+        if not isinstance(sprite_list, SpriteSequence):
             raise TypeError(f"Parameter 2 is a {type(sprite_list)} instead of expected SpriteList.")
 
     sprites_to_check: Iterable[SpriteType]
@@ -321,7 +321,7 @@ def get_sprites_at_point(point: Point, sprite_list: SpriteList[SpriteType]) -> L
 
 
 def get_sprites_at_exact_point(
-    point: Point, sprite_list: SpriteList[SpriteType]
+    point: Point, sprite_list: SpriteSequence[SpriteType]
 ) -> List[SpriteType]:
     """
     Get a list of sprites whose center_x, center_y match the given point.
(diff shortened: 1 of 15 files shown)


pushfoo:
> The only missing pieces are `Scene` and `TileMap`. Unfortunately, their APIs are fundamentally unsound wrt. the type arguments of their SpriteLists.

@DragonMoffon and I want to change `Scene`. I don't make use of `TileMap` that much, so I don't have a strong opinion on it.

arcade/sprite_list/spatial_hash.py
```diff
@@ -1,13 +1,54 @@
...
+class ReadOnlySpatialHash(Generic[SpriteType_co]):
+    """
+    Read-only view of a SpatialHash.
+
+    This is useful when the SpriteType is in covariant position.
+
+    See SpatialHash for more details.
+    """
+
+    @abstractmethod
```
pushfoo, line 29: As a heads up, Python's generics are kind of a mess compared to those in the Java ecosystem and anything related to it (including your Scala projects :eyes:).

I'm confused as to whether/how this actually works. My reading is that `typing.Protocol` may be a better fit since:

1. [`abc.abstractmethod` requires classes to have `ABCMeta` as their metaclass](https://docs.python.org/3/library/abc.html#abc.abstractmethod)
2. `Generic` has no metaclass at all since it's [defined as a C-backed type in `_typingmodule.c`](https://github.com/python/cpython/blob/00cf5eacc52d4790c23a24c835fb6b048cb27d8c/Modules/_typingmodule.c#L62)
3. We may want to support different spatial structures (quad tree, etc) in the future

arcade/sprite_list/spatial_hash.py
```diff
@@ -1,13 +1,54 @@
...
 
-class SpatialHash(Generic[SpriteType]):
+class ReadOnlySpatialHash(Generic[SpriteType_co]):
+    """
+    Read-only view of a SpatialHash.
+
+    This is useful when the SpriteType is in covariant position.
+
+    See SpatialHash for more details.
+    """
```
pushfoo: Maybe something like this? The wording's less a concern to me than getting the cross-refs solid.
```suggestion
    """A read-only view of a :py:class:`.SpatialHash` which helps preserve safety.

    This works like the read-only views of Python's built-in :py:class:`dict` and other types. As an every-day user, it means
    you can put subtypes of the annotated type into the `SpatialHash` or a `SpriteList`, but not superclasses.
    
    This ensures predicable behavior via type safety in cases where:

    #. A spatial hash is annotated with a specific type
    #. The object could be modified outside the original context to add a broader type
   
    Advanced users who want more information on the specifics should see the comments of :py:class:`~arcade.sprite_list.SpriteList`.
    """
```

arcade/sprite_list/sprite_list.py
```diff
@@ -39,8 +40,109 @@
...
+class SpriteSequence(Collection[SpriteType_co]):
+    """
+    A read-only view of a SpriteList.
+
+    Like collections.abc.Sequence, a SpriteSequence is covariant in its sprite
+    type. This is useful when you want to manipulate a collection of
+    SpriteLists of different sprite types.
+
+    See SpriteList for more details.
+    """
```
pushfoo:
```suggestion
    """A read-only view of a :py:class:`.SpriteList`.

    Like other generics such as :py:class:`collections.abc.Sequence`, a `SpriteSequence` requires
    sprites be of a covariant type relative to their annotated type.
    
    See :py:class:`.SpriteList` for more details.
    """
```

arcade/sprite_list/spatial_hash.py
```diff
@@ -1,13 +1,54 @@
...
+        """
+        ...
+
+    @abstractmethod
+    def get_sprites_near_rect(self, rect: Rect) -> set[SpriteType_co]:
+        """
+        Return sprites in the same buckets as the given rectangle.
+
+        Args:
+            rect: The rectangle to check (left, right, bottom, top)
```
pushfoo: If we're adding this, we may as well correct it.
```suggestion
    .. tip:: Use :py:mod:`arcade.types.rect`'s helper functions to create rectangle objects!

    Args:
        rect: The rectangle to check as a :py:class:`~arcade.types.rect.Rect` object.
```

tests/unit/spritelist/test_spritesequence.py
```diff
@@ -0,0 +1,26 @@
...
+
+def test_collective_draw(window: arcade.Window) -> None:
+    sprite_list1: arcade.SpriteList[arcade.Sprite] = arcade.SpriteList()
+    sprite_list1.append(arcade.SpriteSolidColor(16, 16, color=(255, 0, 0, 1)))
+
+    sprite_list2: arcade.SpriteList[_CustomSpriteSolidColor] = arcade.SpriteList()
+    sprite_list2.append(_CustomSpriteSolidColor(16, 16, color=(255, 0, 0, 1)))
+
+    # It really is a SpriteList with a good type; this would not typecheck otherwise
+    custom_sprite: _CustomSpriteSolidColor = sprite_list2[0] # assert_type
```
pushfoo, line 14: Are you running this type check manually? My understanding is that we don't run type checks against the test folder at all. See:

https://github.com/pythonarcade/arcade/blob/94c42806cf63df636e2ae6eb2e8a0ceb25e31225/pyproject.toml#L140-L150

pushfoo (changes requested):
Thank you for starting taking a crack at this. I appreciate it since it's sorely needed (see #2580 and similar :sweat_smile:).

For the moment, I'm a little confused on some `abc`-related details as in [this comment](https://github.com/pythonarcade/arcade/pull/2647#discussion_r2041096869). If nobody beats me to giving this a closer look, I may need a few days before I can give it the full attention it deserves.

@eruvanos btw, do you think we should add a typing-specific tests folder or similar as an exception to the tests folder being ignored?

arcade/sprite_list/spatial_hash.py
pushfoo: Update: At least on Python 3.13, `typing.Generic` doesn't work with `abc.abstractmethod` at all :(

Here's a cut-down version of the output + [the gist I wrote](https://gist.github.com/pushfoo/8a01bc61de554b5b11de8e86ef28c5c4):
```
Trying to instantiate GenericBase...
  FAIL: Did NOT raise a TypeError!

Trying to instantiate GenericSub...
  FAIL: Did NOT raise a TypeError!
```
```python
import abc
from typing import Collection, TypeVar, Generic

T_co = TypeVar('T_co', covariant=True)


def does_it_type_error(t: Generic[T_co]) -> Generic[T_co]:
    """A decorator which tries to instantiate the passed class immediately."""

    name = t.__name__
    try:
        print(F"Trying to instantiate {name}...")
        t()
        print(f"  FAIL: Did NOT raise a TypeError!\n")
    except TypeError as _:
        print(f"  PASS: Raised a TypeError.\n")
    finally:
        return t


@does_it_type_error
class GenericBase(Generic[T_co]):

    @abc.abstractmethod
    def local_method(self) -> int:
        ...


@does_it_type_error
class GenericSub(GenericBase[T_co]):
    ...
```

sjrd: Huh, gosh, indeed it needed `Protocol`. It wasn't checking nearly as much as I thought it was just with `Generic`. Adding `Protocol` flagged the use of `set[SpriteType_co]` in covariant position, which I replaced by `collections.abc.Set`.

Coming from Scala, as you noticed, I'm generally pleasantly surprised by what the Python type system can express these days. But I sure get confused with Python's idea of what an abstract class/interface/ABC is.

arcade/sprite_list/spatial_hash.py
sjrd: I took your suggestion but I amended a few things. LMK if you disagree.

tests/unit/spritelist/test_spritesequence.py
sjrd: I don't think I ran pyright, but I did run mypy manually.

sjrd (commented):
Thanks for the review. I addressed the comments so far.

I'm not sure what your squashing policy is. For now I added a comment addressing the review, but I'm also happy to squash it with the parent commit.

sjrd:
@pushfoo May I suggest a ping here? Is there anything else you'd like me to do? Thank you.

eruvanos:
@pushfoo Regarding the typed test folder: if we have tests that benefit from such a setup I am fine, I wonder if it makes our test structure more complicated. 

Regarding Scene and TileMap: we can iterate on them after we merge this PR. I just created a small code base over easter which uses Scene and TileMap, so I could doublecheck the improved versions.

@sjrd we do not have a strict squashing policy, I prefer squashed PRs to reduce the amount of commits we have to read through during write up of release notes. 
Talking about release notes, could you add your change to the release notes of `3.2 (unreleased)`?

sjrd:
Thanks. I squashed the last two commits, and added the entry to the changelog. I had to rebase on top of #2652 not to have conflicts. The only changes besides rebasing are the `CHANGELOG` file.

eruvanos:
@pushfoo are you fine with the changes? I think we could move on with this and iterate on Scene and Tilemap (and typed tests afterwards). What is your opinion?

arcade/sprite_list/spatial_hash.py
pushfoo: > But I sure get confused with Python's idea of what an abstract class/interface/ABC is.

So does Python, sadly. That's not your fault.

arcade/sprite_list/collision.py
```diff
@@ -365,7 +365,7 @@ def get_sprites_in_rect(rect: Rect, sprite_list: SpriteList[SpriteType]) -> List
         List of sprites colliding, or an empty list.
     """
     if __debug__:
-        if not isinstance(sprite_list, SpriteList):
+        if not isinstance(sprite_list, SpriteSequence):
```
pushfoo, line 368: This is using `Collection` rather than protocol so I think this may work without a perf hit / lying the way `Protocol` isinstance checks work.

pushfoo (approved):
Sorry for the delay. I'll say let's merge it to keep velocity and revisit if we need to. Ty for your contribution and your patience.

tests/unit/spritelist/test_spritesequence.py
pushfoo: We auto-run it in our side for build CI. As an important note, mypy refuses to support for `NamedTuple` at all. Since pyglet uses `NamedTuple` in pyglet's `math.Vec*` and `Mat*` classes:
* they mark `pyglet.math` as an excluded module in their mypy config
* we use pyright

arcade/sprite_list/spatial_hash.py
pushfoo: All good. We can clean it up later. Ty!

PR #2651, opened 2025-04-24
Update step_10.rst - Camera line is incorrect

osm3000:
Very simple fix, to change:

```
self.gui_camera = arcade.SimpleCamera(viewport=(0, 0, width, height))
```
to 
```
self.gui_camera = arcade.Camera2D()
```

since there is no `SimpleCamera`


einarf:
Thanks!

PR #2659, opened 2025-05-01
Fix #2658: Use `is not None` for `boundary_left` and `boundary_right`.

### LINKED ISSUE #2658: boundary_left and boundary_right are ignored if they are 0
When set to exactly `0`, `boundary_left` and `boundary_right` are ignored by the `PhysicsEnginePlatformer`. That's because it uses a truthiness test to exclude `None` here:
https://github.com/pythonarcade/arcade/blob/069861b94252e2cf585545eee60a997c2980bc59/arcade/physics_engines.py#L795
Interestingly, it correctly uses `is not None` for `boundary_top` and `boundary_bottom`, just below.

sjrd:
Like was already done for `boundary_top` and `boundary_bottom`.

### DIFF
--- a/arcade/physics_engines.py
+++ b/arcade/physics_engines.py
@@ -792,12 +792,18 @@ class PhysicsEnginePlatformer:
             for platform in platform_list:
                 if platform.change_x != 0 or platform.change_y != 0:
                     # Check x boundaries and move the platform in x direction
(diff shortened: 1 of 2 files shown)


pushfoo:
Looks good, ty!

PR #2695, opened 2025-05-20
Hexagon

benjamin-kirkbride:
```
  _____
 /     \
/       \
\       /
 \_____/
```


einarf:
Let's not re-format all the test 😄

DragonMoffon:
maybe also somewhere in the documentation reference https://www.redblobgames.com/grids/hexagons/? As its a really good primer on a lot of the hex logic that's being used here esp around the coordinate spaces

DragonMoffon:
I'm also confused why this PR seems to include a bunch of work on the platformer tutorial. Is it just a merge issue or are there two PRs accidentally stuck together?

benjamin-kirkbride:
Sorry about the formatting stuff, I was rushing to get this done before leaving PyCon.

I'll remove that stuff

benjamin-kirkbride:
> maybe also somewhere in the documentation reference https://www.redblobgames.com/grids/hexagons/? As its a really good primer on a lot of the hex logic that's being used here esp around the coordinate spaces

All of what I did was based on that, so this is a good idea if someone wants to pick that up and do it

benjamin-kirkbride:
Okay @einarf @DragonMoffon we should be good now. Sorry about the mess lol

DragonMoffon:
Reviewing this again I realized that this isn't really a hexagon shape? I still think the cube coordinate helpers are useful, but we should also make a Hexagon shape that matches with `Rect` (and maybe other shapes like circle etc). I've talked to Digi and they are working on a Shape protocol so we can leave this PR as just the hex coordinate stuff. We may need to play with naming, but that can be done later imo.

benjamin-kirkbride:
Yeah this is for supporting hexagon tile maps. Sorry, could have made that clear

DigiDuncan:
> Yeah this is for supporting hexagon tile maps. Sorry, could have made that clear

I think what I'm building now will support this endeavor, so no harm no foul! Glad you did this work, I think it's good for the library, I'm only worried about naming stuff.

benjamin-kirkbride:
Maybe my object should be HexTile or something?

DigiDuncan:
> Maybe my object should be HexTile or something?

If that's a compromise you're willing to make, I like that.

benjamin-kirkbride:
Hey, where did we end up with this?

Cleptomania:
I think there's still a question of where this fits in with Rect and other future shape work, but I think that can probably be dealt with in the future if/when more work on shapes happens. The tilemap parts of this can be modified to use a new/different one in the future, but as of right now there's not really WIP on that, so I don't think it makes sense to hold this back because of that.

@benjamin-kirkbride can you resolve the conflicts in this?

benjamin-kirkbride:
Yes

benjamin-kirkbride:
@Cleptomania done

PR #2697, opened 2025-05-23
Add a pyopencl demo

wijat:
Using pyopencl enables high-performance large-scale computations, 
such as frequent destruction and creation of interfaces, and large-scale position calculations.

This example demonstrates how to quickly and easily use pyopencl within arcade to run kernel code.


einarf:
Thanks for the example. Would it not be even better if arcade could access the underlying buffer and batch draw this data using the gpu instead? Maybe that's a bit complicated.

wijat:
I've recently started learning how to use the PyOpenCL library and discovered that it can be used alongside Arcade. I noticed that Arcade's built-in physics engine performs its calculations on the CPU using Python, so I decided to try a simple implementation using PyOpenCL. As of now, I don't yet know other ways to access the GPU. I also don't really understand OpenGL at this point.

wijat:
I can understand why you’d say that—installing the PyOpenCL library is actually quite a hassle, and it does create inconvenience for many users. However, I personally find the library extremely useful, and it makes it quite easy to boost the performance of standard Python logic.

If I knew how to use OpenGL, maybe I would go that route, but for now I’m relying on PyOpenCL to improve my game’s performance.

It would be great if there were a more convenient way to integrate this library into Arcade, so that users wouldn’t need to install it manually. After all, installing PyOpenCL requires setting up some precompiled dependencies, and it’s not as simple as just running pip install.

einarf:
Will have to come back to this one. I suspect there are better options like making a simple transform shader instead of using pyopencl. It's actually not that difficult.

The bottleneck in the example is this

```py
for pos in positions_np:
    arcade.draw_circle_filled(pos[0], pos[1], 2, arcade.color.WHITE)
```

Not the particle calculations itself. In order to get C-like performance the particle positions must be calculated and remain on the gpu and then batch rendered.

Possibly we can make an better example there. There are a few examples here: https://github.com/pythonarcade/arcade/tree/development/arcade/examples/gl

wijat:
Yes, this kind of particle system requirement can indeed be achieved without using the pyopencl library. I should start by saying that I only recently began using the Arcade framework to develop my game. I have been working with Unity for many years.

A side note:
While working on my project, I noticed something unexpected—Arcade doesn’t seem to have a built-in node hierarchy or scene graph system. In many frameworks and engines, there is a tree structure of parent and child nodes (e.g., "Node" or "GameObject" hierarchies) where the transform (position, rotation, scale) of a child is relative to its parent. But in Arcade, I have to manually calculate all the transformations for each sprite, even when they are visually or logically attached to others.

Is there any plan to introduce a scene graph or node-based hierarchy system in the future?

If I weren’t currently focused on rapidly developing my game, I would take the time to build a custom node-based framework myself. However, at the moment, I don’t have the bandwidth, so for now, all of my sprites have to manually handle their own transformations.

einarf:
For a node/graph system it's better to make a new issue.

For particles:
If you are interested I can make an example using transforms that probably can do a million smoothly.

wijat:
I'm currently using Arcade to develop my game, and I'm still learning in this area. Honestly, this Pyopencl examples provided seem a bit complex and don't fully solve the problem. So, I'll try to help by offering some feedback and suggestions to improve Arcade.

einarf:
Be free to join the arcade discord server if you need any help with gpu accelerated stuff.

wijat:
I've already joined. Name is The Dev.
图片

PR #2699, opened 2025-05-25
New spritelist rendering system - Support for non-geo shader system

einarf:
Initial work ...

This needs a bit of thinking because we are using index buffer.

### DIFF
--- a/arcade/context.py
+++ b/arcade/context.py
@@ -99,6 +99,19 @@ class ArcadeContext(Context):
         self.sprite_list_program_cull["sprite_texture"] = 0
         self.sprite_list_program_cull["uv_texture"] = 1
 
+        # self.sprite_list_program_no_geo = self.load_program(
+        #     vertex_shader=":system:shaders/sprites/sprite_list_simple_vs.glsl",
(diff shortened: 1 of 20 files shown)


einarf:
So far this experiment is doing the follwing.

* A new `SpriteListData` base class for storing all gpu resources (buffers, geometry, program)
* A `SpriteListBufferData` implementation with the the standard geo shader and buffers

Next step: Attempt to create texture storage version using a program without geo shader.

NOTE: The single sprite shader also needs to be looked at. Likely we just pass in uniform data for this one just like the old version. We don't as little buffer writing as possible causing stalling.

einarf:
TODO

- [x] Fix tests
- [x] Clean up capacity values (buff vs idx)
- [x] Working no-geo shader using textures as per-instance storage
- [x] Wrap texture lookups as size grows
- [x] Maybe: merge pos and angle buffer
- [x] color data should be normalized bytes (f1)
- [x] Fix spritelist rendering customization examples
- [x] Fix gpu sprite collision checks
- [x] Fix `draw_sprite()` functions
- [x] Add uv bias
- [x] Cache geometry in context

einarf:
GPU collision will not be supported in WebGL

Issue #2706, opened 2025-05-30
Bug (Documentation): Broken link for Bouncing Shapes card in Examples page


Infamous003:
## The problem

Clicking on the "Bouncing Shapes" card image takes you to 
```https://api.arcade.academy/en/stable/example_code/shapes-slow.html```, which doesn't exist.

I believe it was intended to take you to the ```shapes-slow``` section of the ```shapes.html```. Not, ```shapes-slow.html```.

## To recreate the problem
- Go to https://api.arcade.academy/en/stable/example_code/index.html#animating-drawing-primitives
- Click on the **image inside** the "Bouncing Shapes" card. 

## Simple fix
- Editing the target attribute in the ```index.rst``` file from this:
```:target: shapes-slow.html``` 
to this 
```:target: shapes.html#shapes-slow```

pushfoo:
Confirmed. If you want to make a quick PR, I'll look at it and merge.

Follow-up for the future:
1. We may want to replace this with pyglet shapes?
2. @DigiDuncan is working on some shape-related ideas at the moment too
3. These galleries have been a brittle point which we should probably wrap it in a plug-in or other generation tooling

PR #2708, opened 2025-05-30
Fix broken link (#2706)

Infamous003:


pushfoo (approved):
Thank you! This is some very good attention to detail! :eyes:

### For future readers

We're linking the element ID in the generated HTML because:

1. Sphinx has a long-standing issue with `:target:` as a reference
   * See https://github.com/sphinx-doc/sphinx/issues/4351)
   * Maybe it will be fixed one day
   * Don't expect it any time soon
2. We need the thumbnail to be clickable now
3. Some example pages _may_ end up holding multiple headings
4. Sphinx uses `<span id="shapes-slow"></span>` and similar as link headings
5. Adding `#shapes-slow` ID targets the specific one in question in a somewhat future-proof way

A better future solution would help, but this is excellent work!

Issue #2266, opened 2024-07-16
Autobuild does not copy some static file changes


pushfoo:
## System Info
Tip of `development`

### Actual behavior:
Sphinx autobuild:
1. detects the change in the CSS file and triggers rebuild
2. does not seem to copy the CSS file

### Expected behavior:
Sphinx autobuild:
1. detects the change in the CSS file and triggers rebuild
2. copies the CSS file

pushfoo:
It looks like thumbnail generation also doesn't fire when those files are updated.

einarf:
It this still a problem?

Infamous003:
It is still a problem. Any changes made to the stylesheets cause the page to reload, but the styles aren't applied.
A painful fix is to do a fresh build, but that takes too much time.

pushfoo:
> It this still a problem?

**TL;DR:** Our pre-3.0 temp fix should work until we test and deploy upstream's fix for their 10+ year bug

## The Quick Fix

> It is still a problem.

@Infamous003 I want you to try the temp fix:

1. `cd` into your Arcade repo's top-level directory
2. `touch .ENABLE_DEVMACHINE_SPHINX_STATIC_FIX`
3. Check if suddenly works

I apologize for not having this documented. Others seemed unbothered by this `_static` issue, so I made it opt-in only. Everyone who uses it might've forgotten it exists.

## Do I need to read the rest?

| I want "funny" | I want code details |
|------------------|-----------------------------|
| Continue reading | See bottom of comment |

Most of this comment is useful levity in the tradition of Arcade repo classics like:

* ["Reinvent the Rectangle"](https://github.com/pythonarcade/arcade/pull/2088)
* ["At the Mountains of Tech Debt"](https://github.com/pythonarcade/arcade/pull/2109)

...So without further ado, it's time for _another_  Python Arcade Story :tm:!

## Oops! All Forced Rebuilds!

> [!WARNING]
> Sit down and pour a `$BEVERAGE` of choice, because this starts 3.4 billion years ago in the [Sphinx](https://github.com/sphinx-doc/sphinx) repo.

### What happened?

1. Sphinx:
    1. had and fixed a bug (yay)
    2. but not in time for Arcade 3.0 (oh no)
2. We fixed our `_static` copy issues like everyone does:
    1. a helper script
    2. an ugly call in `conf.py`
3. It was [supposed to be temporary], so of course people didn't know about it

 
[...]

PR #2709, opened 2025-05-30
Remove temp fix for Sphinx _static file bug

pushfoo:
**TL;DR:** Remove temp fix for [Sphinx `_static` bug][recap] + add `--jobs` flag for `./make.py html` and `./make.py serve`

## Performance?

> [!WARNING]
> This bumps min developer Py to 3.11 and it **needs** testing to ensure performance issues are solved!

In-depth test instructions are located [in a comment below](https://github.com/pythonarcade/arcade/pull/2709#issuecomment-2923591996). Current test status:

- [ ] Windows
- [ ] Mac
- [x] GitHub CI (~ 5m 6s)
- [ ] Linux
  - [x] Debian (~ 10m, may be due disk load on / old hardware)
  - [ ] Ubuntu
  - [ ] Arch 

We can continue using the [temp fix][recap] until we're sure this PR's good.

[recap]: https://github.com/pythonarcade/arcade/issues/2266#issuecomment-2920827018

### Wait, What Changed in Sphinx 8.2.X?

Seems like:

1. Faster partial rebuilds
2. _Much_ slower first builds

I've brought the first-time build down by adding a `--jobs` / `-j` flag pass-through to `./make.py`. It seems to restore perf (now under 10  min, was 45+ minutes), but Sphinx's parallelism might not work correctly on Windows:

> [Windows is not supported. Note that not all parts and not all builders of Sphinx can be parallelized.][no-windows]

[no-windows]: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-j

## What's Changed

* Remove util/sphinx_static_file_temp_fix and invocation in conf.py
* Upgrade to Sphinx 8.2.3 to include fix for temp files
* Add mention of requiring Python 3.11 as comments in pyproject.toml
* Update Python versions for CI doc build (readthedocs + GitHub)
* Explain the contents of the dev deps


pushfoo:
### Test instructions

Please report:

* Your OS
* Your Python version
* Build time
* Whether static copying works
* (If slower than 5 mins) Disk type (SSD, HDD, etc)
 
To get started, first make sure you have Python >= 3.11 in an activated virtual environment.

#### Time a clean build

"Does it run fast enough on clean build?"

1. `git fetch`
2. `git checkout remove-copy-tempfix`
3. `pip install -I -e .[dev]`
4. `./make.py clean`
5. Time execution of first build:
   * For Linux (and likely Mac): `time ./make.py html`
   * For Windows, the PowerShell  _may_ be `Measure-Command { python make.py html }` ([PowerShell doc])

#### Ensure static file copy works

i.e. "Does Sphinx now handle CSS changes correctly?"

1. `./make.py serve`
2. Make a trivial edit to one of [these files](https://github.com/pythonarcade/arcade/tree/development/doc/_static/css)
   * Example: `* { background: red; blue; }`
3. Make sure to save it
4. Watch for `sphinx-autobuild` to detect the change
5. Once done, refresh the brower
6. If the change does not visually show up, inspect the sources in your browser dev tools


[PowerShell doc]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/measure-command?view=powershell-7.5#examples

pushfoo:
```
OS: Debian
Python: 3.13
Build time:
real    9m29.949s
user    30m14.665s
sys     0m18.968s

Static copying works?: Yes
```

eruvanos:
This changes breaks installation via uv:
```
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split
  │ (python_full_version == '3.10.*'):
  ╰─▶ Because the requested Python version (>=3.10) does not satisfy
      Python>=3.11 and sphinx==8.2.3 depends on Python>=3.11, we can conclude
      that sphinx==8.2.3 cannot be used.
      And because arcade[dev] depends on sphinx==8.2.3 and your project
      requires arcade[dev], we can conclude that your project's requirements
      are unsatisfiable.

      hint: The `requires-python` value (>=3.10) includes Python versions
      that are not supported by your dependencies (e.g., sphinx==8.2.3 only
      supports >=3.11). Consider using a more restrictive `requires-python`
      value (like >=3.11).
```

pushfoo:
How did we have anything work in the past?
* `[dev]` required 3.10
* everyday use required only 3.9

eruvanos:
I always had to change the required python to 3.10 and remember to not commit it to github. 
I would love to not restart this again 😔

einarf:
Dev dependencies could be moved out of the toml if needed (or maybe just doc reqs). The issues it creates for uv is pretty serious and we shouldn't break it.

eruvanos:
do we really require 3.11 for docs? Is there no easy way around it? :/

pushfoo:
**TL;DR:** We can make things easier at the cost of tech debt, but we've still gotta deal with Sphinx \:(

> I always had to change the required python to 3.10 and remember to not commit it to github.
> I would love to not restart this again 😔

I think einarf and I have both have ideas to make it better.

### The "Easy" Way?

> Is there no easy way around it? :/

I'll see what I can do to improve our `./make.py`.

[This `./make.py` script I wrote for another project]( https://gist.github.com/pushfoo/51e1722e7a04cf8062cf2c1be636c7de) suggests I can shim in a no-typer mode.

But first, we need to:
1. remove old build output modes
   * They're commented out right now
   * We could replace them with a comment linking the commit that had them
2. #2710

### Why do all that?

It makes room for nice things!

They could include:
1. local patch handing
2. `./make.py interactive-install` :sunglasses:?
3. helping with a `docs/requirements.txt` like pyglet's old-school setup approach

I don't like the last approach as much, but it should work.

### ... The Non-Easy Way?

**TL;DR:** Moving away first from ReST, then from Sphinx. But that's a mega-project of its own.

Cleptomania:
I think this is not a problem anymore because we're on higher python version across the board? If I'm wrong feel free to re-open.

PR #2728, opened 2025-06-14
fix #2676

### LINKED ISSUE #2676: Typing: Use collections.abc instead of typing module
Now that we dropped python 3.9 we can respect the deprecation of the typing module.

We should be using `collections.abc instead. Most members in the typing module are aliases to these anyway.

eruvanos:
huge cleanup

### DIFF
--- a/arcade/gl/backends/opengl/buffer.py
+++ b/arcade/gl/backends/opengl/buffer.py
@@ -151,11 +151,9 @@ class OpenGLBuffer(Buffer):
         # Manually detect this so it doesn't raise a confusing INVALID_VALUE error
         if size + offset > self._size:
             raise ValueError(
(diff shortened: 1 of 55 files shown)


einarf:
Isn't it better to do this in the pyglet3 branch? I'd imagine there would be crazy conflicts but I haven't looked at the diff yet

Cleptomania:
This doesn’t look like it should have too much conflict I don’t think. If the abstraction work hadn’t been merged in prior to the pyglet3 branch there would be, but I think this is fine.

Not sure if there’s a strong incentive to do this prior to pyglet3 being merged but I think it’s fine.

eruvanos:
I am fine with redoing it after the merge :)

PR #2731, opened 2025-06-15
Fix #2727: Only handle space key events when there is a focused widget

### LINKED ISSUE #2727: UIFocusMixin consumes Space
It seems like the focus is not showing if there is no controller, which is good in my opinion, but the space key press event is still catched
even if there is, seems like its still catched and not forwarded
for some reason, the space event is handled instantly:

```python
if isinstance(event, UIKeyPressEvent):
    if event.symbol == arcade.key.TAB:
        if event.modifiers & arcade.key.MOD_SHIFT:
            self.focus_previous()
        else:
            self.focus_next()

        return EVENT_HANDLED

    elif event.symbol == arcade.key.SPACE:
        self._start_interaction()
        return EVENT_HANDLED
```

https://discord.com/channels/458662222697070613/1344768086821175447/1374397523409768468

## Update

- The UIFocusGroup still catches space even if no controller is connected.
- what do you think about renaming UIFocusGroup to UIFocusAnchorLayout and making a UIFocusBoxLayout and UIFocusGridLayout?

narenrameshb:
Fixes #2727

## Changes
- Modified `UIFocusMixin` to only handle space key events when there is a focused widget
- Added checks for `self.focused_widget is not None` to both key press and key release handlers
- This ensures space key events are only captured when appropriate and passed through otherwise

## Testing
- Verified that space key events are only handled when there is a focused widget
- Confirmed that space key events are passed through when no widget is focused
- Tested both key press and key release events

### DIFF
--- a/arcade/gui/experimental/focus.py
+++ b/arcade/gui/experimental/focus.py
@@ -100,12 +100,12 @@ class UIFocusMixin(UIWidget):
 
                 return EVENT_HANDLED
 
-            elif event.symbol == arcade.key.SPACE:
+            elif event.symbol == arcade.key.SPACE and self.focused_widget is not None:
                 self._start_interaction()
                 return EVENT_HANDLED
 
         elif isinstance(event, UIKeyReleaseEvent):
-            if event.symbol == arcade.key.SPACE:
+            if event.symbol == arcade.key.SPACE and self.focused_widget is not None:
                 self._end_interaction()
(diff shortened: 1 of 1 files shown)


eruvanos:
Hi, thank you for looking into this.

Your changes are already covered by line 90.
I guess the issue title is misleading, I am sorry for that.
I have to doublecheck the Discord conversation.

PR #2751, opened 2025-07-06
Add a parameter `create` to `resolve()` to create the directory/file if it doesn't exist.

JackAshwell11:


arcade/resources/__init__.py
```diff
@@ -50,7 +50,19 @@ def resolve_resource_path(path: str | Path) -> Path:
     return resolve(path)
 
 
-def resolve(path: str | Path) -> Path:
+def create_path(path: Path) -> None:
+    """
+    Create a file or directory at the given path.
+    If the path has a suffix, it's treated as a file, otherwise, as a directory.
+    """
+    if path.suffix:
```
DragonMoffon, line 58: Path objects have an `is_dir` method, which is probably a safer check.

arcade/resources/__init__.py
```diff
@@ -87,22 +100,30 @@ def resolve(path: str | Path) -> Path:
...
+                    )
 
             # Always convert into a Path object
-            path = Path(handle_path / resource)
+            path = Path(path)
         else:
             path = Path(path)
 
+    if create:
+        create_path(path)
```
DragonMoffon, line 125: Should this not also check if the path exists?

DragonMoffon (changes requested):
Looks good otherwise.

Since this PR modifies this code already, could we add an additional check. `importlib.resources.Traversable` is also a common type passed into this method, and it doesn't have a consistent interface with `pathlib.Path`. See the [docs](https://docs.python.org/3/library/importlib.resources.html). 

We should either have the function work perfectly with either a `Path` or `Traversable` or check for and change a `Traversable` to a `Path`

JackAshwell11:
> Looks good otherwise.
> 
> Since this PR modifies this code already, could we add an additional check. `importlib.resources.Traversable` is also a common type passed into this method, and it doesn't have a consistent interface with `pathlib.Path`. See the [docs](https://docs.python.org/3/library/importlib.resources.html).
> 
> We should either have the function work perfectly with either a `Path` or `Traversable` or check for and change a `Traversable` to a `Path`

IIRC, there was a discussion in Discord which deemed that there was a better way to accomplish what I was trying to do with the MR without requiring changes to Arcade. This PR should probably be closed now

PR #2752, opened 2025-07-15
Update the arcade window so it resets gl context properly everyframe.

DragonMoffon:
This was discussed a while ago as part of the 3.0 changes. I put it off so that 3.0 could be released, and because there were larger issues with it to be solved. _That work still needs doing_. 

The `ArcadeContext` reset method used by unit tests did not work well. It has now been updated to use the default camera rather than setting the `Projection` and `View` matrices manually.

The GL state was not being reset before every `on_draw` dispatch. This PR makes it so the gl state is reset, the screen's framebuffer is active, and the default camera is the active before every `on_draw`. The consequences of this change on examples and unit tests should be checked before the PR is merged.

### DIFF
--- a/arcade/context.py
+++ b/arcade/context.py
@@ -56,6 +56,7 @@ class ArcadeContext(Context):
         gc_mode: str = "context_gc",
         gl_api: str = "gl",
     ) -> None:
+        
         super().__init__(window, gc_mode=gc_mode, gl_api=gl_api)
 
         # Set up a default orthogonal projection for sprites and shapes
@@ -68,6 +69,7 @@ class ArcadeContext(Context):
         self.current_camera: Projector = self._default_camera
 
         self.viewport = (0, 0, window.width, window.height)
+        
 
         # --- Pre-load system shaders here ---
         # FIXME: These pre-created resources needs to be packaged nicely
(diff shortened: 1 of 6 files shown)


arcade/context.py
```diff
@@ -311,10 +311,7 @@ def reset(self) -> None:
...
         # self.active_program = None
         self.viewport = 0, 0, self.window.width, self.window.height
-        self.view_matrix = Mat4()
-        self.projection_matrix = Mat4.orthogonal_projection(
-            0, self.window.width, 0, self.window.height, -100, 100
-        )
+        self._default_camera.use()
         self.enable_only(self.BLEND)
         self.blend_func = self.BLEND_DEFAULT
         self.point_size = 1.0
```
einarf, line 317: The reset method was more for unit tests. We can still probably trim it down. These lines should definitly not be called every frame but they should be called in conftest between tests except that blending needs to disable instead.

I also don't know if it's necessary to bind the windowblock every frame but that should be done in unit tests.

DragonMoffon:
I've closed this PR, as it was doing two things. There are code fixes for Camera that are here that should be merged, while the context resetting should be handled better

Issue #2754, opened 2025-07-17
Bug: get_angle_radians function in math.py returns incorrect angle due to swapped atan2 parameters


Raahi-18:
The get_angle_radians function has incorrect parameter order in the math.atan2() call at line 454, causing it to return angles rotated by 90 degrees from the expected result.

Incorrect code:
`return math.atan2(x_diff, y_diff)  # Bug: parameters are swapped`

Bug example:
Getting the angle from (0, 0) to (1, 0) currently returns 1.571 radians (90°) but should return 0.000 radians (0°).

This is a simple one line fix just need to swap the parameters for math.atan2():
`return math.atan2(y_diff, x_diff)`

I can make a PR for the update.

Issue #2753, opened 2025-07-17
Bug: rotate_around_point function in math.py returns incorrect results due to wrong translation


Raahi-18:
The rotate_around_point function in math.py has a bug on line 394 where it adds the rotated displacement vector to the target point instead of the source point (rotation center).

The current code is the following:
`return target[0] + dx, target[1] + dy`

Example of the bug:
Rotating point (5, 0) around (3, 0) by 90° currently returns (5, 2) but should return (3, 2).

This is a simple one line fix just have to update return statement to:
`return source[0] + dx, source[1] + dy`

I can make a PR for the update.

Issue #2648, opened 2025-04-15
Bug: rand_vec_spread_deg incorrectly passes degrees to Vec2.from_polar which requires radians


Tobinaer:
The rand_vec_spread_deg function in the Python Arcade library's math module (arcade/math.py) has a bug where it passes angles in degrees to Vec2.from_polar(), which expects angles in radians.

Current, incorrect implementation:

```
def rand_vec_spread_deg(angle: float, half_angle_spread: float, length: float) -> Point2:
    a = rand_angle_spread_deg(angle, half_angle_spread)
    vel = Vec2.from_polar(a, length)  # Bug: 'a' is in degrees but from_polar expects radians
    return vel.x, vel.y
```

Vec2.from_polar's source code:

```
@staticmethod
def from_polar(angle: float, length: float = 1.0) -> Vec2:
    """Create a new vector from the given polar coordinates.

    Args:
      angle: The angle, in radians.  # <-- Expects radians
      length: The desired length
    """
    return Vec2(length * _math.cos(angle), length * _math.sin(angle))
```

Ibrahim2750mi:
The fix should be this? Or am I missing something?

```python
def rand_vec_spread_deg(angle: float, half_angle_spread: float, length: float) -> Point2:
    a = rand_angle_spread_deg(angle, half_angle_spread)
    vel = Vec2.from_polar(a*math.pi/180, length)  # Bug: 'a' is in degrees but from_polar expects radians
    return vel.x, vel.y
```

Raahi-18:
> The fix should be this? Or am I missing something?
> 
> def rand_vec_spread_deg(angle: float, half_angle_spread: float, length: float) -> Point2:
>     a = rand_angle_spread_deg(angle, half_angle_spread)
>     vel = Vec2.from_polar(a*math.pi/180, length)  # Bug: 'a' is in degrees but from_polar expects radians
>     return vel.x, vel.y

To me it seems like that would work, another method could also be using the built in radians() function that's already part of the math library instead of manually converting.

```
def rand_vec_spread_deg(angle: float, half_angle_spread: float, length: float) -> Point2:
    a = rand_angle_spread_deg(angle, half_angle_spread)
    vel = Vec2.from_polar(math.radians(a), length)
    return vel.x, vel.y
```

Barry-Lenhart:
[...]

PR #2755, opened 2025-07-18
Fix angle and rotation bugs in math utils

Raahi-18:
Fixed three bugs in math.py:

1. Fixed `rand_vec_spread_deg` to convert degrees to radians before passing to `Vec2.from_polar`. #2648 
2. Fixed `rotate_around_point` to properly rotate around the `source` point instead of the `target`. #2753 
3. Corrected argument order in `math.atan2` in `get_angle_radians` function, which previously caused incorrect angle calculations. #2754

### DIFF
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -289,7 +289,7 @@ def rand_vec_spread_deg(angle: float, half_angle_spread: float, length: float) -
         length (float): The length of the vector
     """
     a = rand_angle_spread_deg(angle, half_angle_spread)
-    vel = Vec2.from_polar(a, length)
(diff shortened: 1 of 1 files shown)


DragonMoffon (changes requested):
`rand_vec_magnitude` should also convert to radians, as arcade operates almost wholly on degrees, so this function expecting radians breaks convention

~~Also, it would be good to add unit tests for all of these functions. Preferably a unit test for every function in `arcade.math`, but that's out of the scope of this PR.~~

Having checked the `arcade.math` tests they are sorely lacking and can be fixed in a later PR

Cleptomania:

PR #2759, opened 2025-08-13
Add .rect to Text

DigiDuncan:
What it says on the tin. Adds a `.rect` property to Text objects.

### DIFF
--- a/arcade/text.py
+++ b/arcade/text.py
@@ -581,7 +581,14 @@ class Text:
 
     @property
     def rect(self) -> Rect:
-        """Rect representing the bounds of the text."""
+        """Rect representing the bounds of the text.
+
+        .. tip:: Don't worry about `width` being `None`.
+
+            Although a label can be created with a `width=None`:
+            * The underlying :py:mod:`pyglet` label will have bounding dimensions
+            * This rect is for on-screen click and layout purposes, not maximum possible width
+        """
         return LRBT(self.left, self.right, self.bottom, self.top)
 
     @property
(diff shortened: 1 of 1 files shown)


arcade/text.py
```diff
@@ -578,6 +579,11 @@ def bottom(self) -> float:
         """Pixel location of the bottom content border."""
         return self.label.bottom
 
+    @property
+    def rect(self) -> Rect:
+        """Rect representing the bounds of the text."""
+        return LRBT(self.left, self.right, self.bottom, self.top)
+
```
pushfoo: What happens when there is no valid width? It can be `None`:

https://github.com/pythonarcade/arcade/blob/61248f2b5d381eb08bfc69c6df3bbb470c7f9c01/arcade/text.py#L752-L756

If raising an exception is the end goal, that's fine if we:

1. Document it
2. Add a test for it

Text is already scary, so we should at least label the behavior.

pushfoo (dismissed):
I like this, but I think it needs more doc at minimum since width can be `None` for a label as I understand it.

arcade/text.py
DigiDuncan: From what I can tell, `.left` and `.right` are calculated from Pyglet's [`TextLayout`](https://pyglet.readthedocs.io/en/latest/modules/text/layout.html#pyglet.text.layout.TextLayout.left) which guarantees a return type of `float` and can not be `None`.

I addressed the concern in the docstring, thanks for the phrasing.

Issue #2778, opened 2025-10-08
InputManager raises DeviceOpenException when controller is already opened


MiCurry:
The `InputManager` will raise a `DeviceOpenException` when a Controller is already opened. This can occur
when an `InputManager` is instantiated twice. 

A game could have a main-menu screen which has multiple game views each with it's own specific `InputManager`.  Users may venture into one game view, and then into another, which would instantiate multiple `InputManagers`. We probably should not have to bind/rebind controllers between these switches, instead, `InputManager` should just check if the device is open before attempting to open it.

The following game contains a minimal example of this. It has a Main Menu Screen, and a Game View. After showing the game view, going to the main menu and back into the game view, the error occurs (if a controller is connected). To recreate the error in the example below, do the following:

1. Start the game 
2. Ensure your controller is connected
3. Hit Play
4. Press Escape/Start
5. Press Play again -> `DeviceOpenException` occurs

```Python
import pyglet
import arcade
from arcade.future.input import ActionState, InputManager, Keys, ControllerAxes, ControllerButtons
import arcade.gui
import arcade.gui.widgets
import arcade.gui.widgets.buttons
import arcade.gui.widgets.layout

WIDTH = 800
HEIGHT = 800

class Player(arcade.Sprite):
    def __init__(self, sprite_file, controller):
        super().__init__(sprite_file)
        self.input_manager = InputManager(controller=controller, action_handlers=self.on_action)
        self.input_manager.new_action("pause")
        self.input_manager.add_action_input("pause", Keys.ESCAPE)
        self.input_manager.add_action_input("pause", ControllerButtons.START)
        self.input_manager.new_axis("move_horz")
        self.input_manager.add_axis_input("move_horz", Keys.LEFT, scale=-1.0)
        self.input_manager.add_axis_input("move_horz", Keys.RIGHT, scale=1.0)
        self.input_manager.add_axis_input("move_horz", ControllerAxes.LEFT_STICK_X, scale=1.0)
        self.physics_engine = arcade.PhysicsEngineSimple(self)

    def on_action(self, action: str, state: ActionState):
        if action == 'pause' and state == ActionState.PRESSED:
            main_menu = MainMenu()
            window.show_view(main_menu)

    def update(self, delta_time: float = 1/60) -> None:
        self.input_manager.update()
        self.center_x += self.input_manager.axis("move_horz") * 20 * delta_time

class Game(arcade.View):
    def __init__(self):
        super().__init__()
        self.player_list = None
[...]

PR #2779, opened 2025-10-08
InputManager - Check if Controller device is open before opening

MiCurry:
Before this merge, the `__init__` of the `InputManager` could fail if a controller had already been open. This could occur if an `InputManager` was instantiated twice, see #2778.

This commit closes issue #2778.


Cleptomania:
Looks good, I thought about potentially having the InputManager close the controller it's responsible for, but that seems like it'd be inflexible and just create more problems.

PR #2784, opened 2025-10-14
Add NoArcadeWindowError subclassing RuntimeError

pushfoo:
Per Discord discussion with @DragonMoffon:

> Please can we. its also super useful non daemon threads. If i have a worker thread open for file IO having it close if it catches a no window error would be great
> also a `does_window_exist` function would be nice

The function name is `have_window` in this PR but otherwise it's the same idea.

Changes: 
1. Add `arcade.exceptions.NoArcadeWindowError`
   - subclasses `RuntimeError`
   - no args, just a semantic wrapper
2. In `arcade.window_commands`:
   - Make `get_window()` raise `NoArcadeWindowError` when no window exists
   - Touch up old-style `:returns:` docstring for `get_window()`
   
Deferred until a later PR:
- Add `has_window()` somewhere
  - has/have bikeshedding
  - `window_commands` may be worth moving (especially if we add multi-window support?)

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -90,6 +90,7 @@ from .window_commands import set_window
 from .window_commands import start_render
 from .window_commands import unschedule
 from .window_commands import schedule_once
+from .window_commands import window_exists
 
 from .sections import Section, SectionManager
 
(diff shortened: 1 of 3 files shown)


DragonMoffon (commented):
Documentation comment has the wrong function name, but this is otherwise good

PR #2789, opened 2025-10-20
Add a split screen example using Camera2D

MiCurry:
This is a simple example of how to create a split screen example using Camera2D and uses the PyMunk physics engine.

In the example, we create two cameras, and change their `viewports` to be half of the screen. When drawing the scene, we loop through each camera and draw our scene, then in `update()` we center the cameras on the players. We also use the default camera to draw a divider.

It also contains other useful examples, such as zooming out the camera, and resizing the split-screens when the window is resized.


MiCurry:
I wanted to suggest this example as I believe a split screen example would be popular. This example might be better instead if it uses the simple physics engine and if uses the same 'map' as the other examples in the Camera section (e.g. [Sprite-move-scrolling](https://api.arcade.academy/en/development/example_code/sprite_move_scrolling.html#sprite-move-scrolling)).

Happy to change it to use the simple physics engine, and make any other changes!

Cleptomania:
@MiCurry I think I'd like to see a non pymunk version with a similar map maybe, but there's no such thing as too many examples really, so there's no reason we can't keep the existing one too.

Perhaps the existing one in this PR gets renamed to like `camera2d_splitscreen_pymunk` and add a new one with the simpler physics?

MiCurry:
> @MiCurry I think I'd like to see a non pymunk version with a similar map maybe, but there's no such thing as too many examples really, so there's no reason we can't keep the existing one too.
> 
> Perhaps the existing one in this PR gets renamed to like `camera2d_splitscreen_pymunk` and add a new one with the simpler physics?

@Cleptomania that sounds like a good idea. I'll see if I can make an additional example without pymnk for us!

Cleptomania:
I'm just going to go ahead and merge this, if we wanna follow up with a non pymunk version in another PR feel free.

PR #2797, opened 2025-12-04
Revise README for make.py project task runner

AlexBond23:
Updated README to focus on make.py task runner and its features.


Cleptomania:
The README isn't really intended for contributor documentation, that's the job of CONTRIBUTING.md which mostly already covers usage of `make.py`. Regardless, I don't think we want to remove all the project info from the README.

PR #2798, opened 2025-12-06
Fix broken link in procedural caves BSP

MiCurry:
It appears rougelikedevelopment.org is no more. Perhaps it is now rougebasin.com? Regardless, this replaces the dead link with a good one.

I am not familiar with the original link, but I presume this one is very similar? Perhaps someone remembers?


pvcraven:
Don't think that was the original, but it works.

PR #2802, opened 2025-12-26
Update the `DefaultProjector` to behave better when changing framebuffers

DragonMoffon:
The `DefaultProjector` was supposed to update itself to give a consistent unnoticeable behavior on the backend when switching between framebuffers. This has not been the case since the creation of the 3.0 Camera system.


Cleptomania:
I'm going to go ahead and merge this for now. If it needs more work we can follow up with another PR, but it currently fixes an issue with WebGL so I wanna get it in.

PR #2807, opened 2025-12-27
Disable WebGL Alpha Channel on the backing buffer

Cleptomania:
This PR depends on https://github.com/pyglet/pyglet/pull/1400 in Pyglet to be merged first, and will error without it.

WebGL by default has an alpha channel on the backing buffer which causes the content of the HTML/CSS behind the canvas element to bleed through when doing alpha blending. On desktop OpenGL/GLES the backing buffer is solid black and doesn't have an alpha channel.

This change makes the WebGL context function more in line with how desktop OpenGL does

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -63,6 +63,12 @@ if os.environ.get("ARCADE_TEST"):
 else:
     pyglet.options.dpi_scaling = "stretch"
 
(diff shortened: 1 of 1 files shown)


Cleptomania:
This is being done by default in Pyglet now, and the way to change these settings has changed anyways.

PR #2811, opened 2025-12-28
Create an example for the easing functions

DigiDuncan:
This roughly replicates the functionality of the original Easing Example 1.


DigiDuncan:
Upgrading this from draft as I'm not sure what else we need to do with it. Open to discuss!

Cleptomania:
@DigiDuncan somehow this one slipped through without getting merged, not sure why. We got easing examples that are largely the same from Paul in #2852. I'm going to close this for now just assuming the same things are covered, if you think it needs re-opened feel free to or make a new PR extending those ones, whatever works.

DigiDuncan:
I do believe my version covers some things that the original examples didn't, is this something that we can return to?

Cleptomania:
Yeah I'm fine if you wanna add these as new ones, or add to the other existing ones, whatever works, just seemed like it was mostly covering the same things but I don't actually have any familiarity with this module so I'm fine with whatever, can't have too many examples.

Issue #2785, opened 2025-10-15
Missing unit tests in `arcade.math`


DragonMoffon:
Most of the methods in `arcade.math` are missing unit tests.

I also suspect many other functions in arcade are missing unit tests.

`arcade.math` function tests:
- [ ]     clamp
- [x]     lerp
- [x]     lerp_2d
- [x]     lerp_3d
- [x]     lerp_angle
- [ ]     rand_in_rect
- [ ]     rand_in_circle
- [ ]     rand_on_circle
- [ ]     rand_on_line
- [ ]     rand_angle_360_deg
- [ ]     rand_angle_spread_deg
- [ ]     rand_vec_spread_deg
- [ ]     rand_vec_magnitude
- [ ]     get_distance
- [ ]     rotate_point
- [ ]     get_angle_degrees
- [ ]     get_angle_radians
- [ ]     quaternion_rotation

DragonMoffon:
Nearly every method has a test_function, but they are just smoke tests and do not test functionality.

Pteris-211:
Hi, I'd like to work on it over the next few weeks. I'm new to the open source world, so if there's anything I need to pay particular attention to, please let me know.

AidenHogan:
Hello, I was drafting tests for this when I stumbled across #2812.  It seems Arcade's math is currently being overhauled to mostly use Pyglet, and thus PRs for `arcade.math` are not being accepted right now.  Perhaps this issue should be closed while that's in the work?

PR #2812, opened 2026-01-02
Math tests - issue #2785

Pteris-211:


Cleptomania:

PR #2813, opened 2026-01-03
#issue2770 we could add math functions like this and still use the pr…

### LINKED ISSUE #2770: Inconsistency in math.py
### I've noticed that in math.py there’s an inconsistency in how points are handled:

- Some functions, like `rotate_point` and `get_distance`, take x and y as separate parameters:
```
def get_distance(x1: float, y1: float, x2: float, y2: float) -> float
def rotate_point(x: float, y: float, cx: float, cy: float, angle_degrees: float)
```

- Others, like `rand_on_line`, use `Point2`:

```
def rand_on_line(pos1: Point2, pos2: Point2) -> Point
```

This can be confusing for users of the library and makes it harder to work consistently with points. It might be better to standardize all math functions to use Point2 object.

RydertHuGlIfE:
Fixes #2770

…evious one for backwards compatiblity here's my implementation with some test cases as well they pass

This is just done for one function for now to be checked my maintainers 
if the fix is deemed well enough by the maintainers I could standardize the other functions as well

### DIFF
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -336,38 +336,58 @@ def get_distance(*args) -> float:
     else:
         raise ValueError("get_distance() takes 2 or 4 arguments")
 
+
+@overload
 def rotate_point(
     x: float,
     y: float,
     cx: float,
     cy: float,
     angle_degrees: float,
-) -> Point2:
-    """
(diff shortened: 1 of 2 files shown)


Cleptomania:

PR #2816, opened 2026-01-10
Update README.md with setup instructions for local development

pvcraven:


webplayground/README.md
```diff
@@ -5,6 +5,15 @@ An http server is provided with the `server.py` file. This file can be run with
 
 The index page will provide a list of all Arcade examples. This is generated dynamically on the fly when the page is loaded, and will show all examples in the `arcade.examples` package. This generates links which can be followed to open any example in the browser.
 
+## Setup
+
+1. Create virtual environment `python -m venv venv`
```
eruvanos, line 10: should we maybe move over to uv?

1. uv sync
2. Install more libs: uv pip install ...

Cleptomania:
I think this is somewhat defunct now, and would probably be better addressed with a more holistic revamp of that readme? Or inclusion of that readme into top level contributor documentation or something.

PR #2822, opened 2026-02-07
Input Manager Finalization - Docs and Cleanup

Cleptomania:
This PR contains a finalization of the new input system for 4.0. The original system has been in the codebase for a while already(was in the `future` package during 3.X). This PR mostly exists to actually create documentation for it, and potentially handle small cleanups of the module code itself.

### DIFF
--- a/arcade/__init__.py
+++ b/arcade/__init__.py
@@ -145,6 +145,16 @@ if not headless:
     from .controller import ControllerManager
     from .controller import get_controllers
 
+from .input import ActionState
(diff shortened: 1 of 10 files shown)


Cleptomania:
Going to go ahead and merge this, the docs have a start but could maybe use another pass. There is a basic sprite_move example for the input manager now, and the API/docstrings are mostly finished for it.

Main thing that still needs to be done is updating platformer tutorial to make use of it

PR #2827, opened 2026-02-16
Fix: Jump state reset and multi-jump spam in PhysicsEnginePlatformer

ShahrukhAbrar:
Fixes # 1848

- Added `jump_cooldown` to `PhysicsEnginePlatformer`.
- Fixed grounded check logic in `can_jump`.

### DIFF
--- a/arcade/physics_engines.py
+++ b/arcade/physics_engines.py
@@ -4,6 +4,7 @@ Physics engines for top-down or platformers.
 
 import math
 from collections.abc import Iterable
+import time
 
 from arcade import (
     BasicSprite,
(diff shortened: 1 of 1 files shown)


Cleptomania:
I wonder if we should consider adding `delta_time` as a parameter to the `update` method on Physics Engines such that we can use that for time tracking as opposed to `time.time()`. The built-in time function is known to have precision problems and can't really be relied on in increments less than 1 second.

That's gonna incur a lot of update work on documentation/example code though, so that's perhaps out of band of this PR. We could default it to `1 / 60` to keep backwards compat, but given we're working on 4.0 versions right now, I think it's best to break the API and assume that when using the built-in physics engines you need to passthrough the delta_time, at least for something like the Platformer one and any equally or more advanced than that.

ShahrukhAbrar:
Yeah that makes sense, Since I am already in this part of the code, would you like me to refactor the update method to accept ` delta_time`  and update the jump tracking logic to use it, or would you like to handle this in a separate PR?

PR #2828, opened 2026-02-21
New CLI Infrastructure

Cleptomania:
This adds some new infrastructure to give us the ability to create a more robust CLI for Arcade.

Previously all we really had was that if you ran `arcade` you would get a printout of basic installation/system info. This printout has now been moved to the `info` command(and has also been updated to work properly with the abstracted context class so each backend can provide it's own info).

Broadly this gives us the ability to add new commands to a CLI that ships with Arcade. One potential idea is an `init` command which would bootstrap a skeleton Arcade project for you, which could take the place of a potential cookiecutter repo or similar(or work alongside it). The need for some kind of structured project bootstrap is increased with supporting web browsers, so this is one way we can help make that easier.

Note this PR doesn't actually add any new commands, it just puts in place some basic infrastructure to facilitate it. Given that this is intended for end-users and not Arcade contributors, I didn't want to add any extra dependencies to support this, so this is a barebones CLI implemented with argparse, as opposed to using something like click, typer, or rich.


einarf:
Great to see more work on this. There's a lot of things that can be done with the cli. Some useful built in tooling and user-created commands for baking assets and whatnot.

eruvanos (approved):
I like the structure and better possibility for new commands

PR #2829, opened 2026-03-01
Update tilemap to respect object visibility settings

eruvanos:
@Cleptomania do you see any issue with this change? It is kind of breaking, but only in rare cases.


Cleptomania:
I think it makes sense, I don't know if I'd call it a breaking change necessarily, really this is more correct, but it's possible someone could be relying on the broken functionality and they'd need to adjust.

Issue #2821, opened 2026-02-06
Rework the Texture section in the programming guide


pushfoo:
**TL;DR:** Needs clearer wording, cross-refs, and better organization.

## What Needs Work?

https://api.arcade.academy/en/development/programming_guide/textures.html#custom-textures

## Introduction

In https://api.arcade.academy/en/development/programming_guide/textures.html#introduction :

- [ ] Choose a better phrase than "sprite engine" (we don't use the word elsewhere in the doc or code base)

> There is another texture type in Arcade in the lower level OpenGL API: [arcade.gl.Texture2D](https://api.arcade.academy/en/development/api_docs/gl/texture.html#arcade.gl.Texture2D). This represents an actual OpenGL texture and should only be used when dealing with the low level rendering API.
> 
> Textures can be created/loaded before or after the window is created because they don’t interact with OpenGL directly.

- [ ] Clearly define the separate texture layers:
  - [ ] Software / RAM (Arcade)
  - [ ] pyglet
  - [ ] OpenGL / The GPU

In https://api.arcade.academy/en/development/programming_guide/textures.html#texture-uniqueness :

> When a texture is created a `hash/name` is required. This should be a unique string. If two more more textures have the same name we will run into trouble.
>
> When loading textures the absolute path to the file is used as part of the name including vertical/horizontal/diagonal, size and other parameter for a truly unique name.

- [ ] Add clear definition of the required data
- [ ] Explain what "trouble" means
- [ ] Explain what "unique" means
  - [ ] Cross-reference path handling
  - [ ] Briefly clarify the algorithm with a cross-reference to the source file
- [ ] Explain the flipping options
## Cross-Refs

In https://api.arcade.academy/en/development/programming_guide/textures.html#texture-cache:
- [ ] What attribute? Where?
  <img width="699" height="43" alt="Image" src="https://github.com/user-attachments/assets/a4fd6b6a-9828-4373-adfc-c12686835230" />

- [ ] This should cross-ref https://api.arcade.academy/en/development/api_docs/api/cache.html#arcade.cache.TextureCache.flush :
  <img width="646" height="45" alt="Image" src="https://github.com/user-attachments/assets/66dc7bfa-4f86-4711-a858-2c787e694c44" />

MinhBDoan:
Hi! I'd like to work on this as my first contribution. Is it still available?

einarf:
Pretty sure there's always something to fix in docs. It's a never ending project 😆

[...]

PR #2832, opened 2026-03-06
Improve Textures guide (#2821)

MinhBDoan:
This PR addresses all requested improvements from issue #2821:

- Reworded "sprite engine" to describe hit box usage
- Added Texture Layers section (Arcade / pyglet / OpenGL)
- Clarified Texture Uniqueness:
  - Defined what "unique" means
  - Explained consequences of duplicate names
  - Explained flipping options
  - Added cross-reference to arcade.load_texture
  - Referenced arcade/cache/texture.py
- Updated Texture Cache section:
  - Clarified name/hash attribute
  - Cross-referenced TextureCache class and flush() method
  - Included example code

### DIFF
--- a/doc/programming_guide/textures.rst
+++ b/doc/programming_guide/textures.rst
@@ -8,9 +8,8 @@ Introduction
 The :py:class:`arcade.Texture` type is how Arcade normally interacts with
 images either loaded from disk or created manually. This is basically a
 wrapper for PIL/Pillow images including detection for hit box data
-using pymunk depending on the selected hit box algorithm. These texture
-objects are in other words responsible to provide raw RGBA pixel
-data to OpenGL and hit box geometry to the sprite engine.
+using pymunk depending on the selected hit box algorithm. These texture objects are responsible for providing raw RGBA pixel
+data to OpenGL and hit box geometry used when rendering sprites.
 
 There is another texture type in Arcade in the lower level
 OpenGL API: :py:class:`arcade.gl.Texture2D`. This represents an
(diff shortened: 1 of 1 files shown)


MinhBDoan:
“This PR addresses all requested changes from #2821. First-time contribution, feedback welcome!”

pvcraven:
Claude review:

## Review: Improve Textures guide (#2832)

Thanks for tackling #2821 — the restructuring (Texture Layers section, expanded Uniqueness/Cache sections) is a good shape and hits the checklist. However, I traced the new technical claims against the current source and found a few factual inaccuracies that should be fixed before merge, since the whole point of this PR is correctness/cross-refs.

### 🔴 Issues to fix

1. **Wrong class referenced for `default_texture_cache`.**
   ```rst
   Arcade stores texture instances in the default texture cache
   (:py:class:`arcade.cache.TextureCache`). To clear all cached textures,
   use the :py:meth:`arcade.cache.TextureCache.flush` method:
   ```
   `default_texture_cache = TextureCacheManager()` (`arcade/texture/__init__.py:19`) — it's a `TextureCacheManager`, not a `TextureCache`. The manager wraps a `TextureCache` internally, but `default_texture_cache.flush()` actually calls `TextureCacheManager.flush()` (`arcade/texture/manager.py:64`), which by default also flushes sprite sheets and image data (not just texture entries, and not hit boxes). `arcade.TextureCacheManager` is already a documented Sphinx target (`doc/api_docs/api/texture.rst:25`) — use that instead:
   ```rst
   Arcade stores texture instances in the default texture cache
   (:py:class:`arcade.TextureCacheManager`). To clear all cached textures,
   use the :py:meth:`arcade.TextureCacheManager.flush` method:
   ```

2. **`flipped_vertically=True` doesn't exist.** Grepped the whole codebase — no such parameter/kwarg anywhere. The real API is the zero-arg instance method `texture.flip_vertically()` (`arcade/texture/texture.py:482`), called *after* loading, not a load-time flag.

3. **"the underlying pixel data changes" is incorrect.** `Texture.transform()` (`arcade/texture/texture.py:564-596`), which backs all flip methods, passes the *same* `self.image_data` into the new `Texture` — only hit box points and a GPU-side draw transform differ. This directly contradicts `flip_vertically()`'s own docstring ("returns a new texture with the same image data"). Suggest rewording to something like: flipped textures get a different hit box (computed from the transformed pixel data) even though the underlying image data is shared.

4. **`arcade.load_texture` is probably the wrong cross-ref for cache-naming.** `arcade.load_texture` (`arcade/texture/loading.py:12`) is explicitly docstring'd as "Load a texture from disk (**no caching**)." The unique-name/cache-key logic (`Texture.create_image_cache_name`, `arcade/texture/texture.py:282`) is used by `TextureCacheManager.load_or_get_texture` / `_load_or_get_texture`, not by plain `load_texture`. Worth pointing at `default_texture_cache.load_or_get_texture` instead. (Also: `create_image_cache_name` only combines `path` + `crop`, not flip state — another sign this is the wrong function to cite for "name includes flipping.")

### 🟡 Nit

- Trailing double-spaces (Markdown-style hard breaks — meaningless in RST) on the `**Arcade (Python / RAM)**` and `**pyglet**` bullet lines. The repo's `trailing-whitespace` pre-commit hook (`.pre-commit-config.yaml:10`) will likely flag these.

### Overall

Structure and intent are good; recommend re-verifying each new technical claim against `TextureCacheManager` / `TextureCache` / `Texture.transform` source directly rather than paraphrasing the issue description, then this should be good to merge.

PR #2835, opened 2026-03-12
Rework HitBox to support multiple named collision regions

Cleptomania:
## Summary
- Replaces the single-polygon HitBox with a multi-region system where each HitBox can contain named `Point2List` regions (e.g. separate head/body/feet hitboxes), while maintaining backward compatibility with the single-region API
- Adds per-region collision channel/mask filtering using bitmasks. Each region has a `channel` (what it is) and `mask` (what it collides with). Collision requires bidirectional agreement: `(A.channel & B.mask) != 0 AND (B.channel & A.mask) != 0`
- Provides `channels()` utility to convert 1-based channel numbers to bitmasks (e.g. `channels(1, 3)` = `0b101`)
- Introduces `RawHitBoxRegion` TypedDict and bumps serialization to v2 format (with v1 backward compat), supporting `channel` and `mask` fields per region
- Merges `RotatableHitBox` into `HitBox` and adds JSON serialization support (with optional gzip compression)
- Updates collision detection, pymunk physics engine, sprite lists, and tilemap to work with multi-region hitboxes
- Normalizes all points to tuples of tuples internally for immutability and consistency, while still accepting any valid sequence as input
- Updates `sprite_multi_hitbox` example to demonstrate channel-based collision filtering (body catches coins on channel 1, shield catches gems on channel 2)
- Expanded test coverage including 15 new channel/mask tests

## Test plan
- [ ] Run hitbox unit tests (`tests/unit/hitbox/test_hitbox.py`) — all 34 pass
- [ ] Run GUI tests to verify formatting fixes don't break anything
- [ ] Verify pymunk physics engine works with multi-region hitboxes
- [ ] Test the `sprite_multi_hitbox` example (coins only hit body, gems only hit shield)
- [ ] Confirm backward compatibility with single-region hitbox usage
- [ ] Confirm v1 serialization format loads correctly in v2 code

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/arcade/hitbox/base.py
+++ b/arcade/hitbox/base.py
@@ -4,14 +4,21 @@ import gzip
 import json
 from math import cos, radians, sin
 from pathlib import Path
-from typing import Any
+from typing import Any, TypedDict
 
 from PIL.Image import Image
 from typing_extensions import Self
 
 from arcade.types import EMPTY_POINT_LIST, Point2, Point2List
 
-__all__ = ["HitBoxAlgorithm", "HitBox"]
+__all__ = ["HitBoxAlgorithm", "HitBox", "RawHitBox"]
+
+
+class RawHitBox(TypedDict):
+    """Typed dictionary representing the serialized form of a :py:class:`HitBox`."""
+
+    version: int
+    regions: dict[str, Point2List]
 
 
 class HitBoxAlgorithm:
@@ -134,14 +141,21 @@ class HitBox:
 
         loaded = HitBox.load("hitbox.json")
 
-        # Dict round-trip
+        # Dict round-trip (see RawHitBox for the schema)
         data = box.to_dict()
         copy = HitBox.from_dict(data)
+
+    .. note::
+
+        All points are normalized to tuples of tuples on construction.
+        Any sequence type is accepted as input, but regions will always
+        store tuples internally.
 
     Args:
         points:
             Either a single ``Point2List`` (creates a ``"default"`` region)
             or a ``dict[str, Point2List]`` mapping region names to point lists.
+            Points are normalized to tuples on storage.
         position:
             The center around which the points will be offset.
         scale:
@@ -160,9 +174,11 @@ class HitBox:
         angle: float = 0.0,
     ):
         if isinstance(points, dict):
-            self._regions: dict[str, Point2List] = dict(points)
+            self._regions: dict[str, Point2List] = {
+                name: tuple(tuple(p) for p in pts) for name, pts in points.items()
+            }
         else:
-            self._regions = {self.DEFAULT_REGION: points}
+            self._regions = {self.DEFAULT_REGION: tuple(tuple(p) for p in points)}
 
         self._position = position
         self._scale = scale
@@ -205,7 +221,7 @@ class HitBox:
             name: The name for the new region.
             points: The polygon points for the region.
         """
-        self._regions[name] = points
+        self._regions[name] = tuple(tuple(p) for p in points)
         self._is_single_region = len(self._regions) == 1
         self._adjusted_cache_dirty = True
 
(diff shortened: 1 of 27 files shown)


Cleptomania:
Right now this has a re-work that gets rid of the split between HitBox and RotatableHitBox. Originally this split was decided on because hitboxes are expensive to rotate, but in practice we've not really realized any gains from this, and everything is nearly universally always using RotatableHitBox anyways, so I think this split is not worth the headaches it has caused.

In addition to removing that split, it supports multiple polygons per hitbox, and they can be named. Yet to be done is to allow the collision detects to report what regions were collided, so you could do things like check if a `head` region or an `arm` region were hit. Obviously using multiple regions will incur a greater performance hit, as the check is a similar cost per region, but this can potentially save users from needing to manage things by composing a character or something out of multiple sprites.

The new hitboxes also support being loaded from JSON, I have plans to build an editor for creating these points, but that is not done yet.

Another thing yet to be done, is to update the Tilemap loading to support loading multiple regions from Tiled objects, this is something Tiled supports but Arcade has historically ignored. I will have to investigate how the region naming should work for loading these, but I believe the objects in Tiled can be given names, so that will probably be the way to go.

arcade/hitbox/base.py
```diff
@@ -92,98 +95,194 @@ def create_bounding_box(self, image: Image) -> Point2List:
...
 
     @position.setter
     def position(self, position: Point2):
         self._position = position
         self._adjusted_cache_dirty = True
 
-    # Per Clepto's testing as of around May 2023, these are better
-    # left uncached because caching them is somehow slower than what
-    # we currently do. Any readers should feel free to retest /
-    # investigate further.
```
eruvanos, line 151: ❤️

arcade/hitbox/base.py
```diff
@@ -199,127 +298,138 @@ def scale(self, scale: tuple[float, float]):
...
-    ):
-        super().__init__(points, position=position, scale=scale)
-        self._angle: float = angle
+    # --- Serialization ---
 
-    @property
-    def angle(self) -> float:
-        """
-        The angle to rotate the raw points by in degrees
+    def to_dict(self) -> dict:
```
eruvanos: Does it make sense to create a TypedDict 🤔

arcade/hitbox/base.py
Cleptomania: Yeah I think on this one, previously it was faster not to cache because the base HitBox didn’t take rotation into account, so it was really simple to calculate those fields, but with undoing the split for rotation that’s no longer the case, so using the cache there is faster, that’s my guess anyways. I intend to do some benchmarking against the development branch at some point when this is nearer to its final form.

arcade/hitbox/base.py
Cleptomania: Yeah I think it probably does, I’m a big fan of TypedDict for things that are being saved/loaded. Especially if that format somehow gets expanded to have more metadata that an editor might make use of.

Cleptomania:
I removed a commit for experimenting with a channel/mask based collision system similar to Godot's, I still think it's an interesting idea but I think it only makes sense if we completely revisit physics/collision control in Arcade. It's still largely up to the user when to actually perform collision checks and what things they perform them against, and the channel/mask system only feels like it really makes sense if Arcade were just constantly running a physics/collision system that users could hook into.

Cleptomania:
I'm not sure where this is going, this is kind of an experiment, I don't love the results for inclusion in Arcade yet, maybe can make a new run to JUST support multiple hitboxes easier? I'm not sure, I'm going to close this for now because I think this can be done better or needs thought through what we actually want more. Can use this work as a reference in the future, but I don't think a massive hitbox re-work is gonna be on the table for 4.0 at this point.

PR #2842, opened 2026-04-20
Update pyglet to 3.0.dev3

pvcraven:
## Summary
- Updates pyglet dependency from `3.0.dev2` to `3.0.dev3` (released April 18, 2026)
- Adapts to breaking API changes:
  - `pyglet.config.OpenGLConfig` → `pyglet.config.OpenGLUserConfig` with new `api` parameter using `GraphicsAPI` enum (replaces string-based `opengl_api`)
  - `pyglet.media.load()` split into `pyglet.media.load_audio()` / `pyglet.media.load_video()`
  - `pyglet.gl` module relocated to `pyglet.graphics.api.gl`

## Test plan
- [x] Full test suite: 1175 passed, 4 failed (all 4 pre-existing — 3 HiDPI pixel scaling, 1 encoding issue)
- [x] No new test failures introduced by this update
- [ ] Verify on non-HiDPI display for pixel-perfect tests
- [ ] Test sound loading/playback
- [ ] Test video playback (if applicable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/arcade/examples/gui/exp_controller_inventory.py
+++ b/arcade/examples/gui/exp_controller_inventory.py
@@ -18,7 +18,7 @@ python -m arcade.examples.gui.exp_controller_inventory
 
 import pyglet.font
 from pyglet.event import EVENT_HANDLED
-from pyglet.gl import GL_NEAREST
+from pyglet.graphics.api.gl import GL_NEAREST
 from pyglet.input import Controller
 
 import arcade
--- a/arcade/examples/gui/exp_controller_inventory.py
+++ b/arcade/examples/gui/exp_controller_inventory.py
@@ -18,10 +18,10 @@ python -m arcade.examples.gui.exp_controller_inventory
 
 import pyglet.font
 from pyglet.event import EVENT_HANDLED
-from pyglet.graphics.api.gl import GL_NEAREST
 from pyglet.input import Controller
 
 import arcade
+from arcade.gl.enums import NEAREST
 from arcade import Rect
 from arcade.examples.gui.exp_controller_support_grid import (
     ControllerIndicator,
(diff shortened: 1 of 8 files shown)


arcade/examples/gui/exp_controller_inventory.py
```diff
@@ -18,7 +18,7 @@
 
 import pyglet.font
 from pyglet.event import EVENT_HANDLED
-from pyglet.gl import GL_NEAREST
+from pyglet.graphics.api.gl import GL_NEAREST
```
Cleptomania: This should probably import from arcade.gl.enums instead. Importing this when using WebGL will error, but the top level `arcade.gl` package is backend agnostic.

Cleptomania (changes requested):

arcade/examples/gui/exp_controller_inventory.py
pvcraven: Good catch, updated to use arcade.gl.enums.NEAREST instead.

pyproject.toml
```diff
@@ -20,9 +20,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "pillow>=11.3.0",
+    "pillow>=12.2.0",
```
Cleptomania: The pillow version update is kind of annoying here. For web support, pillow is built into pyodide, meaning the pillow version available is tied to what pyodide has. Currently the latest version of pyodide still only has pillow 11.3.0, so requiring 12.2.0 will cause the arcade installation to error. So we should probably leave this at 11.3.0 for now given I don't think Arcade needs any code changes between them. The list of package versions in pyodide is here: https://pyodide.org/en/stable/usage/packages-in-pyodide.html

This is kind of the unfortunate part of web deployment at the moment, is that pypi doesn't support distribution of WASM wheels, so you can't install packages with native code dynamically yet, even if they build a WASM wheel. For example, pymunk actually has a WASM build that is published to github, and you can install it to pyodide using that wheel, but you can't put it in a project dependency because pypi doesn't support it.

Cleptomania (changes requested):

pyproject.toml
pvcraven: Ah, didn't know any of that. Thanks. Reverted.

Cleptomania (approved):

PR #2875, opened 2026-07-25
fix(math): rotate_around_point offsets from the rotation center

### LINKED ISSUE #2753: Bug: rotate_around_point function in math.py returns incorrect results due to wrong translation

Sanjays2402:
Closes #2753

`rotate_around_point` computed the rotated displacement vector correctly but added it to the target point instead of the source (the rotation center), so the result came out displaced by the source-to-target vector: rotating `(5, 0)` around `(3, 0)` by 90 degrees returned `(5, 2)` instead of `(3, 2)`.

Added a regression test in `tests/unit/test_math.py` (fails on `development`, passes with the fix) that also covers the full-turn and zero-radius early returns.

### DIFF
--- a/arcade/math.py
+++ b/arcade/math.py
@@ -421,7 +421,7 @@ def rotate_around_point(source: Point2, target: Point2, angle: float):
     dx = diff_x * c - diff_y * s
     dy = diff_x * s + diff_y * c
 
-    return target[0] + dx, target[1] + dy
+    return source[0] + dx, source[1] + dy
 
 
 def get_angle_degrees(x1: float, y1: float, x2: float, y2: float) -> float:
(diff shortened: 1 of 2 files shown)


pvcraven:
Fixed with  PR 2878. Both of these PRs were legit fixes for same issue.

PR #2890, opened 2026-09-17
Create python-publish.yml

yFish26:


Cleptomania:
We already have a UV based CI publishing workflow for PyPI releases which is already using trusted publishers.

